PR #3, opened 2019-08-15
Update from internal Khronos GitLab

keryell:


keryell:
@StuartDAdams @bader @tremmelg please sign the CLA or reply that it is not working so we can investigate...

bader:
> please sign the CLA or reply that it is not working so we can investigate...

Seems to work for me.

StuartDAdams:
Apologies for taking so long to sign.

PR #5, opened 2019-08-23
Add initial support for hipSYCL

illuhad:
This extends the cmake infrastructure to allow for compilation with hipSYCL. Other implementations are not affected. Also note that compilation at the moment still fails (mostly due to OpenCL stuff) and is still work in progress.

Signed-off-by: Aksel Alpay <aksel.alpay@uni-heidelberg.de>

### DIFF
--- a/cmake/FindhipSYCL.cmake
+++ b/cmake/FindhipSYCL.cmake
@@ -1,0 +1,37 @@
+find_program(SYCLCC_EXECUTABLE syclcc-clang
+    PATH_SUFFIXES bin)
+
+if(NOT SYCLCC_EXECUTABLE)
+  message(SEND_ERROR "Could not find hipSYCL syclcc-clang compiler")
+endif()
+
+set(CMAKE_C_COMPILER    ${SYCLCC_EXECUTABLE})
+set(CMAKE_CXX_COMPILER  ${SYCLCC_EXECUTABLE})
+set(CMAKE_CXX_STANDARD 14)
+
+add_library(SYCL::SYCL INTERFACE IMPORTED GLOBAL)
+# add_sycl_executable_implementation function
+# Builds a SYCL program, compiling multiple SYCL test case source files into a
+# test executable, invoking a single-source/device compiler
+# Parameters are:
+#   - NAME             Name of the test executable
+#   - OBJECT_LIBRARY   Name of the object library of all the compiled test cases
+#   - TESTS            List of SYCL test case source files to be built into the
+# test executable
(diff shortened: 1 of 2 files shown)


cmake/FindhipSYCL.cmake
```diff
@@ -0,0 +1,37 @@
+find_program(SYCLCC_EXECUTABLE syclcc-clang
+    PATH_SUFFIXES bin)
+
+if(SYCLCC_EXECUTABLE MATCHES "-NOTFOUND")
+  message(SEND_ERROR "Could not find hipSYCL syclcc-clang compiler")
+endif()
+
+set(CMAKE_C_COMPILER    ${SYCLCC_EXECUTABLE})
+set(CMAKE_CXX_COMPILER  ${SYCLCC_EXECUTABLE})
+set(CMAKE_CXX_STANDARD 14)
```
Ruyk: Is there any particular reason C++14 is required? SYCL 1.2.1 is C++11.

cmake/FindhipSYCL.cmake
```diff
@@ -0,0 +1,37 @@
+find_program(SYCLCC_EXECUTABLE syclcc-clang
+    PATH_SUFFIXES bin)
+
+if(SYCLCC_EXECUTABLE MATCHES "-NOTFOUND")
```
Ruyk: In CMake twisted logic, if `SYCLCC_EXECUTABLE` ends with `-NOTFOUND` its already false, so the MATCHES is not required.
Quoting CMake documentation:
```markdown
if(<constant>)

True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number. 
**False** if the constant is 0, OFF, NO, FALSE, N, IGNORE, **NOTFOUND**, the empty string, **or ends in the suffix NOTFOUND**. 
Named boolean constants are case-insensitive. 
If the argument is not one of these specific constants, it is treated as a variable or string and the following signature is used.
```

cmake/FindhipSYCL.cmake
illuhad: Yes, hipSYCL requires C++14.

cmake/FindhipSYCL.cmake
illuhad: Okay, I can change this if you like.

illuhad: I have corrected this as requested.

bader:
@keryell, please, merge this PR - we have all required approvals.

keryell:
Thank you for your contribution!

PR #7, opened 2019-08-26
Make image accessor feature checks conditional

bader:
Images is optional feature, so conformance tests must skip the test
if feature is not supported.

Add image support check to accessor_constructors_image test.

### DIFF
--- a/tests/accessor/accessor_api_image.cpp
+++ b/tests/accessor/accessor_api_image.cpp
@@ -38,19 +38,26 @@ class TEST_NAME : public util::test_base {
     try {
       auto queue = util::get_cts_object::queue();
 
-      /** check image accessor api for cl_int4
-       */
-      check_image_accessor_api_type<cl::sycl::cl_int4>(log, queue);
+      try {
+        /** check image accessor api for cl_int4
+         */
+        check_image_accessor_api_type<cl::sycl::cl_int4>(log, queue);
 
-      /** check image accessor api for cl_uint4
-       */
-      check_image_accessor_api_type<cl::sycl::cl_uint4>(log, queue);
+        /** check image accessor api for cl_uint4
+         */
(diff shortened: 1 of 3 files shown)


tests/accessor/accessor_api_image.cpp
```diff
@@ -38,19 +38,26 @@ class TEST_NAME : public util::test_base {
     try {
       auto queue = util::get_cts_object::queue();
 
-      /** check image accessor api for cl_int4
-       */
-      check_image_accessor_api_type<cl::sycl::cl_int4>(log, queue);
+      try {
```
AerialMantis: This check looks correct, however this relies on the implementation throwing a `feature_not_supported` exception, which I believe would be considered negative testing and for the Khronos CTS we can only use positive testing.

We could alter this so that the check for image support is done before doing the checks.

keryell: I am not sure it is really negative testing:
- if the feature is supported fine;
- if the feature is not supported, it should throw and the check is skipped from the CTS. So it is not really negative testing since it has no side effect;
- if the feature is not supported but does as if, the CTS will probably fail as the operation will not work correctly.

bader: I agree with @keryell, that it not really a negative test, but we might want to apply @AerialMantis suggestion for the sake of consistency.
I think the approach we use to skip tests relying on unsupported features should be consistent for the whole project.
Does it make sense?

bader:
@AerialMantis, https://github.com/KhronosGroup/SYCL-CTS/pull/7/commits/8242d241fcf07348c6fa90356f0a79159e36b493 should address your comment.

fraggamuffin:
Sept 9:
*  Gordon asked for some refactoring
*  do the runtime check before the test
*  applied, wait Gordon review

PR #8, opened 2019-08-29
Remove dependency on OpenCL from queue test

bader:


bader:
The intent is to avoid using OpenCL API directly in SYCL tests except `opencl_interop`.

AerialMantis:
That makes sense, we generally tried to restrict the OpenCL API calls to the interop tests, but some of the non-interop tests still required checks on `get` such as this.

Rather than removing it, could you move the check to `opencl_interop` instead?

keryell:
This reminds me the private https://gitlab.khronos.org/sycl/conformance-testsuite/issues/74

bader:
@AerialMantis, https://github.com/KhronosGroup/SYCL-CTS/pull/8/commits/82558f573862a5cc872c2bd587dde17646613cb8 should address your [comment](https://github.com/KhronosGroup/SYCL-CTS/pull/8#issuecomment-527887472).

fraggamuffin:
Sept 9:
*  need Gordon to approve moving the functionality from one test to another - opencl INterop
*  reviewed change , this is the only change left

AerialMantis:
This looks good to me, thanks @bader

PR #10, opened 2019-09-03
Avoid narrowing conversion error when size of tested types does not m…

bader:
…atch 64 bits

### DIFF
--- a/tests/atomic/atomic_api_64_extended.cpp
+++ b/tests/atomic/atomic_api_64_extended.cpp
@@ -81,9 +81,17 @@ class TEST_NAME : public util::test_base {
     set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
   }
 
-  template <typename T>
+  template <typename T,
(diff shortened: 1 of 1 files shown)


tests/atomic/atomic_api_64_extended.cpp
```diff
@@ -81,11 +81,19 @@ class TEST_NAME : public util::test_base {
     set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
   }
 
-  template <typename T>
+  template <typename T,
+            typename std::enable_if<sizeof(T) == 8>::type * = nullptr>
   void check_atomics_for_type(util::logger &log, cl::sycl::queue testQueue) {
     return generic_check_for_atomics<T, check_extended_atomics>(log, testQueue);
```
keryell: I am puzzled by this code...

But it looks it is possible to return a `void` returned by a `void` function...
https://godbolt.org/z/296pjT

Nevertheless it is not possible to `return void {};`...

PR #11, opened 2019-09-04
Do not rely on availability of compiler/linker

keryell:
Group API test builds test kernels at run time using
build_with_kernel_type method of cl::sycl::program class.
Skip these test cases if online compilation is not supported.

Signed-off-by: Alexey Bader <alexey.bader@intel.com>
(cherry picked from commit be801b896f64684c5f5c4b04cc5fd16c233009ba)


fraggamuffin:
Sept 9:
* already merged.

PR #12, opened 2019-09-26
Enable opencl_interop on implementations w/o online compilation support

bader:
Test devices that do not support online compile using binaries in same
directory as executable

### DIFF
--- a/tests/opencl_interop/opencl_interop_constructors.cpp
+++ b/tests/opencl_interop/opencl_interop_constructors.cpp
@@ -149,7 +149,7 @@ class TEST_NAME : public sycl_cts::util::test_base_opencl {
       /** check program (context, cl_program) constructor
        */
       {
-        cl_program clProgram = nullptr;
+        cl_program clProgram{};
         if (online_compiler_supported(ctsDevice.get(), log)) {
           if (!create_built_program(kernelSource, ctsContext.get(),
                                     ctsDevice.get(), clProgram, log)) {
@@ -181,7 +181,7 @@ class TEST_NAME : public sycl_cts::util::test_base_opencl {
       /** check kernel (cl_kernel, const context&) constructor
        */
       {
-        cl_program clProgram = nullptr;
+        cl_program clProgram{};
         if (online_compiler_supported(ctsDevice.get(), log)) {
           if (!create_built_program(kernelSource, ctsContext.get(),
                                     ctsDevice.get(), clProgram, log)) {
@@ -199,7 +199,7 @@ class TEST_NAME : public sycl_cts::util::test_base_opencl {
           }
         }
 
-        cl_kernel clKernel = nullptr;
+        cl_kernel clKernel{};
         if (!create_kernel(clProgram, "opencl_interop_constructors_kernel",
                            clKernel, log)) {
           FAIL(log, "create_kernel failed");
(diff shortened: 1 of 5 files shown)


tests/opencl_interop/opencl_interop_constructors.cpp
```diff
@@ -147,54 +149,70 @@ class TEST_NAME : public sycl_cts::util::test_base_opencl {
       /** check program (context, cl_program) constructor
        */
       {
+        cl_program clProgram = nullptr;
```
keryell: Is it standard OpenCL to assume that a `cl_program` is a pointer, by the way?

util/test_base_opencl.cpp
```diff
@@ -107,6 +113,28 @@ bool test_base_opencl::online_compiler_supported(cl_device_id clDeviceId,
...
+#else
+  ssize_t n = readlink("/proc/self/exe", path, max_path_len - 1);
+  if (n < 0) {
+    return false;
+  }
+  path[n] = '\0';
+#endif
+  // Replace all characters with '\0' from back until '/' or '\\'
+  for (size_t i = strnlen(path, max_path_len) - 1;
+       i > 0 && path[i] != '/' && path[i] != '\\'; --i) {
```
keryell: You have to decide whether you are on Windows or not here too.
For example this does not work on Unix if you have `\` in the path...

tests/opencl_interop/opencl_interop_constructors.cpp
bader: `cl_program` is defined as a [pointer](https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl.h#L46) in OpenCL-Headers.
I don't know better way to initialize `cl_program` object, so I open to suggestions.

util/test_base_opencl.cpp
bader: Done.

util/test_base_opencl.cpp
```diff
@@ -107,6 +113,31 @@ bool test_base_opencl::online_compiler_supported(cl_device_id clDeviceId,
...
 }
 
+bool test_base_opencl::get_exec_dir(char *path, size_t max_path_len) {
+  assert(max_path_len > 0 && "No space to store the path");
+#ifdef _MSC_VER
+  HMODULE hMod = GetModuleHandle(NULL);
+  if (!GetModuleFileNameA(hMod, path, max_path_len)) {
+    return false;
+  }
+  const char pathDelim = '\\';
```
keryell: `constexpr` instead of `const` to avoid any storage
```
   constexpr auto pathDelim = '\\';
```

tests/opencl_interop/opencl_interop_constructors.cpp
keryell: Possible suggestions:
- keep your solution because this is a well-known implementation that leaks to all the implementation because of ICD;
- improve the OpenCL specification to introduce initializers in the standard (probably required by things like AUTOSAR/MISRA/...);
- do not use such an unsafe API :-)

So probably your solution is good to have things done... :-)

keryell: Actually C++ provides a portable way to do it:
```
cl_program clProgram {};
```

bader: I applied this suggestion in https://github.com/KhronosGroup/SYCL-CTS/pull/12/commits/aab8c8dfe4903455937dad2b597e1d1eeb220d95.

PR #20, opened 2019-11-11
Fix more checks for accessor's get_count()/get_size() methods

bader:
Made these methods return the values of the accessor, not the buffer
from which this accessor was created.

Patch by Alexey Voronov <alexey.voronov@intel.com>.


bader:
> I guess the problem arise when using an accessor constructor with an explicit range and/or offset.

"the problem arise when using an accessor constructor with an explicit range **not matching buffer range**"

keryell:
> > I guess the problem arise when using an accessor constructor with an explicit range and/or offset.
> 
> "the problem arise when using an accessor constructor with an explicit range **not matching buffer range**"

Yes, that is true that it is possible to construct an accessor with a range which is exactly the same as the buffer range... :-)

PR #25, opened 2019-11-27
Allow additional flags to be passed to Intel SYCL

### LINKED ISSUE #13: [SYCL-CTS] The CTS selector queries name for vendor, instead of simply vendor
The snippet below shows what I think its an obvious typo:

https://github.com/KhronosGroup/SYCL-CTS/blob/7e469d115df4326b4683dd7ce200d1230ff27014/tests/common/cts_selector.h#L100-L101

If everyone agrees it is a typo, we can submit an MR with a fix.

StuartDAdams:
This patch allows new options for Intel SYCL to be passed to the SYCL CTS.

A minor fix is included that will resolve #13. The correct vendor information is now used within the device selector.

### DIFF
--- a/tests/common/cts_selector.h
+++ b/tests/common/cts_selector.h
@@ -98,7 +98,7 @@ class cts_selector : public cl::sycl::device_selector {
     }
 
     cl::sycl::string_class vendor =
-        dev.get_platform().get_info<cl::sycl::info::platform::name>();
+        dev.get_platform().get_info<cl::sycl::info::platform::vendor>();
     cl::sycl::info::device_type type =
         dev.get_info<cl::sycl::info::device::device_type>();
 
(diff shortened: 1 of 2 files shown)


tests/common/cts_selector.h
```diff
@@ -98,7 +98,7 @@ class cts_selector : public cl::sycl::device_selector {
     }
 
     cl::sycl::string_class vendor =
-        dev.get_platform().get_info<cl::sycl::info::platform::name>();
+        dev.get_platform().get_info<cl::sycl::info::platform::vendor>();
```
keryell: This looks like a bug fix...

keryell: Rebooting my brain... Yes as said in the PR description this fixes https://github.com/KhronosGroup/SYCL-CTS/issues/13

PR #28, opened 2019-12-04
Add missing clReleaseSampler

bader:


keryell:
A good example why we should rather promote interoperability with `cl2.hpp` or better, rather than the plain OpenCL C host API interface. This is too dangerous, even for the specialists...

PR #29, opened 2019-12-04
Make test_base_opencl framework respect SYCL device selector choice

### LINKED ISSUE #21: Update CTS for stream interface clarifications
Tracking of CTS investigation for spec MR: https://github.com/KhronosGroup/SYCL-Docs/pull/58

bader:
Fixes #21

### DIFF
--- a/util/test_base_opencl.cpp
+++ b/util/test_base_opencl.cpp
@@ -46,7 +46,7 @@ bool test_base_opencl::setup(logger &log) {
   cts_selector ctsSelector;
   const auto ctsContext = util::get_cts_object::context(ctsSelector);
 
-  if (ctsContext.get_devices().size() < 1) {
+  if (ctsContext.get_devices().empty()) {
     FAIL(log, "Unable to retrieve list of devices via cts_selector");
     return false;
   }
--- a/util/test_base_opencl.cpp
+++ b/util/test_base_opencl.cpp
@@ -7,6 +7,8 @@
 *******************************************************************************/
 
 #include "test_base_opencl.h"
+#include "../tests/common/cts_selector.h"
+#include "../tests/common/get_cts_object.h"
 #include "../tests/common/macros.h"
 #include "../util/opencl_helper.h"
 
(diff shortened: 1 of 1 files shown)


util/test_base_opencl.cpp
```diff
@@ -33,64 +35,32 @@ bool test_base_opencl::setup(logger &log) {
...
-    FAIL(log, "Unable to retrieve list of platforms via clGetPlatformIDs()");
-    return false;
-  }
+  cts_selector ctsSelector;
+  const auto ctsContext = util::get_cts_object::context(ctsSelector);
 
-  platforms.reset(new cl_platform_id[N]);
-  if (platforms.get() == nullptr) {
-    FAIL(log, "Malloc failed to allocate memory for platform list");
+   if ( ctsContext.get_devices().size() < 1) {
```
keryell: Weird spacing

bader: Thanks!
Fixed.

ProGTX:
I've looked into this and the selector chooses a host device when running the test on the host, and then the `.get()` function throws. This affects the `buffer`, `kernel`, `opencl_interop`, and `program` tests. I was hoping that updating the branch to https://github.com/KhronosGroup/SYCL-CTS/pull/24/ would fix this, but it doesn't, `test_base_opencl::setup` is still being called for the host device.

bader:
> I've looked into this and the selector chooses a host device when running the test on the host, and then the `.get()` function throws.

I think we can skip setup if the selector chooses a host device. Right?

> This affects the `buffer`, `kernel`, `opencl_interop`, and `program` tests. I was hoping that updating the branch to #24 would fix this, but it doesn't, `test_base_opencl::setup` is still being called for the host device.

Hm... I expect only `opencl_inteop` test to be affected as #24 replaced `test_base_opencl::setup` to `test_base::setup` for `buffer`, `kernel` and `program` tests. Did I miss something?

bader:
@ProGTX, I've updated the patch. Please, let me know if you have any issues with it.

bader:
@ProGTX, sorry for the mess, but I somehow missed the fix. Please, check https://github.com/KhronosGroup/SYCL-CTS/commit/40fe5d21f5a578a196d2822de7cdbb41612edd38.

bader:
@AerialMantis, great news. Thanks for the update.
@keryell, could you approve updated version of the patch, please?

util/test_base_opencl.cpp
```diff
@@ -39,64 +41,35 @@ bool test_base_opencl::setup(logger &log) {
...
-
-  error =
-      clGetDeviceIDs(platforms.get()[0], CL_DEVICE_TYPE_ALL, 0, nullptr, &N);
-  if (!CHECK_CL_SUCCESS(log, error)) return false;
+  cts_selector ctsSelector;
+  const auto ctsContext = util::get_cts_object::context(ctsSelector);
 
-  if (N < 1) {
-    FAIL(log, "Unable to retrieve list of devices via clGetDeviceIDs()");
+  if (ctsContext.get_devices().size() < 1) {
```
keryell:
```suggestion
  if (ctsContext.get_devices().empty()) {
```

bader: Done.

PR #30, opened 2019-12-06
Support additional CTest arguments

ProGTX:
1. `--additional-ctest-args`
2. Replaces the `--verbose` option
3. Include CTest call in conformance report

### DIFF
--- a/run_conformance_tests.py
+++ b/run_conformance_tests.py
@@ -5,6 +5,7 @@ import sys
 import xml.etree.ElementTree as ET
 import json
 import argparse
+import shlex
 
 REPORT_HEADER = """<?xml version="1.0" encoding="UTF-8"?>
 <?xml-stylesheet xmlns="http://www.w3.org/1999/xhtml" type="text/xsl" href="#stylesheet"?>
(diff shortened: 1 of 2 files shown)


run_conformance_tests.py
```diff
@@ -81,22 +82,19 @@ def handle_args(argv):
         help='The name of the implementation to be displayed in the report.',
         type=str,
         required=True)
-    parser.add_argument(
-        '-V',
-        '--verbose',
-        help='Enable verbose CTest output',
-        type=bool,
-        default=False,
-        required=False)
```
keryell: Where is this verbose mode gone?

run_conformance_tests.py
```diff
@@ -105,7 +103,6 @@ def generate_cmake_call(cmake_exe, build_system_name, conformance_filter,
     Generates a CMake call based on the input in a form accepted by
     subprocess.call().
     """
-    import shlex
```
keryell: Quite less obfuscated... :-)

run_conformance_tests.py
```diff
@@ -127,21 +135,15 @@ def subprocess_call(parameter_list):
...
     Runs ctest, overwriting any cached results.
     """
 
     build_system_call = build_system_call.split()
-    ctest_call = [
-        'ctest', '.', '-T', 'Test', '--no-compress-output',
-        '--test-output-size-passed', '0', '--test-output-size-failed', '0'
-    ]
-    if verbose:
-        ctest_call.append('-V')
```
keryell: I see. You have generalized the verbose case.
I was thinking "verbose" for the Python script first...

ProGTX: I guess the option was a bit deceiving before, as it referred only to CTest.

run_conformance_tests.py
ProGTX: Replaced by `--additional-ctest-args`, so the caller can pass `-V` there instead

PR #31, opened 2019-12-11
[NFC] Remove unnecessary const qualifier

bader:
constexpr implies const in this case.


util/test_base_opencl.cpp
```diff
@@ -127,7 +127,7 @@ bool test_base_opencl::get_exec_dir(char *path, size_t max_path_len) {
     return false;
   }
   path[n] = '\0';
-  constexpr const char pathDelim = '/';
+  constexpr char pathDelim = '/';
```
keryell: I wonder if we could have in C++20
```
constexpr const constinit consteval char pathDelim = '/';
```
just to be sure it will not move... :-)

PR #32, opened 2019-12-11
Adjust work-group size in nd_item mem_fence test

bader:
Align requested work-group size with the implementation limitations.
- requested work-group size can't be greater than the maximum work-group
  size supported by the device
- requested work-group size can't be greater than the maximum work-group
  size supported by a kernel on specified device

Signed-off-by: Alexey Bader <alexey.bader@intel.com>

### DIFF
--- a/tests/nd_item/nd_item_mem_fence.cpp
+++ b/tests/nd_item/nd_item_mem_fence.cpp
@@ -28,8 +28,11 @@ void test_mem_fence(util::logger &log, cl::sycl::queue &queue) {
     localSize = (maxDeviceWorkGroupSize < localSize) ? maxDeviceWorkGroupSize
                                                      : localSize;
 
-    /* check work-group size kernel limit */
-    /* this check requires online compiler */
+    /* Check work-group size kernel limit - it must be >= 2 to test
+     * nd_item::mem_fence member function. To query
+     * info::kernel_work_group::work_group_size property, we need obtain test
+     * kernel handler, which requires online compilation
+     * */
     auto devices = queue.get_context().get_devices();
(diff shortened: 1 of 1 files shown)


tests/nd_item/nd_item_mem_fence.cpp
```diff
@@ -16,8 +16,32 @@ using namespace sycl_cts;
...
+    auto maxDeviceWorkGroupSize =
+        device.template get_info<cl::sycl::info::device::max_work_group_size>();
+    localSize = (maxDeviceWorkGroupSize < localSize) ? maxDeviceWorkGroupSize
+                                                     : localSize;
+
+    /* check work-group size kernel limit */
+    /* this check requires online compiler */
+    auto devices = queue.get_context().get_devices();
+    if (!is_compiler_available(devices) || !is_linker_available(devices))
+      return;
```
AerialMantis: If the devices being targeted do not support online compilation this function returns so the test is skipped, is this the intention?

bader: Yes. Online compilation is required to query `cl::sycl::info::kernel_work_group::work_group_size` property. We must make sure that value for this property is >= 2 to test `mem_fence` method.

keryell: This is a good motivation showing why we need more comments in the code, because otherwise there are more questions during review... Feel free to add this as a comment for clarification. :-)

bader: Added more details in the comments.
@AerialMantis, please, let me know if any changes are required.

keryell: Great clarification!
Thanks.

PR #33, opened 2019-12-13
[NFC] Fix two kinds of compilation warnings in generated files

bader:
- Fix #pragma OPENCL EXTENSION usage 
- Fix uninitialized variable warning

### DIFF
--- a/tests/math_builtin_api/generate_math_builtin.py
+++ b/tests/math_builtin_api/generate_math_builtin.py
@@ -47,7 +47,7 @@ def write_cases_to_file(generated_test_cases, inputFile, outputFile, extension=N
     if extension is None:
         extension = ""
     else:
-        extension = "#ifdef __SYCL_DEVICE_ONLY__\n#pragma OPENCL EXTENSION %s : enable\n#endif" % extension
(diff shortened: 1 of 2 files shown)


tests/math_builtin_api/generate_math_builtin.py
```diff
@@ -47,7 +47,7 @@ def write_cases_to_file(generated_test_cases, inputFile, outputFile, extension=N
     if extension is None:
         extension = ""
     else:
-        extension = "#ifdef __SYCL_DEVICE_ONLY__\n#pragma OPENCL EXTENSION %s : enable\n#endif" % extension
+        extension = "#ifdef __SYCL_DEVICE_ONLY__\n#ifdef $s\n#pragma OPENCL EXTENSION %s : enable\n#endif\n#endif" % extension
```
keryell: By looking at this and above it looks like all this is very old-fashioned...
Now there are directly templates https://docs.python.org/3/library/string.html#template-strings and more Python 3 friendly format https://docs.python.org/3/library/stdtypes.html#str.format instead of manual templates with `replace` and `%`.
Some stuff for the future generations to do...
Interestingly, the other file of this PR uses this more modern Python approach.

PR #34, opened 2019-12-13
[NFC] Remove dead code from vector_constructors test

bader:
This file is ignored by the CMake build system.
test_vector_constructors sources are generated by the python script and
they cover cases tested in this file.


Ruyk:
@ProGTX

PR #41, opened 2020-01-17
Enhance handler copy test cases

bader:
Add new tests covering cases when source and destination parameters of
"explicit copy API" have different types.


keryell:
That looks like the only missing commit from internal GitLab.
Thanks.

PR #35, opened 2019-12-18
Accessor `get_size` and `get_range` fixes

ProGTX:
* https://github.com/KhronosGroup/SYCL-Docs/pull/31
* Removed `get_size` check for image accessors
* Check `get_range` for local accessors
* Check `get_range` for image accessors

PR #44, opened 2020-02-04
Update image accessors API

bader:
Follow-up https://github.com/KhronosGroup/SYCL-CTS/pull/35, which
adjusts accessor API change from
https://github.com/KhronosGroup/SYCL-Docs/pull/31.


bader:
@ProGTX, FYI.

PR #48, opened 2020-03-11
Re-enabled hierarchical_implicit_reduce

### LINKED ISSUE #46: hierarchical_implicit_reduce test is disabled due to return at the very first line
https://github.com/KhronosGroup/SYCL-CTS/blob/806294de8e9e69fe0057432f3c4fdb668b12db0b/tests/hierarchical/hierarchical_implicit_reduce.cpp#L150-L154

ProGTX:
Fixes #46 

* Removed early return
* Fixed segfault
    * Use `localId` as index instead of `i` for local data
* Use 64-bit integers
* Simplified `Adder` and `Multiplier`
* Fixed default value for `Multiplier`
* Use `write` access mode for group sums in first pass
* Ensure `mTotal` is written to `parallel_for_work_group`
* Pass proper ranges to
* Better error messages
* clang-format

### DIFF
--- a/tests/hierarchical/hierarchical_implicit_reduce.cpp
+++ b/tests/hierarchical/hierarchical_implicit_reduce.cpp
@@ -8,17 +8,26 @@
 
 #include "../common/common.h"
 
-#define TEST_NAME hierarchical_reduce
+#define TEST_NAME hierarchical_implicit_reduce
 
 namespace TEST_NAMESPACE {
 
-static const int groupItems1d = 2;
-static const int localItems1d = 2;
-static const int groupItemsTotal = groupItems1d * groupItems1d * groupItems1d;
-static const int localItemsTotal = localItems1d * localItems1d * localItems1d;
-static const int numGroups = groupItemsTotal / localItemsTotal;
+static constexpr int globalItems1d = 26;
+static constexpr int globalItems2d = 33;
+static constexpr int globalItems3d = 35;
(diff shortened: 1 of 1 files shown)


tests/hierarchical/hierarchical_implicit_reduce.cpp
```diff
@@ -8,15 +8,23 @@
...
+static const int localItems3d = 1;
+static const int groupItems1d = (globalItems1d / localItems1d);
+static const int groupItems2d = (globalItems2d / localItems2d);
+static const int groupItems3d = (globalItems3d / localItems3d);
+static const int globalItemsTotal =
+    (globalItems1d * globalItems2d * globalItems3d);
+static const int localItemsTotal = (localItems1d * localItems2d * localItems3d);
+static const int numGroups = (globalItemsTotal / localItemsTotal);
 
 static const int inputSize = 32;
```
keryell: I guess that most of these could be constexpr.
Otherwise it would be nice to have some less trivial sizes to detect some trivial bugs, such as wrong dimensions, transpositions, etc.
For example use some different small prime numbers.

ProGTX: Done and done, they're all prime numbers now - global items are a product of two prime numbers.

AerialMantis:
Thanks, merging this as we have two implementor approvals.

PR #51, opened 2020-04-20
Set INTERFACE_LINK_OPTIONS for Clang

alan-forbes-cp:
Set specific INTERFACE_LINK_OPTIONS option for Clang.

### DIFF
--- a/cmake/FindIntel_SYCL.cmake
+++ b/cmake/FindIntel_SYCL.cmake
@@ -16,6 +16,7 @@ if(DEFINED INTEL_SYCL_FLAGS)
 endif()
 
 add_library(INTEL_SYCL::Runtime INTERFACE IMPORTED GLOBAL)
+# -fsycl-device-code-split option not yet supported for CUDA
 if(${INTEL_SYCL_TRIPLE} MATCHES ".*-nvidia-cuda-.*")
     set_target_properties(INTEL_SYCL::Runtime PROPERTIES
         INTERFACE_LINK_LIBRARIES    OpenCL::OpenCL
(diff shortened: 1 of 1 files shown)


bader:
@alan-forbes-cp, could you elaborate on the motivation for this change, please?
I don't get the reasons to remove `-fsycl-device-code-split=per_source` option. AFAIK, it improves compilation time for the tests like `vector_*` by splitting huge device module into smaller parts.

alan-forbes-cp:
Apologies @bader - that commit was incorrect. It should apply to CUDA only. I have updated the PR accordingly. Thanks.

bader:
@alan-forbes-cp, could you clarify why this change is needed for CUDA, please?

alan-forbes-cp:
@bader The option is not currently supported for CUDA and causes a CTS build fail as a result.

bader:
Thanks for clarification. Could you it to the cmake file as a comment, please?

Do you know if there are plans to support this feature in future? Should we file an issue to enable this support?

alan-forbes-cp:
Done. Support has been discussed informally - but there are no firm plans at the moment to my knowledge. So - yes - please file an issue.

Thanks!

bader:
@keryell, @illuhad are you interested in reviewing changes in CMake configuration files for the implementation from `intel/llvm`?

keryell:
> @keryell, @illuhad are you interested in reviewing changes in CMake configuration files for the implementation from `intel/llvm`?

Thanks for the invite.
I am not familiar with this...
Just curious to have somthing talking about OpenCL inside a test for CUDA only... :-)

illuhad:
Same here, also wondering about OpenCL inside the CUDA section...

bader:
@alan-forbes-cp should know better, but it might be related to this issue: https://github.com/intel/llvm/issues/1559.

Ruyk:
It is related to https://github.com/intel/llvm/issues/1559. The DPC++ compiler needs the OpenCL symbols and headers somewhere even when using CUDA, but is not used at runtime.

alan-forbes-cp:
Comments updated as requested. Thanks!

PR #52, opened 2020-07-02
Change accessor image channel data types when using DPCPP CUDA backend

steffenlarsen:
The DPCPP CUDA backend currently only support reading from and writing to image accessors with data type corresponding to the channel data type of the underlying image.

This changes the channel types of images used in the accessor tests to correspond to the supported accessor data types when using the DPCPP CUDA backend.

### DIFF
--- a/tests/accessor/accessor_api_image_common.h
+++ b/tests/accessor/accessor_api_image_common.h
@@ -166,11 +166,17 @@ struct image_format_channel;
 */
 template <>
 struct image_format_channel<cl::sycl::cl_int4> {
+#ifdef SYCL_CTS_INTEL_PI_CUDA
+  static constexpr cl::sycl::image_channel_type type =
+      cl::sycl::image_channel_type::signed_int32;
+  using storage_t = cl::sycl::cl_int;
+#else
   static constexpr cl::sycl::image_channel_type type =
       cl::sycl::image_channel_type::signed_int8;
+  using storage_t = cl::sycl::cl_char;
+#endif
   static constexpr cl::sycl::image_channel_order order =
       cl::sycl::image_channel_order::rgba;
-  using storage_t = cl::sycl::cl_char;
   static constexpr size_t elementSize = 4 * sizeof(storage_t);
 };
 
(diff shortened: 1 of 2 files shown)


bader (approved):

tests/accessor/accessor_api_image_common.h
```diff
@@ -166,23 +166,35 @@ struct image_format_channel;
 */
 template <>
 struct image_format_channel<cl::sycl::cl_int4> {
+#ifdef SYCL_CTS_INTEL_PI_CUDA
+  static constexpr cl::sycl::image_channel_type type =
+      cl::sycl::image_channel_type::signed_int32;
```
keryell, line 171: Just curious: does this mean that the pixel values are still in [-128, +127] range?

steffenlarsen: It's been some time and the image tests are a bit convoluted, but I believe - based on `get_expected_image_value` - that the values are either 0 or 17, so nothing really changed with the pixel values.

keryell: OK, so 17 is the value we test here https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/tests/accessor/accessor_api_image_common.h#L243
But my question is actually broader and after looking at the spec now I have a more generic question.
Since I am not working on GPU I am unfamiliar with all these images...
It looks like the DPC++ CUDA backend does not support image with a channel type `cl::sycl::cl_int4` and so you update the test to use with just 1 `cl::sycl::cl_int` instead and assume that somehow the byte ordering will work when you extract the `argb` values. Is it what's going on here?
If so, can we expect the CUDA backend to provide the requested image API at the first place or do we expect the CUDA back-end to advertise this image type unsupported?

steffenlarsen: The image support in the DPC++ CUDA backend is currently limited to reading/writing halfs, floats, 32bit uints and 32bit ints. In short this is because the image doesn't carry any information about its format after creation, i.e. if we create an image with `cl::sycl::cl_int4` it will be an image where each pixel is 4x32 bits. Since the operations on the image doesn't know the format it relies on read/written value having the same size as the channels in order to correctly calculate offsets.

I believe the CUDA backend tells the user if they're using unsupported channel types.

steffenlarsen: https://github.com/codeplaysoftware/standards-proposals/pull/131 explains the limitations and suggests an extension for a work-around for SYCL 1.2.1.

keryell (approved):

PR #54, opened 2020-08-13
Test that handler invoke APIs pass correct indices to kernels

psalz:
I've noticed on various occasions that for some SYCL implementations and platform combinations, the global kernel offset mechanism doesn't seem to work quite right, in that the ids received by the kernel function are not properly shifted one way or another. Since the offset API is very near and dear to my heart, I'd like to propose adding functionality to the CTS that explicitly tests for this.

To that end, I've adjusted the `handler_invoke_api` test suite to not only check whether a kernel is launched at all, but also whether it receives the correct indices corresponding to the requested iteration space. I guess it could be debated whether these checks belong to the `item`/`h_item`/`nd_item` tests instead, but I feel like they better fit with the launch API, as this is where the iteration space is being defined.

To actually test for this, each kernel thread now writes its global work item id into an output buffer, and the `check_api_call` function checks whether all expected ids are present afterwards. The buffer is initialized with canary values to verify that only the expected range of work items was launched. The offset overloads of `parallel_for` now receive non-zero offsets and I've also added offset tests for the `nd_range` variants.

I handled `parallel_for_work_item` variants that process an unknown number of work items by always only writing to index 0 (although their existence might be considered a spec bug, see https://github.com/KhronosGroup/SYCL-Docs/issues/85). This technically constitutes a data race if there is more than one physical thread, but since they all write the same value I think it should be fine (albeit not ideal). However to my understanding this was also the case previously.

What I didn't do for now but what could be considered for the future is to also test the correctness of `get_local_[physical|logical]_id`, and to extend the test cases to 2D/3D as well.

Here are some preliminary results I've obtained on whether offsets are implemented correctly:
| Impl    |  Platform | Correct? | Note                                                                      |
|---------|----------:|----------|---------------------------------------------------------------------------|
| DPC++   | Host      | NO*      | Works for `nd_range`, otherwise always starts at 0                          |
| DPC++   | CUDA      | NO*       | Works for `nd_range`, otherwise starts at correct offset, but goes beyond requested number of work items  |
| DPC++   | Intel OCL | YES      | |
| CCPP    | Host      | NO*      | Works for `nd_range`, otherwise always starts at 0                          |
| CCPP    | Intel OCL | YES      |                                                                           |
| CCPP    | PTX       | NO       | Always starts at 0                                                        |
| hipSYCL | CUDA      | YES*     | Had to hack quite a bit to get it to compile so take with a grain of salt |

### DIFF
--- a/tests/handler/handler_invoke_api.cpp
+++ b/tests/handler/handler_invoke_api.cpp
@@ -30,18 +30,23 @@ using accessor_t =
                        cl::sycl::access::target::global_buffer>;
 
 struct single_task_functor {
-  single_task_functor(accessor_t acc) : acc(acc) {}
+  single_task_functor(accessor_t acc, size_t range) : acc(acc), range(range) {}
 
-  void operator()() { acc[0] = 10; }
+  void operator()() {
+    for (size_t i = 0; i < range; ++i) {
+      acc[i] = i;
+    }
+  }
 
   accessor_t acc;
+  size_t range;
 };
 
 template <int useOffset>
 struct parallel_for_range_id_functor {
   parallel_for_range_id_functor(accessor_t acc) : acc(acc) {}
 
-  void operator()(cl::sycl::id<1> id) { acc[0] = 10; }
+  void operator()(cl::sycl::id<1> id) { acc[id] = id[0]; }
 
   accessor_t acc;
 };
@@ -50,7 +55,7 @@ template <int useOffset>
 struct parallel_for_range_item_functor {
   parallel_for_range_item_functor(accessor_t acc) : acc(acc) {}
 
-  void operator()(cl::sycl::item<1> item) { acc[0] = 10; }
+  void operator()(cl::sycl::item<1> item) { acc[item] = item[0]; }
 
   accessor_t acc;
 };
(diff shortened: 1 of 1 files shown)


bader:
Tagging @illuhad.

illuhad:
> hipSYCL | CUDA | YES* | Had to hack quite a bit to get it to compile so take with a grain of salt
> -- | -- | -- | --

Yes, last time I tried compiling CTS with hipSYCL I also had some trouble, mostly because of some OpenCL stuff. But I still believe that offsets should work as expected in hipSYCL since they are also tested in our unit tests. 

In general I agree with @psalz that this is something that we should test more explicitly, especially in light of the fact that SYCL is now more general than OpenCL. While with OpenCL one could always argue that the SYCL implementation only needs to forward offsets correctly to OpenCL, and the correct implementation is then the responsibility of OpenCL, this is no longer the case with SYCL backends (such as CUDA) that do not support offsets by themselves.

psalz:
Update: All tests pass on DPC++ Intel OpenCL (Xeon E5-4650). I've updated the table above.

tests/handler/handler_invoke_api.cpp
```diff
@@ -30,18 +30,23 @@ using accessor_t =
                        cl::sycl::access::target::global_buffer>;
 
 struct single_task_functor {
-  single_task_functor(accessor_t acc) : acc(acc) {}
+  single_task_functor(accessor_t acc, size_t range) : acc(acc), range(range) {}
```
ProGTX, line 33: I believe the accessor should be able to access its own range, unless maybe we're using ranged accessors?

ProGTX:
Sorry for being a bit late, but I've ran your branch through our testing and it seems to pass on both host and Intel CPU, am I missing something?

tests/handler/handler_invoke_api.cpp
psalz: The passed accessor has access to the full buffer range (256), but we only want to write `defaultRange` (128) to test whether an implementation writes more than it should (granted, highly unlikely in the single task variant, as it simply does a for-loop).

psalz:
> Sorry for being a bit late, but I've ran your branch through our testing and it seems to pass on both host and Intel CPU, am I missing something?

That's curious. On both Windows and Ubuntu, using ComputeCPP CE 2.1.0, I get this output for the host device:

```
./bin/test_handler -p host -d host
--- handler_invoke_api
  . Check single_task(lambda)
  . Check single_task(functor)
  . online compiler is not available -- skipping test of single_task with prebuilt kernel
  . Check parallel_for(range, lambda) with id
  . Check parallel_for(range, functor) with id
  . online compiler is not available -- skipping test of parallel_for with id with prebuilt kernel
  . Check parallel_for(range, id, lambda) with id
  . line: 188
  . Expected 12345 {12345} but got 0 {result[i]} at element 0
  . Check parallel_for(range, id, functor) with id
  . line: 188
  . Expected 12345 {12345} but got 0 {result[i]} at element 0
  . online compiler is not available -- skipping test of parallel_for with offset and id with prebuilt kernel
  . Check parallel_for(range, lambda) with item
  . Check parallel_for(range, functor) with item
  . online compiler is not available -- skipping test of parallel_for with prebuilt kernel
  . Check parallel_for(range, id, lambda) with item
  . line: 188
  . Expected 12345 {12345} but got 0 {result[i]} at element 0
  . Check parallel_for(range, id, functor) with item
  . line: 188
  . Expected 12345 {12345} but got 0 {result[i]} at element 0
  . online compiler is not available -- skipping test of parallel_for with offset and item with prebuilt kernel
  . Check parallel_for(nd_range, lambda)
  . Check parallel_for(nd_range, functor)
  . online compiler is not available -- skipping test of parallel_for over nd_range with prebuilt kernel
  . Check parallel_for(nd_range, lambda) with offset
  . Check parallel_for(nd_range, functor) with offset
  . online compiler is not available -- skipping test of parallel_for over nd_range with offset and prebuilt kernel
  . Check parallel_for_work_group(range, lambda)
  . Check parallel_for_work_group(range, functor)
  . online compiler is not available -- skipping test of parallel_for_work_group (range) with prebuilt kernel
  . Check parallel_for_work_group(range, range, lambda)
  . Check parallel_for_work_group(range, range, functor)
  . online compiler is not available -- skipping test of parallel_for_work_group (range, range) with prebuilt kernel
  . online compiler is not available -- skipping test of single_task with kernel object
  . online compiler is not available -- skipping test of parallel_for with kernel object
  - fail

  . Passed 0/1 tests (0%)
```

tests/handler/handler_invoke_api.cpp
```diff
@@ -126,23 +161,35 @@ class TEST_NAME : public sycl_cts::util::test_base {
...
+      auto buf = cl::sycl::buffer<int, 1>(result.data(), result.size());
       queue.submit([&](cl::sycl::handler &cgh) {
         auto acc = buf.get_access<cl::sycl::access::mode::read_write>(cgh);
         kernelWrapper(cgh, acc);
       });
     }
-    CHECK_VALUE_SCALAR(log, result, 10);
+    for (size_t i = 0; i < bufferSize; ++i) {
+      if (i < startIndex || i >= startIndex + numModified) {
+        if (!CHECK_VALUE(log, result[i], 12345, i)) return;
```
keryell, line 188: That sounds clear like an improvement.

keryell (approved):
Thanks for the paranoid improvement.

keryell (approved):

ProGTX (approved):

tests/handler/handler_invoke_api.cpp
```diff
@@ -30,18 +30,23 @@ using accessor_t =
                        cl::sycl::access::target::global_buffer>;
 
 struct single_task_functor {
-  single_task_functor(accessor_t acc) : acc(acc) {}
+  single_task_functor(accessor_t acc, size_t range) : acc(acc), range(range) {}
 
-  void operator()() { acc[0] = 10; }
+  void operator()() {
```
bader, line 35:
```suggestion
  void operator()() const {
```

I would appreciate if we can make all call operators constant to ensure compatibility with SYCL-2020.

psalz: Sure thing, done!

psalz:
I've marked all functor call operators as const, ensuring forward compatibility with SYCL 2020. I also reformatted the file using `clang-format-10` (previously it was formatted with `clang-format-6`).

I did some quick testing and am happy to report that ComputeCpp 2.2.1 now passes all tests using the host device, and DPC++ CUDA (a793a033) now passes as well.

bader (approved):

PR #55, opened 2020-08-20
Move stream half tests to a separate translation unit.

bader:


tests/stream/stream_api_fp16.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+  /** execute the test
+   */
+  void run(util::logger &log) override {
+    try {
+      // Check stream operator for cl::sycl::half
+      {
+        auto testQueue = util::get_cts_object::queue();
+
+        if (testQueue.get_device().has_extension("cl_khr_fp16")) {
```
ProGTX, line 37: Very minor, but it might be better to invert the logic and do an early return.

ProGTX (approved):

PR #57, opened 2020-09-03
Specify SYCL standard version for Intel implementation

bader:
Intel's implementation is about to switch default SYCL version to
SYCL-2020 and it will be a breaking change for test_header, which checks
if `CL_SYCL_LANGUAGE_VERSION` macro is defined by the compiler.


mkinsner (approved):

bader:
+@ProGTX

AerialMantis (approved):

bader:
Thanks for quick feedback.

PR #58, opened 2020-09-21
Improve copy tests

psalz:
In similar vein to #54, this PR improves the "handler_copy" tests to be much more robust, verifying that the various explicit memory operations actually do what they are supposed to do. Since these operations are highly configurable and thus prone to edge case bugs, I've spent considerable time on a solution that (I think) covers most if not all possible combinations.

Note that the test suite should probably be called "handler_explicit_memory_operations" or something similar, but I've decided to stick with the existing file name and terminology used within the test cases. So do keep in mind that while I'll mostly be talking about "copies" here, this usually also includes the `handler::fill` and to a lesser extent the `handler::update_host` function.

For a more complete changelog please see the commit messages, but here is the basic gist:

- Copies can now be strided (using ranged accessors) as well as transposed (for example, copying from a 2D buffer of shape [4,8] to a buffer of shape [8,4]).
- Copy results are being validated.
- Source buffers are filled with unique values for each index (a scalar encoding of the N-dimensional index) to make error detection more robust.
- Target buffers are filled with canary values to detect whether an implementation writes somewhere it shouldn't.
- Since there is a ton of test case permutations, making debugging a real challenge, I've also added a logging utility that can produce helpful error messages to pinpoint the exact template instantiation of a test case where something went wrong. Note that there are certain operations that are only parameterized by a subset of the various template parameters, which means that they are executed several times using the same configuration (this was however already the case previously).

With that being said, I think explicit memory operations are rather underspecified as of SYCL 1.2.1 Rev 7 (and not much has changed in SYCL 2020 Provisional, as far I can tell). This is apparently not just my opinion, because when you look at the results below it seems like implementors also have very different ideas as to what these functions should do. The tests I've implemented reflect my understanding of the spec, but it's entirely possible that I've made a wrong assumption somewhere. I've created a [corresponding issue in the SYCL-Docs repository](https://github.com/KhronosGroup/SYCL-Docs/issues/104), asking to clarify some of the underspecified points.

What this doesn't test yet is explicit memory operations on image accessors, and copying between buffers holding different data types.

## Results

I've done extensive testing to come up with these tables of the current state of implementors' support for explicit memory operations. Note that because there are _a lot_ of permutations to these test cases, I will only report results for a single data type (`long`), and a single writing access mode (`access::mode::write`). While results seem to be mostly the same across the different types and modes, I've seen some crashes for various combinations here and there -- while I tried to be careful, I won't rule out any mistakes on my part here.

Tested configurations:

- ComputeCpp 2.1 Intel OpenCL, Ubuntu 20.04 / Intel Xeon E5-4650
- ComputeCpp 2.1 NVIDIA OpenCL (experimental PTX), Ubuntu 18.04 / RTX 2070
- ComputeCpp 2.1 Host Device, Ubuntu 18.04
- hipSYCL CUDA (641d8990, Aug 22), Ubuntu 18.04 / RTX 2070 (CUDA 10.2)
  - As in #54, hipSYCL requires some additional workarounds within the CTS to compile at all (mostly related to OpenCL interop).
- LLVM (Intel SYCL / DPC++) OpenCL (14e227c4, Sep 16) Ubuntu 20.04 / Intel Xeon E5-4650
- LLVM (Intel SYCL / DPC++) CUDA (cb3a5d71, Sep 17) Ubuntu 18.04 / RTX 2070 (CUDA 10.2)
- LLVM (Intel SYCL / DPC++) Host Device (cb3a5d71, Sep 17) Ubuntu 18.04

### Copying

These are the results for `handler::copy` for various combinations of parameters. I'm differentiating between host-to-device (H2D) copies (`copy(ptr, accessor)`), device-to-device (D2D) copies (`copy(accessor, accessor)`) and device-to-host (D2H) copies (`copy(accessor, ptr)`). All variants can be dense (copying the full buffer) or strided (using a ranged accessor). For strided copies, the host pointer in H2D/D2H copies is always assumed to be dense (which is in accordance to my reading of the spec). D2D copies can additionally be transposed, where the shape of the source and destination accessor is mirrored (really only relevant in 2D/3D). On top of this, all copies are tested between all different combinations of source and destination dimensionalities (e.g., copying from a 2D accessor to a 3D accessor).

|           | H2D Dense          | H2D Strided        | D2D Dense            | D2D Strided          | D2D Transposed         | D2H Dense          | D2H Strided        |
| --------- | ------------------ | ------------------ | -------------------- | -------------------- | ---------------------- | ------------------ | ------------------ |
| CCPP OCL  | ✔                  | ❌<sup>CCPP1</sup> | ✔                    | ❌<sup>CCPP2</sup>   | ❌<sup>CCPP2</sup>     | ✔                  | ❌<sup>CCPP3</sup> |
| CCPP PTX  | ✔                  | ❌<sup>CCPP4</sup> | ✔                    | ❌<sup>CCPP2</sup>   | ❌<sup>CCPP2</sup>     | ✔                  | ❌<sup>CCPP4</sup> |
| CCPP Host | ✔                  | ❌<sup>CCPP5</sup> | ✔                    | ❌<sup>CCPP2</sup>   | ❌<sup>CCPP2</sup>     | ✔                  | ❌<sup>CCPP6</sup> |
| hipSYCL   | ✔                  | ✔                  | ✔<sup>HIPSYCL1</sup> | ✔<sup>HIPSYCL1</sup> | N/A<sup>HIPSYCL1</sup> | ✔                  | ✔                  |
| LLVM OCL  | ✔                  | ❌<sup>LLVM1</sup> | ✔                    | ❌<sup>LLVM2</sup>   | ✔                      | ✔                  | ❌<sup>LLVM3</sup> |
| LLVM CUDA | ✔                  | ❌<sup>LLVM4</sup> | ✔                    | ❌<sup>LLVM2</sup>   | ✔                      | ✔                  | ❌<sup>LLVM5</sup> |
| LLVM Host | ❌<sup>LLVM6</sup> | ❌<sup>LLVM7</sup> | ✔                    | ✔                    | ✔                      | ❌<sup>LLVM8</sup> | ❌<sup>LLVM3</sup> |

<sup>CCPP1</sup> Works for 1D. Segfaults for 2D / 3D.

<sup>CCPP2</sup> Only seems to work if `dim_src == dim_dst`. Segfaults for certain combinations of dimensions on Intel OCL.

<sup>CCPP3</sup> Appears to start copying from wrong offset.

<sup>CCPP4</sup> Works for 1D. Throws for 2D / 3D.

<sup>CCPP5</sup> Works for 1D. This appears to start to copy from the first host byte, but then assumes a stride on the host as well, ultimately reading beyond `accessor::get_size()` bytes.

<sup>CCPP6</sup> Works for 1D. This appears to start to copy from the first accessed element, but then assumes a stride on the host as well, ultimately writing beyond `accessor::get_size()` bytes.

<sup>HIPSYCL1</sup> hipSYCL does not support copying between buffers of different dimensionalities, or between differently shaped accessors (transposed copy). It also doesn't support `constant_buffer` accessors.

<sup>LLVM1</sup> Works for 1D, where the first byte pointed to by the host pointer is being copied to the first element at the specified offset (i.e., no offset is added to the host pointer, which is what I would expect). In 2D, the first host byte is also copied, but to the wrong location on the device. Not sure what is happening in 3D.

<sup>LLVM2</sup> Works for 1D. This is interesting, as dense, transposed and strided + transposed (not shown in table above) works. Likely an optimization gone wrong.

<sup>LLVM3</sup> Works for 1D. Writes directly to the first byte pointed to by the host pointer (i.e., again, no offset is added), however seems to start reading at the wrong location in the source accessor.

<sup>LLVM4</sup> Works for 1D and 2D. Not sure what is happening in 3D.

<sup>LLVM5</sup> Works for 1D, throws for 2D. Writes to first byte pointed to by the host pointer, but seems to start reading at wrong location in source accessor.

<sup>LLVM6</sup> Works for 1D, segfaults in 2D and produces wrong result in 3D.

<sup>LLVM7</sup> Works for 1D only.

<sup>LLVM8</sup> Works for 1D and 2D. Seems to start reading at wrong location in accessor.

### Fill / Update Host

These are the results for the (arguably somewhat less important) `handler::fill` and `handler::update_host` functions. For `fill`, im differentiating between dense fills, which fill the entire buffer, and ranged fills, which use a ranged accessor to only fill parts of a buffer with the specified value.

|           | Dense Fill | Ranged Fill | Update Host |
| --------- | ---------- | ----------- | ----------- |
| CCPP OCL  | ✔          | ❌          | ❌          |
| CCPP PTX  | ✔          | ❌          | ❌          |
| CCPP Host | ✔          | ❌          | ❌          |
| hipSYCL   | ✔          | ✔          | ✔           |
| LLVM OCL  | ✔          | ?*          | ✔           |
| LLVM CUDA | ✔          | ✔          | ✔           |
| LLVM Host | ✔          | ✔          | ✔           |

\* LLVM OCL currently crashes on me during linking, I haven't had time to look into this yet.

EDIT: There was a bug in the ranged fill test, in that it didn't create ranged accessors in the first place. The results above were updated after this was fixed.

**Key Takeaways:**

- As you can see, the current state of affairs surrounding `handler::copy` is rather dire, once one wants to go beyond a basic dense copy. However, for every test there is at least one implementation that behaves as I would expect. Full disclosure: I originally implemented support for `handler::copy` in hipSYCL, which might be a reason as to why it behaves as I would expect ;-).
- ~~None of the tested implementations supports a "ranged fill", i.e., calling `handler::fill` on a ranged accessor. Instead of only filling the accessed range, all tested implementations fill the entire buffer.~~ EDIT: This turned out to be a bug, see above.
- The CTS needs to test this.

### DIFF
--- a/tests/handler/handler_copy.cpp
+++ b/tests/handler/handler_copy.cpp
@@ -792,7 +792,8 @@ void test_write_acc_copy_functions(log_helper lh, cl::sycl::queue& queue) {
     copy_test_context<dataT, dim_src, dim_dst, strided, transposed> ctx(queue);
     ctx.verify_fill(
         [&](cl::sycl::handler& cgh) {
-          auto w = ctx.getDstBuf().template get_access<mode_dst, target>(cgh);
+          auto w = ctx.getDstBuf().template get_access<mode_dst, target>(
+              cgh, ctx.getDstCopyRange(), ctx.getDstCopyOffset());
           cgh.fill(w, pattern);
         },
         pattern,
--- a/tests/handler/handler_copy.cpp
+++ b/tests/handler/handler_copy.cpp
@@ -6,6 +6,11 @@
 //
 *******************************************************************************/
 
+#include <regex>
+#include <mutex>
+#include <sstream>
+#include <memory>
+
 #include "../common/common.h"
 
 #define TEST_NAME handler_copy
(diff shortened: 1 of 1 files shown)


illuhad:
Wow, thanks for all this work!

> None of the tested implementations supports a "ranged fill", i.e., calling handler::fill on a ranged accessor. Instead of only filling the accessed range, all tested implementations fill the entire buffer.

If this doesn't work in hipSYCL (at least with the new runtime) it's a bug. I agree the correct behavior is to only fill the ranged accessor, and there's code in hipSYCL to make it behave that way.

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,886 @@ using mode_t = cl::sycl::access::mode;
...
-      const auto pattern = dataT(117);
-      cgh.fill(accWrite, pattern);
-    };
-    TEST_SINGLE_COPY_FUNCTION_ACC_WRITE(dataT, dim_src, dim_dst, readMode,
-                                        writeMode, target, queue, func);
+    const auto pattern = type_helper<dataT>::make(117);
+    copy_test_context<dataT, dim_src, dim_dst, strided, transposed> ctx(queue);
+    ctx.verify_fill(
+        [&](cl::sycl::handler& cgh) {
+          auto w = ctx.getDstBuf().template get_access<mode_dst, target>(cgh);
```
sergey-semenov, line 795: Seems like there's a bug here: the accessor is created without range or offset, but `verify_fill` checks the results as if it was a ranged accessor.

psalz: Oof, you are right! That might offer an explanation for why all implementations wrote the full range...

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,887 @@ using mode_t = cl::sycl::access::mode;
...
-    return cl::sycl::accessor<dataT, dims, mode, target>(allocationSize, cgh);
+
+template <template <int dims> class T, size_t default_value>
+struct range_id_helper<T, 3, default_value> {
+  static T<3> make(size_t d0, size_t d1, size_t d2) { return T<3>{d0, d1, d2}; }
+
+  template <template <int> class other, int other_dims,
+            size_t dv = default_value>
+  static T<3> cast(const other<other_dims>& o) {
+    return T<3>{o[0], other_dims >= 2 ? o[1] : dv, other_dims == 3 ? o[2] : dv};
```
keryell, line 259: Naming a class "helper" does not really help the reader. :-)
Add a few comments to explain the high-level behavior.

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,887 @@ using mode_t = cl::sycl::access::mode;
...
+   *
+   * Note that we don't check any canary values outside of the updated region
+   * (when using a ranged accessor) as a SYCL implementation is free to also
+   * update other parts of the host memory.
+   */
+  template <typename test_fn>
+  void verify_update_host(test_fn fn, const log_helper& lh) const {
+    run_test_function(fn, lh);
+
+    std::lock_guard<cl::sycl::mutex_class> lock(srcBufHostMemoryMutex);
```
keryell, line 371: Why do you need this?

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,887 @@ using mode_t = cl::sycl::access::mode;
...
+      if (is_within_window(dstCopyOffset, dstCopyRange, dstAbsIdx)) {
+        // Compute relative linear index within destination copy range.
+        const size_t relLinearIdx =
+            compute_relative_linear_id(dstCopyOffset, dstCopyRange, dstAbsIdx);
+
+        // Now compute relative index in source copy range.
+        const auto srcRelIdx = reconstruct_index(srcCopyRange, relLinearIdx);
+
+        // Convert to absolute source index and use to compute expected value.
+        const auto expected =
```
keryell, line 442: Factorizing the 2 functions and having a lambda expressing the 2 line differences?

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,887 @@ using mode_t = cl::sycl::access::mode;
...
+template <typename dataT, int dim_src, int dim_dst, bool strided,
+          bool transposed>
+void test_all_write_acc_copy_functions(log_helper lh, cl::sycl::queue& queue) {
+  lh = lh.set_dim_src(dim_src).set_dim_dst(dim_dst);
+  constexpr auto target = target_t::global_buffer;
+
+  constexpr auto st = strided;
+  constexpr auto tr = transposed;
+
+  test_write_acc_copy_functions<dataT, dim_src, dim_dst, mode_t::read,
```
keryell, line 837:
```c++
for(auto s_m : { mode_t::read, mode_t::read_write })
  for (auto d_m : { mode_t::write, mode_t::read_write, mode_t::discard_write, mode_t::discard_read_write })
...
```

keryell (commented):
Quite impressive contribution!
Is this the basis for your next presentation "brute-forcing the SYCL memory API"? :-)

tests/handler/handler_copy.cpp
```diff
@@ -6,6 +6,11 @@
 //
 *******************************************************************************/
 
+#include <regex>
+#include <mutex>
+#include <sstream>
+#include <memory>
```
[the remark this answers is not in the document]
psalz: While I don't disagree, whenever a project provides a `.clang-format` file I just assume that everything is in order and don't manually check for things like this. Since I see you making this review on pretty much every PR, maybe you should consider setting [`SortIncludes: true`](https://github.com/KhronosGroup/SYCL-CTS/blob/26f29d2c8010125a85e8b215172a17f40424c8e8/.clang-format#L61)? :)

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,887 @@ using mode_t = cl::sycl::access::mode;
...
+
+ private:
+  util::logger* logger;
+  std::string dataType = "(unknown data type)";
+  int dimSrc = -1;
+  int dimDst = -1;
+  std::string modeSrc = "(unknown mode)";
+  std::string modeDst = "(unknown mode)";
+  std::string target = "(unknown target)";
+  int line = __LINE__;
```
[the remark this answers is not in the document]
psalz: What do you mean, the `line` or the defaults in general? Basically they're just some values that are never used in the current implementation, but if anyone should ever decide to modifyl/extent these tests, they'll be pointed to here if they do something I didn't anticipate.

tests/handler/handler_copy.cpp
```diff
@@ -17,334 +22,887 @@ using mode_t = cl::sycl::access::mode;
...
+  std::string info = "";
+  std::string pattern = "";
+
+  static std::string get_mode_string(mode_t mode) {
+    switch (mode) {
+      case mode_t::read:
+        return "read";
```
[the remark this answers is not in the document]
psalz: Whoa, that's pretty cool!

tests/handler/handler_copy.cpp
psalz: Well, I just followed the conventions of the previous implementation ;-). Sure, will do.

tests/handler/handler_copy.cpp
psalz: `verify_h2d_copy` and `verify_d2d_copy` you mean? I can look into it, sure!

tests/handler/handler_copy.cpp
psalz: Ah yes, I have "open issue about shared ownership host-buffer usage semantics in SYCL-Docs repo" somewhere in my backlog... It was my understanding from section 4.7.5 that the host memory region is only guaranteed to be updated once the mutex has been released by the SYCL runtime.

tests/handler/handler_copy.cpp
psalz: That would be nice, yes, but I'd need a `for constexpr` for that :-). Once the CTS upgrades to C++17 we can implement that!

ProGTX:
I've tested this branch with ComputeCpp and I got a few failures, as you've already reported.

However I'm also a bit concerned about the increased compilation times for just the handler tests - it went from 2 minutes to 20. I believe there was a point where we switched from templates to macros in order to keep the compilation times down, but now there are more templates again. Is there any chance we could optimize this?

tests/handler/handler_copy.cpp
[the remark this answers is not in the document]
keryell: It is! I use it pretty extensively on my triSYCL internal extensions.

tests/handler/handler_copy.cpp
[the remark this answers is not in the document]
keryell: The "problem" here is the `__LINE__` is 120 which you now about, not something that comes from the user, such as like https://en.cppreference.com/w/cpp/utility/source_location used with the default argument trick to inject it into user code.

tests/handler/handler_copy.cpp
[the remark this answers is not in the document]
keryell: Interesting point!
I guess you assume that everybody or some CI uses `clang-format`. :-)
It was added by @Ruyk on internal GitLab commit 42ee2db00faf8d31f2b665134312bd0092a2d87f Date:   Wed Aug 23 16:16:12 2017 +0100 and never been updated since.
Perhaps we have just discovered our new `clang-format` champion with this PR? :-)
Feel free to suggest a lot of improvements in this area.

tests/handler/handler_copy.cpp
keryell: Yes they look pretty much the same.

tests/handler/handler_copy.cpp
keryell: Oh right, I missed that the mode where used in some template arguments. :-(

keryell:
> However I'm also a bit concerned about the increased compilation times for just the handler tests - it went from 2 minutes to 20. 

Interesting.
Are the tests executed in parallel in such a way we put all the lengthy tests first, for example by sorting by estimated execution time as an heuristic to minimize global time execution?

tests/handler/handler_copy.cpp
keryell: Yes, I know... Thank you for making the world better and more understandable. :-)

tests/handler/handler_copy.cpp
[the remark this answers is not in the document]
keryell: > While I don't disagree, whenever a project provides a `.clang-format` file I just assume that everything is in order and don't manually check for things like this. Since I see you making this review on pretty much every PR, maybe you should consider setting [`SortIncludes: true`](https://github.com/KhronosGroup/SYCL-CTS/blob/26f29d2c8010125a85e8b215172a17f40424c8e8/.clang-format#L61)? :)

Feel free to change it. :-)
I cannot believe that our `.clang-format` is so complex and different from a standard one...

keryell (approved):
Thanks.

psalz:
Sorry for being slow on this one; I haven't tested with the newest versions of DPC++ / ComputeCPP, but I expect the compile times to still be prohibitively long. I think the best way to move forward with this would be to retarget it for SYCL 2020, which should hopefully reduce some of the complexity (and thus drive down compile times). I haven't had time to thoroughly review the new spec regarding explicit copy operations though.

tests/handler/handler_copy.cpp
keryell: The nice side effect of procrastination is that you get C++17... :-)

bader: NOTE: this PR targets SYCL-1.2.1, which is based on C++11. This suggestion can be applied in `SYCL-2020 branch` though.

psalz:
Okay, SYCL 2020 didn't really change much regarding explicit copy operations (seems like I need to do some more lobbying internally ;-)), so we might as well go ahead and merge this for 1.2.1.

I've pulled #73 into `SYCL-1.2.1/master`, rebased this onto it, and am now using the `SYCL_CTS_ENABLE_EXTENSIVE_MODE` macro to decide whether to test for all the data types we've had so far, or just three. Of course a question now becomes which types to test in the non-extensive / default case. I've opted for `int`, `double` and `double16`, but I'm open to suggestions.

I've additionally moved everything but the test class into an unnamed namespace, and marked all functions as `static`. Here's some non-scientific timings I've obtained:

|      | SYCL-1.2.1/master | This PR (default) | This PR (extensive) |
| ---- | ----------------- | ----------------- | ------------------- |
| LLVM | 7:44              | 4:02              | 32:04               |
| CCPP | 0:48              | 1:16              | 4:16                |

These are in `minutes:seconds`. As you can see, at least on my machine, LLVM is a lot slower. Somewhat surprisingly though, the compile time actually goes down with this PR, when using the default configuration. For the extensive configuration, ComputeCpp sees a 4x increase in compile time, which makes sense to me, as we move from 3 to 12 tested types. LLVM outright explodes with a 8x time increase.

Tested with LLVM@98505e40 and CCPP 2.4.

(By the way: I always have to set the `CMAKE_CXX_STANDARD` standard to C++14 manually to compile with LLVM/DPC++, what's the deal with that?).

bader:
> As you can see, at least on my machine, LLVM is a lot slower.

Do you set optimization level for DPC++ using `INTEL_SYCL_FLAGS` CMake variable?
ComputeCPP uses [-O2 by default](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindComputeCpp.cmake#L80), whereas DPC++ doesn't set optimization level by default. This might be the reason for performance difference.

> By the way: I always have to set the `CMAKE_CXX_STANDARD` standard to C++14 manually to compile with LLVM/DPC++, what's the deal with that?

DPC++ development team is working on adding SYCL 2020 features and many of them require C++17, so I expect DPC++ will require C++17 soon (if it doesn't require it already).

psalz:
> Do you set optimization level for DPC++ using `INTEL_SYCL_FLAGS` CMake variable?
> ComputeCPP uses [-O2 by default](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindComputeCpp.cmake#L80), whereas DPC++ doesn't set optimization level by default. This might be the reason for performance difference.

Sorry, I should've been more explicit: These results are for compile time, not run time. I did not set the optimization level, but I would expect it to only negatively impact these timings, if at all. I can do some more testing tomorrow, however.

bader:
> > Do you set optimization level for DPC++ using `INTEL_SYCL_FLAGS` CMake variable?
> > ComputeCPP uses [-O2 by default](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindComputeCpp.cmake#L80), whereas DPC++ doesn't set optimization level by default. This might be the reason for performance difference.
> 
> Sorry, I should've been more explicit: These results are for compile time, not run time. I did not set the optimization level, but I would expect it to only negatively impact these timings, if at all. I can do some more testing tomorrow, however.

I need more clarification still :). Do you include the "jit compile" time to "compile time" or it's only the time to build the executable?
What you can also try to do is to change `-fsycl-device-code-split=per_source` to `-fsycl-device-code-split=per_kernel`. I haven't tested it myself, but it might help to reduce the time spent on translating to SPIR-V format. https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindIntel_SYCL.cmake#L30
I'm not that familiar with ComputeCPP compilation flow, but IIRC, it produces SPIR 1.2 by default. On the other hand, I don't expect this difference to cause 4-8 times difference. Philip, do you build Release version of DPC++ compiler toolchain?

psalz:
> Do you include the "jit compile" time to "compile time" or it's only the time to build the executable?

Only plain compilation. Basically `time ninja test_handler`.

> What you can also try to do is to change `-fsycl-device-code-split=per_source` to `-fsycl-device-code-split=per_kernel`. I haven't tested it myself, but it might help to reduce the time spent on translating to SPIR-V format.

I tried this, but it doesn't seem to have any significant effect. From what I can tell (by grepping for `-fsycl-is-device` and `-fsycl-is-host` in my process tree), most of the time is spent on the compilation, not linking, and about equal time is spent compilation for device and host. I also tried passing `-fno-sycl-early-optimizations`, but that had an adverse effect, causing build times of over an hour.

> On the other hand, I don't expect this difference to cause 4-8 times difference.

I agree, and would appreciate if someone could try and replicate this, to see if something is fishy with my setup.

> Philip, do you build Release version of DPC++ compiler toolchain?

Yes, I configured my build like this: `./buildbot/configure.py -s . -o build -t Release --cuda`. This was a while ago however, and to be sure, I just did a clean rebuild -- with similar results.

psalz:
Update: I installed the DPC++ compiler directly from the oneAPI toolkit (version 2020.10.0.1113). Here I'm getting compile times much closer to ComputeCpp: **1:33 for the default** and **6:03 for the extended variant**. I'd like to know where this difference is coming from, but then again that is somewhat ouf of scope for this PR.

In any case, I'd say the compile times are reasonable now, and I think we can go ahead and merge once #73 has been finalized.

bader:
Interesting... I see even slower compilation time than 7:44 when using tip of the `sycl` branch from `intel/llvm`.
What CPU do you use for compile time measurements?

> Update: I installed the DPC++ compiler directly from the oneAPI toolkit (version 2020.10.0.1113). Here I'm getting compile times much closer to ComputeCpp: **1:33 for the default** and **6:03 for the extended variant**.

What is the test execution status? Does the test compiled with oneAPI toolkit version pass?

psalz:
> Interesting... I see even slower compilation time than 7:44 when using tip of the `sycl` branch from `intel/llvm`.
> What CPU do you use for compile time measurements?

Unless noted otherwise my measurements are done on a Ryzen Threadripper 2920X. However I just went ahead and did some testing on an (ancient) Xeon E5-4650. For the default (non-extended) variant, a current build of `intel/llvm` requires 6:35, while the prebuilt one from the toolkit requires 2:47 to compile.

> What is the test execution status? Does the test compiled with oneAPI toolkit version pass?

I just did some preliminary testing on the aforementioned Xeon CPU, using the default variant, and the results depend on the platform: When running with `-p host -d host`, both the prebuilt and current manual build pass all tests for `int` and `double`, but fail on `update_host` for `double16` -- which is in line with what I reported above on Ocotber 30th. When running with `-p intel -d opencl_cpu`, the manually built version behaves the same way (only failing for `update_host` with `double16`). The prebuilt version however fails a lot more tests. I didn't parse the results in detail, but it looks like it is also failing various `copy` tests for `int` and `double`,

fraggamuffin:
rebase then merge
LGTM

bader:
@psalz, I just realized that #73 is merged to `SYCL-2020` branch whereas this PR is targeting `SYCL-1.2.1/master`.

psalz:
@bader Hm, will it be backported, or should I retarget this for 2020?

bader:
I would prefer `SYCL-2020`, but I'm okay to have it in both branches.
@ProGTX, @keryell, do you have any preferences?

keryell:
Just targeting SYCL 2020 is enough. It does not make sense to spend a lot of time improving SYCL 1.2.1 CTS.

psalz:
Okay I've retargeted this against `SYCL-2020` and changed the extensive mode macro to `SYCL_CTS_FULL_CONFORMANCE`. Should be good to go!

bader:
Great. Thanks!

PR #59, opened 2020-10-02
Extend type coverage for relational and integer functions

kopylovanat:
Fixed ommiting tests for char and vec<char,N>

Remaining gap (test coverage): result correctness isn't checked

### DIFF
--- a/tests/math_builtin_api/modules/test_generator.py
+++ b/tests/math_builtin_api/modules/test_generator.py
@@ -156,6 +156,7 @@ def generate_test_cases(test_id, types, sig_list):
 def attempt_match(runner, var_type, base_type, dim, unsigned, current_type):
     ct_part_keys = []
     for t in current_type.keys():
+        # Matching by all keys except name
         ct_part_keys.append(t[:-1])
     # Change sign.
     if (var_type, base_type, dim, not unsigned) in ct_part_keys:
@@ -181,9 +182,15 @@ def attempt_match(runner, var_type, base_type, dim, unsigned, current_type):
 def expand_signature(runner, types, signature):
     current_types = [types[signature.ret_type]]
 
+    # to control cases when some arg types are base type of ret type
+    sgeninteger = False
     exp_sig = []
     for arg in signature.arg_types:
         current_types.extend([types[arg]])
+        if arg == "sgeninteger":
+            sgeninteger = True
+
+    name_key_index = 4
 
     # Iterate over all basic types
     # Try to match all function types (return type and argument types).
@@ -194,15 +201,16 @@ def expand_signature(runner, types, signature):
         index = 0
         for ct in current_types:
             ct_keys = ct.keys()
-            ct_part_keys = []
+            name_keys = []
             for t in ct_keys:
-                ct_part_keys.append(t[4:])
-            if (name, ) in ct_part_keys:
+                # Matching by name
+                name_keys.append(t[name_key_index])
+            if name in name_keys:
                 match = True
             else:
                 nomatch.append(index)
             index += 1
-        if match is True:
+        if match:
             # Return value and all arguments are of the same type.
             if len(nomatch) == 0:
                 new_sig = sycl_functions.funsig(signature.namespace, name, signature.name, [
@@ -215,14 +223,29 @@ def expand_signature(runner, types, signature):
                 extra = False
                 for ct in current_types:
                     ct_keys = ct.keys()
-                    ct_part_keys = []
+                    name_keys = []
                     for t in ct_keys:
-                        ct_part_keys.append(t[4:])
+                        name_keys.append(t[name_key_index])
                     # Current function type already matches - no need for mutation.
-                    if (name, ) in ct_part_keys:
+                    if name in name_keys:
                         function_types.append(name)
                         function_types_extra.append(name)
                     else:
+                        # Return value and all arguments are of the same base type.
+                        # Types are selected by scalar/vector, size, dim and unsigned/signed
+                        # but these parameters are the same for signed char and char.
+                        # It works well when we need to select corresponding signed type for unsigned char
+                        # but in cases of functions such as clamp where we have parameters like (vec<T,N>, T..)
+                        # we shouldn't test it with paramenters (scharN, char..) or (charN, signed char) and we need special case for it
+                        if sgeninteger:
+                            part_keys = ('scalar', all_types[name].base_type, 1, all_types[name].unsigned)
+                            arg_types = [ct[key] for key in ct.keys() if key[:-1] == part_keys]
+                            if len(arg_types) > 0:
+                                if len(arg_types) > 1 and 'schar' in name:
+                                    function_types.append(arg_types[1].name)
+                                else:
+                                    function_types.append(arg_types[0].name)
+                                continue
                         var_type = all_types[name].var_type
                         base_type = all_types[name].base_type
                         dim = all_types[name].dim
@@ -230,7 +253,7 @@ def expand_signature(runner, types, signature):
 
                         # Get var_type, base_type, dim, unsigned of suitable type
                         mutation = attempt_match(runner, var_type, base_type, dim, unsigned, ct)
-                        if mutation is not None:
+                        if mutation:
                             # Find suitable type by all key members except name
                             new_types = [ct[key] for key in ct.keys() if key[:-1] == mutation]
                             function_types.append(new_types[0].name)
@@ -239,7 +262,7 @@ def expand_signature(runner, types, signature):
                                 extra = True
                         else:
                             all_matched = False
-                if all_matched is True:
+                if all_matched:
                     new_sig = sycl_functions.funsig(
                         signature.namespace, function_types[0], signature.name, function_types[1:], signature.pntr_indx[:])
                     exp_sig.append(new_sig)
--- a/tests/math_builtin_api/modules/test_generator.py
+++ b/tests/math_builtin_api/modules/test_generator.py
@@ -1,4 +1,5 @@
 from . import sycl_functions
+from . import sycl_types
 import random
 
 test_case_templates = { "private" : ("\n\n{\n"
@@ -153,22 +154,26 @@ def generate_test_cases(test_id, types, sig_list):
 # We attempt to find a combination that is part of the current generic type (e.g. floatn)
 
 def attempt_match(runner, var_type, base_type, dim, unsigned, current_type):
+    ct_keys = list(current_type.keys())
+    ct_part_keys = []
+    for t in ct_keys:
+        ct_part_keys.append(t[:-1])
     # Change sign.
-    if (var_type, base_type, dim, not unsigned) in list(current_type.keys()):
+    if (var_type, base_type, dim, not unsigned) in ct_part_keys:
         return (var_type, base_type, dim, not unsigned)
     # Change type and sign, same dimensionality.
     for new_sign in [unsigned, not unsigned]:
         for new_type in runner.base_types:
-            if (var_type, new_type, dim, new_sign) in list(current_type.keys()):
+            if (var_type, new_type, dim, new_sign) in ct_part_keys:
                 return (var_type, new_type, dim, new_sign)
     # Change to base type (for scalars).
     for new_sign in [unsigned, not unsigned]:
-        if ("scalar", base_type, 1, new_sign) in list(current_type.keys()):
+        if ("scalar", base_type, 1, new_sign) in ct_part_keys:
             return ("scalar", base_type, 1, new_sign)
     # See if any other scalar type match.
     for new_sign in [unsigned, not unsigned]:
         for new_type in runner.base_types:
-            if ("scalar", new_type, 1, new_sign) in list(current_type.keys()):
+            if ("scalar", new_type, 1, new_sign) in ct_part_keys:
                 return ("scalar", new_type, 1, new_sign)
     return None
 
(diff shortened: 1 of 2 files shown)


bader:
@ProGTX, @illuhad, could you take a look, please?

bader (approved):

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -153,22 +154,26 @@ def generate_test_cases(test_id, types, sig_list):
 # We attempt to find a combination that is part of the current generic type (e.g. floatn)
 
 def attempt_match(runner, var_type, base_type, dim, unsigned, current_type):
+    ct_keys = list(current_type.keys())
+    ct_part_keys = []
+    for t in ct_keys:
```
keryell, line 159: Would this work?
```suggestion
    for t in current_type.keys():
```

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +186,71 @@ def expand_signature(runner, types, signature):
...
-                    if match is True:
-                        # Return value and all arguments are of the same type.
-                        if len(nomatch) == 0:
-                            common_type = current_types[0][(
-                                var_type, base_type, dim, unsigned)]
-                            new_sig = sycl_functions.funsig(signature.namespace, common_type.name, signature.name, [
-                                                            common_type.name for i in range(len(signature.arg_types))], signature.pntr_indx[:])
-                            exp_sig.append(new_sig)
+    all_types = sycl_types.create_basic_types()
+    for name in list(all_types.keys()):
```
keryell, line 192:
```suggestion
    for name in all_types.keys():
```
?

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +186,71 @@ def expand_signature(runner, types, signature):
...
+        match = False
+        nomatch = []
+        index = 0
+        for ct in current_types:
+            ct_keys = list(ct.keys())
```
keryell, line 197: Needed?

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +186,71 @@ def expand_signature(runner, types, signature):
...
+                new_sig = sycl_functions.funsig(signature.namespace, name, signature.name, [
+                                                name for i in range(len(signature.arg_types))], signature.pntr_indx[:])
+                exp_sig.append(new_sig)
+            else:
+                function_types = []
+                function_types_extra = []
+                all_matched = True
+                extra = False
+                for ct in current_types:
+                    ct_keys = list(ct.keys())
```
keryell, line 218: Needed?

keryell (commented):
Great!

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -1,4 +1,5 @@
 from . import sycl_functions
+from . import sycl_types
```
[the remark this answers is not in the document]
kopylovanat: It's for getting all_types = sycl_types.create_basic_types()

tests/math_builtin_api/modules/test_generator.py
kopylovanat: Yes, it works, thank you! Removed list in other places too.

tests/math_builtin_api/modules/test_generator.py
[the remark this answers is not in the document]
keryell: Oh I see, there is a `__init__.py` in the directory.
I have never used `.` like that. https://docs.python.org/3/tutorial/modules.html#intra-package-references

keryell (approved):
Thanks.

bader:
@AerialMantis, @ProGTX, @illuhad, I'd like to get at least one more approval from Codeplay team or Aksel to merge this PR.
The same applies to two more PRs: https://github.com/KhronosGroup/SYCL-CTS/pull/60 and https://github.com/KhronosGroup/SYCL-CTS/pull/61.

bader:
> @AerialMantis, @ProGTX, @illuhad, I'd like to get at least one more approval from Codeplay team or Aksel to merge this PR.
> The same applies to two more PRs: #60 and #61.

Friendly ping.

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +185,71 @@ def expand_signature(runner, types, signature):
...
+            ct_keys = ct.keys()
+            ct_part_keys = []
+            for t in ct_keys:
+                ct_part_keys.append(t[4:])
+            if (name, ) in ct_part_keys:
+                match = True
+            else:
+                nomatch.append(index)
+            index += 1
+        if match is True:
```
ProGTX, line 205:
```suggestion
        if match:
```

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +185,71 @@ def expand_signature(runner, types, signature):
...
+    for name in all_types.keys():
+        match = False
+        nomatch = []
+        index = 0
+        for ct in current_types:
+            ct_keys = ct.keys()
+            ct_part_keys = []
+            for t in ct_keys:
+                ct_part_keys.append(t[4:])
+            if (name, ) in ct_part_keys:
```
ProGTX, line 200: I'm unsure about his, why not just:
```suggestion
            if name in ct_part_keys:
```

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +185,71 @@ def expand_signature(runner, types, signature):
...
+    all_types = sycl_types.create_basic_types()
+    for name in all_types.keys():
+        match = False
+        nomatch = []
+        index = 0
+        for ct in current_types:
+            ct_keys = ct.keys()
+            ct_part_keys = []
+            for t in ct_keys:
+                ct_part_keys.append(t[4:])
```
ProGTX, line 199: Why `4`? Could this please have a comment, or maybe be stored in a named variable?

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +185,71 @@ def expand_signature(runner, types, signature):
...
+                        function_types_extra.append(name)
+                    else:
+                        var_type = all_types[name].var_type
+                        base_type = all_types[name].base_type
+                        dim = all_types[name].dim
+                        unsigned = all_types[name].unsigned
+
+                        # Get var_type, base_type, dim, unsigned of suitable type
+                        mutation = attempt_match(runner, var_type, base_type, dim, unsigned, ct)
+                        if mutation is not None:
```
ProGTX, line 233:
```suggestion
                        if mutation:
```

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +185,71 @@ def expand_signature(runner, types, signature):
...
-                                        all_matched = False
-                            if all_matched is True:
-                                new_sig = sycl_functions.funsig(
-                                    signature.namespace, function_types[0], signature.name, function_types[1:], signature.pntr_indx[:])
-                                exp_sig.append(new_sig)
-                            else:
-                                print("[WARNING] Unable to fully match function " + signature.name + " for: " +
-                                      var_type + ", " + base_type + ", " + str(dim) + ", " + str(unsigned))
+                            all_matched = False
+                if all_matched is True:
```
ProGTX, line 242:
```suggestion
                if all_matched:
```

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -181,52 +185,71 @@ def expand_signature(runner, types, signature):
...
+            else:
+                function_types = []
+                function_types_extra = []
+                all_matched = True
+                extra = False
+                for ct in current_types:
+                    ct_keys = ct.keys()
+                    ct_part_keys = []
+                    for t in ct_keys:
+                        ct_part_keys.append(t[4:])
```
ProGTX, line 220: There's another magic number here, I think a common variable is the best bet then

tests/math_builtin_api/modules/test_generator.py
```diff
@@ -153,22 +154,25 @@ def generate_test_cases(test_id, types, sig_list):
 # We attempt to find a combination that is part of the current generic type (e.g. floatn)
 
 def attempt_match(runner, var_type, base_type, dim, unsigned, current_type):
+    ct_part_keys = []
+    for t in current_type.keys():
+        ct_part_keys.append(t[:-1])
```
ProGTX, line 159: Can a comment please be added indicating why we skip the last value?

ProGTX (changes requested):
Sorry for not reviewing this earlier. The changes mostly seem like a welcome addition, I've left some suggestions.

What's more worrying is that ComputeCpp fails to compile the newly generated file, specifically this is where the first error happens:
```cpp
cl::sycl::char2 inputData_0(58,-109);
signed char inputData_1(90);
return cl::sycl::max(inputData_0,inputData_1);
```

I'm not sure if the SYCL spec supports this use case, or whether it should. `std::max` will also fail to compile if you pass in a `char` and a `signed char`. Even if that were defined, I think the result might depend on whether `char` itself is signed or not. Does this work in DPC++, though?

I don't know exactly which builtins are impacted by this, but it seems like there might need to be more restrictions when generating the file.

bader:
>Even if that were defined, I think the result might depend on whether `char` itself is signed or not.

Unless I missed something, it seems to me that the spec doesn't allow ambiguity for this case. According to the SYCL-1.2.1 specification: 

> `cl_char` - Alias to a signed 8-bit integer, as defined bythe C++11 ISO Standard

`cl::sycl::cl_char2` is expected to be an alias to 2-element vector of `signed int8_t` or `signed char` (if `char` == `int8_t`).

keryell:
> > Even if that were defined, I think the result might depend on whether `char` itself is signed or not.
> 
> Unless I missed something, it seems to me that the spec doesn't allow ambiguity for this case. According to the SYCL-1.2.1 specification:
> 
> > `cl_char` - Alias to a signed 8-bit integer, as defined bythe C++11 ISO Standard
> 
> `cl::sycl::cl_char2` is expected to be an alias to 2-element vector of `int8_t` or `signed char` (if `sizeof(char)` == `sizeof(int8_t)`).

I think you are right but then why not using `std::int8_t` to be clearer and to fail in the rare case a `signed char` has not 8 bits?

ProGTX:
I combed through the 1.2.1 and 2020 provisional specs and discussed this internally a bit.

This is what the spec (both of them) says:
```cpp
geninteger max (geninteger x, geninteger y)
geninteger max (geninteger x, sgeninteger y)
```
> The built-in functions can take as input char, unsigned char, short, unsigned short, int, unsigned int, long long int,
unsigned long long int and their vec and marray counterparts.

This is the conclusion we came to:
1. SYCL builtins are defined in terms of C++ types, not OpenCL ones
2. `max(scalar, vector)` is probably fine, though that might need a clarification in the spec
3. `max(signed, unsigned)` is fine
4. `max(char, (un)signed char)` is not allowed because the result would not be portable
5. The spec doesn't even mention `signed char` for the builtins

So in theory this patch wouldn't be correct because of point 5 alone. But we believe that to be an oversight, so we might want to fix the spec.

Point 4 is the problematic one, though. Specifically the example I posted already gives different results on X86 and ARM, for example.

bader:
> 3. `max(signed, unsigned)` **is fine**
> 5. The spec **doesn't even mention** `signed char` for the builtins

Are there any conflicts between these items?

ProGTX:
I meant `max(signed, unsigned)` for the generic case, so `max(int, unsigned int)` should be fine, and I believe `max(signed char, unsigned char)` would be as well if the spec mentioned `signed char`. `char` is a distint type from both, though, so I don't think there's a conflict there.

bader:
> I meant `max(signed, unsigned)` for the generic case, so `max(int, unsigned int)` should be fine, and I believe `max(signed char, unsigned char)` would be as well if the spec mentioned `signed char`. `char` is a distint type from both, though, so I don't think there's a conflict there.

I agree. Thanks for clarification.

kopylovanat:
@ProGTX thank you for comments! Regarding `max` with `char` issue:

> 2. `max(scalar, vector)` is probably fine, though that might need a clarification in the spec

`max(vector, scalar)` is fine, because it corresponds to `max (geninteger x, sgeninteger y)`
> 3. `max(signed, unsigned)` is fine

From my understanding it's not, because in `max (geninteger x, geninteger y)` x and y types should be the same and in `max (geninteger x, sgeninteger y)` x and y types should be vector and scalar with the same base type.
> 4. `max(char, (un)signed char)` is not allowed because the result would not be portable

And this case also shouldn't be allowed with the same reasoning as for 3.
> 5. The spec doesn't even mention `signed char` for the builtins
> 
> So in theory this patch wouldn't be correct because of point 5 alone. But we believe that to be an oversight, so we might want to fix the spec.

It's too seems to me as spec for integer functions missed `signed char` because generic type `geninteger` contains `signed char` according to Table 4.109

> 
> Point 4 is the problematic one, though. Specifically the example I posted already gives different results on X86 and ARM, for example.

Example failed because it has `vec<char,2>` and `signed char` as arguments, so it's vector and scalar with different base types. And it's shouldn't be tested (even though it works on DPC++ because `char` and `signed char` are treated as the same). I'll modify generation code to stop these cases from being generated.

tests/math_builtin_api/modules/test_generator.py
kopylovanat: Done.

tests/math_builtin_api/modules/test_generator.py
kopylovanat: Done.

bader:
@ProGTX, could you check if https://github.com/KhronosGroup/SYCL-CTS/pull/59/commits/95d4c475decae999cc49cb38de30354f069ea099 addresses your comments, please? Let me know if any other changes or other actions are required.

ProGTX (approved):
You're right about `max(signed, unsigned)` not being valid and the implications of that.

The code works now with ComputeCpp, thanks for the changes.

PR #63, opened 2020-10-16
Improve test coverage for nd_item and group constructors and equality operators

yuriykoch:
Improve test coverage for nd_item and group constructors and equality operators

Fixed gaps:
  - removed possible false negatives for constructors and assignment operators
    due verification chaining
  - added const-correctness check and verification for host for constructors,
    assignment and equality operators
  - added symmetric and transitive requirements check for device for equality
    operators


yuriykoch:
@illuhad @keryell @ProGTX, could you possibly take a look, please?

tests/common/common_by_value.h
```diff
@@ -0,0 +1,242 @@
+/*******************************************************************************
+//
+//  SYCL 1.2.1 Conformance Test Suite
+//
+//  Copyright:	(c) 2017 by Codeplay Software LTD. All Rights Reserved.
```
keryell, line 5: I do not know how to update this when it comes from a file which is later improved by someone else like here...
IANAL @Ruyk

tests/common/common_by_value.h
```diff
@@ -0,0 +1,242 @@
...
+*******************************************************************************/
+
+#ifndef __SYCLCTS_TESTS_COMMON_COMMON_BY_VALUE_H
+#define __SYCLCTS_TESTS_COMMON_COMMON_BY_VALUE_H
+
+#include "common.h"
+
+#include <string>
+#include <array>
+#include <type_traits>
```
keryell, line 18: Sort.

tests/common/common_by_value.h
```diff
@@ -0,0 +1,242 @@
...
+
+namespace {
+
+/**
+ * @brief Check equality-comparable operations on the host side
+ */
+template <typename T>
+void check_equality_comparable_generic(sycl_cts::util::logger& log, const T& a,
+                                       std::string test_name) {
```
keryell, line 27:
```suggestion
                                       const std::string& test_name) {
```

tests/common/common_by_value.h
```diff
@@ -0,0 +1,242 @@
...
+  /** check for reflexivity
+   */
+  if (!(a == a)) {
+    FAIL(log, (test_name +
+               " is not equality-comparable (operator== reflexivity failed)")
+                  .c_str());
```
keryell, line 33: This `FAIL` should be updated to take also `std::string` some day...

tests/group/group_common.h
```diff
@@ -0,0 +1,49 @@
...
+    std::array<cl::sycl::group<numDims>, numItems>& items)
+{
+  using item_t = cl::sycl::group<numDims>;
+
+  const auto oneElemRange =
+      sycl_cts::util::get_cts_object::range<numDims>::get(1, 1, 1);
+  const auto itemRange =
+      sycl_cts::util::get_cts_object::range<numDims>::get(numItems, 1, 1);
+
+  cl::sycl::buffer<item_t> itemBuf(items.data(),
```
keryell, line 32: I was somehow puzzled by this.
Actually it works since the data in a buffer is not initialized and so there is no need to have a default constructor.

tests/group/group_common.h
```diff
@@ -0,0 +1,49 @@
...
+#include "../common/common.h"
+#include <array>
+
+namespace {
+
+/** Retrieve group objects and store them
+ */
+template <typename kernelT, int numDims, size_t numItems>
+void store_group_instances(
+    std::array<cl::sycl::group<numDims>, numItems>& items)
```
keryell, line 23: I am unsure it works since `cl::sycl::group<numDims>` does not have a public default constructor.

tests/group/group_common.h
```diff
@@ -0,0 +1,49 @@
...
+    cgh.parallel_for_work_group<kernelT>(
+        itemRange, oneElemRange,
+        [=](item_t group) { itemAcc[group.get_linear_id()] = group; });
+  });
+  queue.wait_and_throw();
+}
+
+} //namespace
+
+#endif // __SYCLCTS_TESTS_GROUP_GROUP_COMMON_H
```
keryell, line 49: Lacking end of line.

tests/group/group_constructors.cpp
```diff
@@ -91,53 +142,114 @@ class TEST_NAME : public util::test_base {
...
+        testQueue.wait_and_throw();
+      }
+
+      // Check on the host side only if copy assignment works as expected
+      if (success[to_integral(current_check::copy_assignment)]) {
+        // group is not default constructible, store two objects into the array
+        static constexpr size_t numItems = 2;
+        using item_array_t = std::array<cl::sycl::group<numDims>, numItems>;
+        char rawItems[sizeof(item_array_t)];
+        auto& items = *reinterpret_cast<item_array_t*>(rawItems);
```
keryell, line 203: OK, now I see the hack to create a `std::array` without a default constructor... :-)

I am worried about alignment constraints on `item_array_t` here that might not be the same as a byte in all the ABIs.

keryell: Do we say in the spec that these objects are trivially copyable, by the way?

So I am not sure we can use accessors of `group` for example...

tests/nd_item/nd_item_common.h
```diff
@@ -0,0 +1,51 @@
...
+
+  auto queue = sycl_cts::util::get_cts_object::queue();
+  queue.submit([&](cl::sycl::handler& cgh) {
+    auto itemAcc =
+        itemBuf.template get_access<cl::sycl::access::mode::write>(cgh);
+
+    cgh.parallel_for<kernelT>(
+        cl::sycl::nd_range<numDims>(itemRange, oneElemRange),
+        [=](item_t item) {
+          itemAcc[item.get_global_linear_id()] = item;
```
keryell, line 43: Could it possible to factorize the similar files to this one?

keryell (commented):
Thank you for working on this and sorry for the delay.

tests/common/common_by_value.h
yuriykoch: Done by 340c849068068

tests/common/common_by_value.h
yuriykoch: Thank you for noticing. I've made changes by 9d88b50c4adc1c9351e30362077ded9caabc36e5. Also please take a look on fail_proxy changes by 4d664edf45fcf40ea9ae9492f1764f4e94db7b84

tests/group/group_common.h
yuriykoch: File was removed by 5a98c20c87bd68dca67b7d15faa8237656ca6391

tests/nd_item/nd_item_common.h
yuriykoch: I've moved out common code into the single file by 5a98c20c87bd68dca67b7d15faa8237656ca6391

tests/group/group_constructors.cpp
bader: > Each of the following SYCL runtime classes: id, range, item, nd_item, h_item, group, sub_group and nd_range
> must follow the following statements, where T is the runtime class type:
> • T must be default copy constructible and copy assignable on the host application (in the case where T is
> available on the host) and within SYCL kernel functions.
> • T must be default destructible on the host application (in the case where T is available on the host) and
> within SYCL kernel functions.
> • T must be default move constructible and default move assignable on the host application (in the case where
> T is available on the host) and within SYCL kernel functions.
> • T must be equality comparable on the host application (in the case where T is available on the host) and
> within SYCL kernel functions. Equality between two instances of T (i.e. a == b) must be true if the value
> of all members are equal and non-equality between two instances of T (i.e. a != b) must be true if the
> value of any members are not equal, unless either instance has become invalidated by a move operation. By
> extension of the requirements above, equality on T must guarantee to be reflexive (i.e. a == a), symmetric
> (i.e. a == b implies b == a and a != b implies b != a) and transitive (i.e. a == b && b == c implies
> c == a).

This seems to imply "trivially copiable" requirement.

NOTE: existing code already uses "the hack", so no new requirements are enforced. @keryell, I suggest filing an separate issue, if this question requires additional clarification.

yuriykoch: Great catch regarding alignment. I've fixed it by 5ef2c768737b0cc7236c46f6c0984097fc47b5ca.
I've also moved all the storage logic to the same place by 5a98c20c87bd68dca67b7d15faa8237656ca6391, so may be it will be less confusing to read.

tests/common/common_by_value.h
yuriykoch: Since this is new file created I removed this notice for now. Please let me know if any other action is required.

ProGTX (approved):

bader:
@keryell, could you take a look at the updates, please?

ProGTX (approved):

illuhad (approved):

bader:
@ProGTX, @illuhad, thanks a lot for reviews.
@keryell, I'd like to make sure that all your comments are addressed before merging this PR.

keryell (approved):
That looks good.
Thanks!

PR #64, opened 2020-10-21
Extending Type coverage for nd_item and group classes

kopylovanat:
Changed test names *work_group_copy to *work_group_copy_core to allow it to be run separately from other *work_group_copy tests.

Added type coverage for
  - nd_item::async_work_group_copy member functions
  - group::async_work_group_copy member functions

### DIFF
--- a/tests/group/group_async_work_group_copy_common.h
+++ b/tests/group/group_async_work_group_copy_common.h
@@ -2,7 +2,7 @@
 //
 //  SYCL 1.2.1 Conformance Test Suite
 //
-//  Copyright:	(c) 2017 by Codeplay Software LTD. All Rights Reserved.
+//  Provides common methods for group async_work_group_copy tests
 //
 *******************************************************************************/
 
--- a/tests/group/group_async_work_group_copy_common.h
+++ b/tests/group/group_async_work_group_copy_common.h
@@ -1,0 +1,98 @@
+/*******************************************************************************
+//
+//  SYCL 1.2.1 Conformance Test Suite
+//
+//  Copyright:	(c) 2017 by Codeplay Software LTD. All Rights Reserved.
+//
+*******************************************************************************/
+
+#ifndef SYCL_1_2_1_TESTS_GROUP_ASYNC_WORK_GROUP_COPY_COMMON_H
+#define SYCL_1_2_1_TESTS_GROUP_ASYNC_WORK_GROUP_COPY_COMMON_H
+
+#include "../common/common.h"
+
+namespace {
+  inline cl::sycl::queue makeQueueOnce() {
+    static cl::sycl::queue q = sycl_cts::util::get_cts_object::queue();
+    return q;
+  }
+
+  using namespace sycl_cts;
+
+  static constexpr size_t GROUP_RANGE_1D = 2;
+  static constexpr size_t GROUP_RANGE_2D = 4;
+  static constexpr size_t GROUP_RANGE_3D = 8;
+  static constexpr size_t BUFFER_SIZE = 128;
+
+  template<typename T, int dim>
+  class test_kernel;
+
+  template<typename T, int dim>
+  void check_dim(cl::sycl::queue &queue, cl::sycl::buffer<T, 1> &buf, cl::sycl::range<dim> range) {
+    queue.submit([&](cl::sycl::handler &cgh) {
+    auto accGlobal =
+        buf.template get_access<cl::sycl::access::mode::read_write>(cgh);
+
+    auto accLocal =
+        cl::sycl::accessor<T, 1, cl::sycl::access::mode::read_write,
+                            cl::sycl::access::target::local>(
+            cl::sycl::range<1>(BUFFER_SIZE), cgh);
+
+    cgh.parallel_for_work_group<test_kernel<T, dim>>(range,
+        [=](cl::sycl::group<dim> my_group) {
+        auto ptrGlobal = accGlobal.get_pointer();
+        auto ptrLocal = accLocal.get_pointer();
+
+        my_group.async_work_group_copy(ptrLocal, ptrGlobal,
+                                        BUFFER_SIZE);
+        my_group.async_work_group_copy(ptrGlobal, ptrLocal,
+                                        BUFFER_SIZE);
+
+        constexpr size_t stride = 2;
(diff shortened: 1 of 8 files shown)


kopylovanat:
@illuhad @keryell @ProGTX, could you possibly take a look, please?

tests/nd_item/nd_item_async_work_group_copy_fp16.cpp
```diff
@@ -0,0 +1,37 @@
...
+  */
+  void get_info(test_base::info &out) const override {
+    set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
+  }
+
+  /** execute the test
+  */
+  void run(util::logger &log) override {
+
+    if(makeQueueOnce().get_device().has_extension("cl_khr_fp16")){
```
bader, line 27: Please, log about skipping the tests in all *[fp16,fp64].cpp tests.
Like this https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/tests/accessor/accessor_api_buffer_fp16.cpp#L37.

tests/group/group_async_work_group_copy_common.h
```diff
@@ -0,0 +1,98 @@
...
+                                        BUFFER_SIZE);
+
+        constexpr size_t stride = 2;
+        constexpr size_t numElements = BUFFER_SIZE / stride;
+        my_group.async_work_group_copy(ptrLocal, ptrGlobal, numElements,
+                                        stride);
+        my_group.async_work_group_copy(ptrGlobal, ptrLocal, numElements,
+                                        stride);
+        });
+});
```
bader, line 58:
```suggestion
    });
```

bader (changes requested):

tests/group/group_async_work_group_copy_fp16.cpp
```diff
@@ -0,0 +1,43 @@
...
+  */
+  void run(util::logger &log) override {
+
+    auto queue = util::get_cts_object::queue();
+
+    if (!queue().get_device().has_extension("cl_khr_fp16")) {
+        log.note(
+            "Device does not support half precision floating point operations");
+        return;
+      }
```
bader, line 33:
```suggestion
    if (!queue().get_device().has_extension("cl_khr_fp16")) {
      log.note(
            "Device does not support half precision floating point operations");
      return;
    }
```

Formatting in https://github.com/KhronosGroup/SYCL-CTS/pull/64/commits/5a18918e03698283b75cdb516bb77c301c76080c seems to be a bit off.

bader (approved):

tests/nd_item/nd_item_async_work_group_copy.cpp
```diff
@@ -33,111 +23,65 @@ class TEST_NAME : public util::test_base {
...
+    check_type_and_vec<cl::sycl::cl_uchar>(log);
+    check_type_and_vec<cl::sycl::cl_short>(log);
+    check_type_and_vec<cl::sycl::cl_ushort>(log);
+    check_type_and_vec<cl::sycl::cl_int>(log);
+    check_type_and_vec<cl::sycl::cl_uint>(log);
+    check_type_and_vec<cl::sycl::cl_long>(log);
+    check_type_and_vec<cl::sycl::cl_ulong>(log);
+    check_type_and_vec<cl::sycl::cl_float>(log);
+
+#ifdef INT8_MAX
```
keryell, line 53: I appreciate the pedantism and the micro-optimization in this area.
I guess it might be difficult to test the tests if we do not have 9-bit host computer, for example... :-)
But is it possible to have `std::is_same<cl::sycl::cl_char, std::int8_t>::value` false according to the spec?
In that case we should not try to run a test but just `static_assert` at least say it is inconsistent.

tests/nd_item/nd_item_async_work_group_copy_fp16.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto queue = util::get_cts_object::queue();
+
+    if (!queue.get_device().has_extension("cl_khr_fp16")) {
+      log.note(
+        "Device does not support half precision floating point operations");
+      return;
+    }
+
+    check_type_and_vec<cl::sycl::half>(log);
+    check_type_and_vec<cl::sycl::cl_half>(log);
```
keryell, line 36: And here you do not try to skip the test, which is probably good.

tests/nd_item/nd_item_async_work_group_copy_fp64.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto queue = util::get_cts_object::queue();
+
+    if (!queue.get_device().has_extension("cl_khr_fp64")) {
+      log.note(
+        "Device does not support double precision floating point operations");
+      return;
+    }
+
+    check_type_and_vec<double>(log);
+    check_type_and_vec<cl::sycl::cl_double>(log);
```
keryell, line 36: And here you do not try to skip the test, which is probably good.

tests/nd_item/nd_item_async_work_group_copy_fp16.cpp
kopylovanat: Not sure I understand what you mean. You think these tests shouldn't be skipped even if cl_khr_fp16 is not supported? Or you think it's a good example for how to handle intN_t tests?

tests/nd_item/nd_item_async_work_group_copy.cpp
kopylovanat: Thank you for comments and sorry for the delay.
According to the spec cl_char is "Alias to a signed 8-bit integer, as defined by the C++11 ISO Standard" and from my understanding it doesn't guaranteed that it's the same type as std::int8_t, even if it's likely, that's why I think we can't skip tests for intN_t types without these checks. And since they are optional in C++11 Standard it's not guaranteed that they are defined, that's why #ifdef INT8_MAX is used - to make sure int8_t is defined. 

 Do you think it would be better to just run tests for intN_t types if they are defined?

bader:
@illuhad @keryell @ProGTX, could you possibly take a look, please?

ProGTX (approved):

bader:
@keryell, could you take a look, please?

keryell (approved):

PR #65, opened 2020-12-04
Enable half/double cmake options

yuriykoch:
Enable cmake options filtering out half/double tests at compile time:
 - SYCL_CTS_ENABLE_HALF_TESTS
 - SYCL_CTS_ENABLE_DOUBLE_TESTS

Default option values are switched to ON, because:
- neither SYCL_CTS_ENABLE_DOUBLE_TESTS nor SYCL_CTS_ENABLE_HALF_TESTS
  affect any of the current tests in the CTS:
  tests currently behave like if these options are ON by default
- half types should be available at the compile time according to the
  SYCL 1.2.1, 5.1 and SYCL 2020, D.6.1
- in case there is no  fp16/fp64 extension available for device, a
  feature_not_supported exception should be thrown at the runtime
- double types should be available at the compile time according to
  the SYCL 1.2.1, 4.10.1 and 6.5
- according to the SYCL 2020, 6.5 an implementation may not raise an
  error in case application did not specifically ask to submit the
  kernel to the specific device or to compile the kernel for it

Options are enabled for new fp16/fp64 tests naming convention only;
any elder tests can be modified to use macro defines if required

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,8 +27,8 @@ include(AddSYCLExecutable)
 
 # ------------------
 # Double and Half variables
-option(SYCL_CTS_ENABLE_DOUBLE_TESTS "Enable Double tests." OFF)
-option(SYCL_CTS_ENABLE_HALF_TESTS "Enable Half tests." OFF)
+option(SYCL_CTS_ENABLE_DOUBLE_TESTS "Enable Double tests." ON)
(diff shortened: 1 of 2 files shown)


yuriykoch:
@bader @illuhad @keryell @ProGTX could you possibly take a look, please?

ProGTX (approved):

bader (approved):

keryell (approved):

PR #66, opened 2020-12-05
[CMake] Set explicit value of fp-model option for Intel compiler

bader:
`fp-model` default value might enable optimizations that are not
value-safe on floating-point data.


bader:
Tagging @ProGTX, @illuhad.

ProGTX (approved):

illuhad (approved):

PR #67, opened 2021-02-02
Allow small ULP errors for vec operator/

kopylovanat:
Changing the verification of vec division operator for floating point types to allow errors within 2.5 ulp (according to OpenCL spec, section 7.4)


keryell (approved):
While I am not happy by discovering pointers in the CTS, you are not guilty. :-)
C++17 `if constexpr` will simplify this a lot in the future...

ProGTX (approved):

bader:
Thanks!

PR #68, opened 2021-02-04
Added check for result correctness for Relational functions (4.13.7)

kopylovanat:
Remaining gap (test coverage): result correctness isn't checked for Integer functions, Geometric Functions, Math functions

### DIFF
--- a/util/math_helper.h
+++ b/util/math_helper.h
@@ -28,6 +28,20 @@ cl::sycl::vec<R, N> run_func_on_vector(funT fun, Args... args) {
   return res;
 }
 
+/* helper for relational functions where true result gives 1 for scalar
+    and -1 for vector argument types */
+template <typename R, typename T, int N, typename funT, typename... Args>
+cl::sycl::vec<R, N> run_rel_func_on_vector(funT fun, Args... args) {
+  cl::sycl::vec<R, N> res;
+  for (int i = 0; i < N; i++) {
+    if (fun(getElement<T, N>(args, i)...) == 1)
+      setElement<R, N>(res, i, -1);
+    else
+      setElement<R, N>(res, i, 0);
+  }
+  return res;
+}
+
 /* cast an integer to a float */
 float int_to_float(uint32_t x);
 
--- a/util/math_helper.h
+++ b/util/math_helper.h
@@ -9,6 +9,8 @@
 #ifndef __SYCLCTS_UTIL_MATH_HELPER_H
 #define __SYCLCTS_UTIL_MATH_HELPER_H
 
+#include <climits>
+
 #include "../util/stl.h"
 #include "../tests/common/sycl.h"
 #include "./../oclmath/mt19937.h"
@@ -81,7 +83,7 @@ rel_func_dispatcher(funT fun, cl::sycl::vec<T, N> a, Args... args) {
 
 template <typename T>
 int32_t num_bits(T) {
-  return int32_t(sizeof(T) * 8u);
+  return int32_t(sizeof(T) * CHAR_BIT);
 }
 
 template <typename T> bool if_bit_set(T num, int bit) {
--- a/util/math_helper.h
+++ b/util/math_helper.h
@@ -34,7 +34,7 @@ template <typename R, typename T, int N, typename funT, typename... Args>
 cl::sycl::vec<R, N> run_rel_func_on_vector(funT fun, Args... args) {
   cl::sycl::vec<R, N> res;
   for (int i = 0; i < N; i++) {
-    if (fun(getElement<T, N>(args, i)...) == 1)
+    if (fun(getElement<T, N>(args, i)...))
       setElement<R, N>(res, i, -1);
     else
       setElement<R, N>(res, i, 0);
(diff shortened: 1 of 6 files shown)


util/math_reference.h
```diff
@@ -15,22 +15,289 @@
...
+      [](float x, float y) { return isequal(x, y); }, a, b);
+}
+template <int N>
+cl::sycl::vec<int64_t, N> isequal(cl::sycl::vec<double, N> a,
+                                  cl::sycl::vec<double, N> b) {
+  return sycl_cts::math::run_rel_func_on_vector<int64_t, double, N>(
+      [](double x, double y) { return isequal(x, y); }, a, b);
+}
+
+int32_t isnotequal(float x, float y);
```
keryell, line 33: In theory all these could be metaprogrammed as
```c++
template <typename T>
auto isnotequal(const T& a, const T& b) {
  return some_generic_dispatcher<std::not_equal_to>(a, b);
}
```
and writing `some_generic_dispatcher` looks like a good coding exercise. :-)

util/math_reference.h
```diff
@@ -15,22 +15,289 @@
...
+template <int N> cl::sycl::vec<int32_t, N> signbit(cl::sycl::vec<float, N> a) {
+  return sycl_cts::math::run_rel_func_on_vector<int32_t, float, N>(
+      [](float x) { return signbit(x); }, a);
+}
+template <int N> cl::sycl::vec<int64_t, N> signbit(cl::sycl::vec<double, N> a) {
+  return sycl_cts::math::run_rel_func_on_vector<int64_t, double, N>(
+      [](double x) { return signbit(x); }, a);
+}
+
+int any(signed char);
```
keryell, line 200: It is not clear what are all these declarations.
If it is useful, explain in a comment.

keryell (changes requested):
Thanks for the improvement.
I have the feeling that the code could be less redundant.

util/math_helper.h
```diff
@@ -28,6 +28,20 @@ cl::sycl::vec<R, N> run_func_on_vector(funT fun, Args... args) {
   return res;
 }
 
+/* helper for relational functions where true result gives 1 for scalar
+    and -1 for vector argument types */
```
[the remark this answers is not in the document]
keryell: Looking at the SYCL 2020 specification, it is not clear.
For example `isunordered` returns a vector of 0 or non null in the  case of `sycl::vec` and a vector of `bool` for `sycl::marray`.
Curious these differences according to the various kinds of functions. So comparing with `== 1` seems OK for `bool` but not for the other case.
So at least just removing ` == 1` looks safer.
But more generally about these differences between the functions, do you think this is a bug in the specification?

util/math_reference.cpp
```diff
@@ -825,56 +800,56 @@ int64_t upsample(int32_t h, uint32_t l) {
...
-  for (int i = 0; i < num_bits(x); i++)
+  for (int i = 0; i < sycl_cts::math::num_bits(x); i++)
     if (x & (1 << i)) lz++;
   return lz;
 }
 
 int64_t popcount(const int64_t x) {
   int lz = 0;
-  for (int i = 0; i < num_bits(x); i++)
+  for (int i = 0; i < sycl_cts::math::num_bits(x); i++)
```
keryell, line 852: Some day we should factorized all this code...
Obviously not in this PR.

util/math_reference.h
```diff
@@ -15,22 +15,178 @@
...
-int isinf(float x);
-int isnan(float x);
-int isnormal(float x);
-int signbit(float x);
+template<typename T>
+struct isfinite_func{
+  constexpr bool operator()(const T& x) const {
+   return std::isfinite(x);
+  }
+};
```
keryell, line 87: Not fundamental, but you could be lazy for all these functions everywhere with C++14:
```suggestion
auto constexpr isfinite_func = [] (const auto& x) { return std::isfinite(x); };
```

util/math_helper.h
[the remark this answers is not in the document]
kopylovanat: I don't think it's a bug in the specification - `isunordered` description says it returns non-zero if x or y is NaN and to me it seems to coincide with the idea that these functions return 1 or -1 depending on is it scalar or vector version that referenced for relational functions in 'as defined in the OpenCL 1.2 specification document par. 6.12.6'. May be if it was stated explicitly it would be more clear.

Removing `== 1` would be safer anyway indeed. Done.

util/math_reference.h
kopylovanat: Changed to lambda and added dispatcher to work with it.

ProGTX (approved):

bader:
@keryell, could you check that your comments are addressed, please?

util/math_helper.h
```diff
@@ -28,6 +28,70 @@ cl::sycl::vec<R, N> run_func_on_vector(funT fun, Args... args) {
...
+template <typename funT, typename T, int N, typename... Args>
+typename cl::sycl::vec<typename rel_funcs_return<T>::type, N>
+rel_func_dispatcher(funT fun, cl::sycl::vec<T, N> a, Args... args) {
+  return run_rel_func_on_vector<typename rel_funcs_return<T>::type, T, N>(
+      fun, a, args...);
+}
+
+template <typename T>
+int32_t num_bits(T) {
+  return int32_t(sizeof(T) * 8u);
```
keryell, line 84: The standard way is to use `CHAR_BIT` from `<climits>`

util/math_reference.cpp
```diff
@@ -26,11 +26,6 @@ void type_punn(const A &from, B &to) {
...
 }
 
-template <typename T>
-int32_t num_bits(T) {
-  return int32_t(sizeof(T) * 8u);
-}
-
 const uint64_t max_uint64_t = (~0x0ull);
 const uint32_t max_uint32_t = 0xffffffff;
 const uint16_t max_uint16_t = 0xffff;
```
keryell, line 31: In the future, while modernizing the CTS we should replace these by https://en.cppreference.com/w/cpp/types/numeric_limits

util/math_helper.h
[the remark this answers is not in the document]
keryell: Thanks.

keryell (approved):
That looks quite clearer and more modern.
Thanks!
Just 1 pedantic detail...

util/math_helper.h
bader: @kopylovanat, could you update the header, please?

kopylovanat: Updated.

PR #70, opened 2021-02-06
[CMake] Update C++ and SYCL versions for 2020.

bader:
SYCL 2020 requires compiler to support C++17 at least.
Update SYCL version for DPC++ compiler from 1.2.1 to 2020.

### DIFF
--- a/cmake/FindComputeCpp.cmake
+++ b/cmake/FindComputeCpp.cmake
@@ -35,16 +35,11 @@ set_target_properties(ComputeCpp::Runtime PROPERTIES
     IMPORTED_LOCATION                    "${ComputeCpp_LIBRARIES}"
     INTERFACE_INCLUDE_DIRECTORIES        "${ComputeCpp_INCLUDE_DIRS}"
     INTERFACE_LINK_LIBRARIES             "OpenCL::OpenCL;Threads::Threads"
-    INTERFACE_DEVICE_COMPILE_OPTIONS     "-sycl")
+    INTERFACE_DEVICE_COMPILE_OPTIONS     "-sycl;-DSYCL_LANGUAGE_VERSION=2020")
 if (WIN32)
     set_property(TARGET ComputeCpp::Runtime APPEND PROPERTY
                  INTERFACE_DEVICE_COMPILE_DEFINITIONS
                  "_SIZE_T_DEFINED;_NO_CRT_STDIO_INLINE")
-    set_property(TARGET ComputeCpp::Runtime APPEND PROPERTY
-                 INTERFACE_DEVICE_COMPILE_OPTIONS "-std=c++14")
-else()
-    set_property(TARGET ComputeCpp::Runtime APPEND PROPERTY
-                 INTERFACE_DEVICE_COMPILE_OPTIONS "-std=c++11")
 endif()
 
 add_library(SYCL::SYCL INTERFACE IMPORTED GLOBAL)
@@ -80,6 +75,7 @@ function(build_spir exe_name spir_target_name source_file output_path)
             -O2
             -mllvm -inline-threshold=1000
             -intelspirmetadata
+            -DSYCL_LANGUAGE_VERSION=2020
             ${COMPUTECPP_USER_FLAGS}
             ${platform_specific_args}
             $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\t-I\">\">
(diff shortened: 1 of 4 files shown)


bader:
@illuhad, I see that FindhipSYCL.cmake file explicitly [sets C++14](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindhipSYCL.cmake#L10) version. Can we drop this add line and keep the default project setting for SYCL 2020 branch?

@ProGTX, there is also C++ version customization for [ComputeCPP](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindComputeCpp.cmake#L39-L48), can I remove it for SYCL 2020 branch?

illuhad:
> @illuhad, I see that FindhipSYCL.cmake file explicitly sets C++14 version. Can we drop this add line and keep the default project setting for SYCL 2020 branch?

@bader Yes - hipSYCL requires C++17 as well now. This line was added when the global setting was still C++11, but hipSYCL was already requiring C++14.

bader:
> > @illuhad, I see that FindhipSYCL.cmake file explicitly sets C++14 version. Can we drop this add line and keep the default project setting for SYCL 2020 branch?
> 
> @bader Yes - hipSYCL requires C++17 as well now. This line was added when the global setting was still C++11, but hipSYCL was already requiring C++14.

Thanks for clarification. I removed C++14 requirement for hipSYCL as project default version is been updated to C++17 with this patch.

illuhad (approved):

keryell (approved):

bader:
> @ProGTX, there is also C++ version customization for [ComputeCPP](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-1.2.1/master/cmake/FindComputeCpp.cmake#L39-L48), can I remove it for SYCL 2020 branch?

This PR is ready to go. 
@ProGTX, I can update CMake files for ComputeCPP before merging (although I won't be able to test it) or in a separate PR. Let me know which way do you prefer.

ProGTX:
It should be fine to set `CMAKE_CXX_STANDARD` to `17` across the board, that shouldn't break things. For ComputeCpp you also need to define `-DSYCL_LANGUAGE_VERSION=2020` for both the host and device compilers. I'd prefer this be done in this PR, but I don't mind if it's separate.

ProGTX (approved):

bader:
> It should be fine to set `CMAKE_CXX_STANDARD` to `17` across the board, that shouldn't break things. For ComputeCpp you also need to define `-DSYCL_LANGUAGE_VERSION=2020` for both the host and device compilers. I'd prefer this be done in this PR, but I don't mind if it's separate.

Let's try updating all CMake configuration files in this PR. I tried to incorporate this feedback, but I'm not sure if I did it correctly.
@ProGTX, could you check, please?

cmake/FindComputeCpp.cmake
```diff
@@ -35,16 +35,11 @@ set_target_properties(ComputeCpp::Runtime PROPERTIES
     IMPORTED_LOCATION                    "${ComputeCpp_LIBRARIES}"
     INTERFACE_INCLUDE_DIRECTORIES        "${ComputeCpp_INCLUDE_DIRS}"
     INTERFACE_LINK_LIBRARIES             "OpenCL::OpenCL;Threads::Threads"
-    INTERFACE_DEVICE_COMPILE_OPTIONS     "-sycl")
+    INTERFACE_DEVICE_COMPILE_OPTIONS     "-sycl;-DSYCL_LANGUAGE_VERSION=2020")
```
ProGTX, line 38:
```suggestion
    INTERFACE_COMPILE_DEFINITIONS        "-DSYCL_LANGUAGE_VERSION=2020"
    INTERFACE_DEVICE_COMPILE_DEFINITIONS "-DSYCL_LANGUAGE_VERSION=2020"
    INTERFACE_DEVICE_COMPILE_OPTIONS     "-sycl;-std=c++17"
  )
```

cmake/FindComputeCpp.cmake
```diff
@@ -80,6 +75,7 @@ function(build_spir exe_name spir_target_name source_file output_path)
             -O2
             -mllvm -inline-threshold=1000
             -intelspirmetadata
+            -DSYCL_LANGUAGE_VERSION=2020
```
ProGTX, line 78:
```suggestion
```

I don't think this is necessary along with the changes above.

ProGTX (approved):
Thanks @bader for the changes. I've added some suggestions that I believe are necessary.

bader:
> Thanks @bader for the changes. I've added some suggestions that I believe are necessary.

Thanks. I applied your suggestions and I think this PR is ready now.

PR #71, opened 2021-02-18
[Test generation] Add DEPENDS argument

ProGTX:
1. Added `DEPENDS` argument to `generate_cts_test`
2. Added files for `math_builtin_api` test to depend on

### DIFF
--- a/tests/math_builtin_api/CMakeLists.txt
+++ b/tests/math_builtin_api/CMakeLists.txt
@@ -5,6 +5,12 @@ set(MATH_VARIANT base double half)
 
 set(MATH_CAT integer geometric native half)
 
+set(math_builtin_depends
+  "modules/sycl_functions.py"
+  "modules/sycl_types.py"
+  "modules/test_generator.py"
+)
+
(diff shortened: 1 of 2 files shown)


tests/math_builtin_api/CMakeLists.txt
```diff
@@ -5,6 +5,13 @@ set(MATH_VARIANT base double half)
 
 set(MATH_CAT integer geometric native half)
 
+set(math_builtin_depends
+  "math_builtin.h"
```
bader, line 9: Should we put a header file as a dependency for source generation?

bader (approved):
LGTM, although I'm not sure if `math_builtin.h` should be in the list.

tests/math_builtin_api/CMakeLists.txt
ProGTX: I think you're right, it shouldn't have an impact on the generator

bader (approved):

bader:
@keryell, could you take a look, please?

keryell (approved):
That looks good, from my limited CMake knowledge.

PR #73, opened 2021-03-02
Add extensive mode cmake option

yuriykoch:
Improving data type coverage up to the full one can give an
unreasonable performance overhead for the specific functionality,
like accessors.
Type coverage for the default mode should be balanced in terms of
coverage and performance, but there should be an option to make all
extra checks when performance doesn't matter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,15 @@ endif()
 include(AddSYCLExecutable)
 
 # ------------------
+# Extensive mode for improved coverage
+option(SYCL_CTS_ENABLE_EXTENSIVE_MODE
+       "Enable extensive mode with extended type coverage" OFF)
+if(SYCL_CTS_ENABLE_EXTENSIVE_MODE)
+    add_definitions(-DSYCL_CTS_EXTENSIVE_MODE)
+endif()
+# ------------------
+
+# ------------------
 # Double and Half variables
 option(SYCL_CTS_ENABLE_DOUBLE_TESTS "Enable Double tests." ON)
 option(SYCL_CTS_ENABLE_HALF_TESTS "Enable Half tests." ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,11 +26,16 @@ endif()
 include(AddSYCLExecutable)
 
 # ------------------
-# Extensive mode for improved coverage
-option(SYCL_CTS_ENABLE_EXTENSIVE_MODE
-       "Enable extensive mode with extended type coverage" OFF)
-if(SYCL_CTS_ENABLE_EXTENSIVE_MODE)
-    add_definitions(-DSYCL_CTS_EXTENSIVE_MODE)
+# Extensive mode for full coverage
+option(SYCL_CTS_ENABLE_FULL_CONFORMANCE
+       "Enable full conformance mode with extensive tests" OFF)
+if(SYCL_CTS_ENABLE_FULL_CONFORMANCE)
+    message(STATUS "Full conformance mode: ON")
+    add_definitions(-DSYCL_CTS_FULL_CONFORMANCE)
+else()
+    message(STATUS "Full conformance mode: OFF")
+    message(WARNING
+            "Full conformance mode should be used for conformance submission")
 endif()
 # ------------------
 
--- a/README.rst
+++ b/README.rst
@@ -84,8 +84,9 @@ When configuring CMake, it is possible to use these flags:
 ``SYCL_CTS_TEST_FILTER``
   Specify which filter to use when building tests.
 
-``SYCL_CTS_ENABLE_EXTENSIVE_MODE``
-  Enable extended coverage with huge compilation and execution time
+``SYCL_CTS_ENABLE_FULL_CONFORMANCE``
+  Enable extensive coverage with huge compilation and execution time.
+  This mode is switched off by default. Should be swithed on for conformance.
 
 ``HOST_COMPILER_FLAGS``
   Flags that will be passed to the host compiler.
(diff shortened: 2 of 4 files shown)


bader:
@yuriykoch, can we complete this patch with usages of pre-processor macro in test suite source code?

psalz:
Yes! This is what I need for #58 :)

bader:
> Yes! This is what I need for #58 :)

If so, I'm okay to merge this right away as is to not block #58 and update tests source code in follow-up patches.

psalz:
I'm going to rebase #58 onto this and will report back then!

CMakeLists.txt
```diff
@@ -25,6 +25,15 @@ endif()
 
 include(AddSYCLExecutable)
 
+# ------------------
+# Extensive mode for improved coverage
+option(SYCL_CTS_ENABLE_EXTENSIVE_MODE
+       "Enable extensive mode with extended type coverage" OFF)
```
ProGTX, line 31:
```suggestion
       "Enable extensive mode with extended type coverage" ON)
```

ProGTX (changes requested):
OpenCL CTS does the inverse, it has a "wimpy" mode that is used for implementation convenience, but it cannot be used for conformance. I think we should have it defined in a similar way - the tests don't represent "extended" coverage, they're required for passing conformance.

So by default test the full CTS, but the user can opt-in to the less intensive CTS checking, which is not valid for passing conformance. That fact should also be clearly documented, both in the README and in the conformance report.

psalz:
I agree that it certainly makes sense to clarify which one is required for conformance. However, build times can get pretty brutal in non-wimpy/extensive mode, severely impacting the developer experience if one just wants to test a quick change and forgets to pass the appropriate mode flag. So how about keeping wimpy as a default, but printing a warning during CMake configuration and at runtime, informing the user that this build cannot be used for establishing conformance?

ProGTX (changes requested):
Based on the WG discussions your proposal makes sense @psalz . It seems like people prefer the CTS to not take too long to run by default, but we agreed that all tests are required for conformance, and that fact should be clearly signaled.
So it seems we need to:
1. Rename the option to something that signals conformance, e.g. `SYCL_CTS_FULL_CONFORMANCE`
1. Have that option be `OFF` by default
1. Print the value of that option in the conformance report
    * Potentially also add a warning somewhere if it's set to OFF
1. Add some documentation (README) that submitting for conformance requires that flag to be set to `ON`

yuriykoch:
Looks like I need some clarification regarding a conformance report.

1. Do you mean printing this value to the "Implementation" table of the conformance_report.xml?
  Why not to extend generate_cmake_call() of the run_conformance_tests.py with an appropriate cmake option?
  Should the run_conformance_tests.py script support running without full conformance option ON?

2. README.rst doesn't have run_conformance_tests.py mentioned at all, but it has a section for the runtests.py script.
  Should we use run_conformance_tests.py or runtests.py for a conformance report?

ProGTX:
Passing the flag in `run_conformance_tests.py` seems fine - it is necessary to set it to `ON` anyway there (since the default is `OFF`), and then the conformance report will print out the flags as you mentioned.

That's unfortunate that the README doesn't talk about `run_conformance_tests.py` - this is the script that is run to get the conformance report. Could you please add a quick note about it? Properly documenting it is out of scope of this PR, but we could at least mention it, and also mention this new flag there.

CMakeLists.txt
yuriykoch: @ProGTX Please, review the latest changes at 9b95609 and d773e0a

psalz:
On a somewhat related note: This might also be a good time to get rid of the `--wimpy` run time parameter that is  [still supported](https://github.com/KhronosGroup/SYCL-CTS/blob/a3960a9a84557cefd6053894e5d2aed84d31a713/util/test_manager.cpp#L134), but as far as I can tell, not being used anywhere.

ProGTX (approved):

CMakeLists.txt
ProGTX: Thank you for the changes!

README.rst
```diff
@@ -186,4 +190,9 @@ quickly identify failures and conformance rate::
 This report should be packaged with the run tests and sent to Khronos for
 conformance submission.
 
-The conformance submission requires the use of the core.csv filter.
+Conformance submission
+----------------------
+
+The conformance submission requires the use of the core.csv filter alongside
+with the run_conformance_tests.py script. Please, note that this script will
+switch on the SYCL_CTS_ENABLE_FULL_CONFORMANCE option.
```
keryell, line 198:
```suggestion
The conformance submission requires the use of the ``core.csv`` filter alongside
with the ``run_conformance_tests.py`` script. Please, note that this script will
switch on the ``SYCL_CTS_ENABLE_FULL_CONFORMANCE`` option.
```
I am unsure about the position of the last "on" in English. Should it be at the end?

run_conformance_tests.py
```diff
@@ -90,15 +95,24 @@ def handle_args(argv):
...
-            args.conformance_filter, args.implementation_name,
+            full_conformance, args.conformance_filter, args.implementation_name,
             args.additional_cmake_args, host_names, opencl_names,
             args.additional_ctest_args, args.build_only)
 
-
-def generate_cmake_call(cmake_exe, build_system_name, conformance_filter,
-                        additional_cmake_args, host_names, opencl_names):
+def split_additional_args(additional_args):
+    if additional_args is None:
```
keryell, line 106: Could you add a documentation string here?
I have issues understanding what you are trying to do here.

run_conformance_tests.py
```diff
@@ -90,15 +95,24 @@ def handle_args(argv):
...
+        return []
+    if os.path.sep is '\\':
+        return shlex.split(additional_args.replace('\\','\\\\'))
```
keryell, line 109: I have the feeling that even on Unix if you have a `\` this is a special character and you have to escape it with... `\`!
It is not clear to me why not using for example `shlex.quote()` here.

run_conformance_tests.py
yuriykoch: Done. Shlex doesn't support None, so run_conformance_tests.py crashed in case any of "additional arguments" parameter was not set.

run_conformance_tests.py
yuriykoch: Yes, ``shlex.quote()`` fits much better. I've missed it whilst trying to fix passing any windows path within ``additional-cmake-args``. Thanks.

keryell (approved):
Thanks.

PR #74, opened 2021-03-05
Update .clang-format config to sort includes

psalz:
This updates the `.clang-format` configuration file to `SortIncludes: true`.

I've also run the new configuration on all existing test and util files, however, as not all files fully conform to clang-format, this only adds the changes to includes. Unfortunately I cannot confirm whether this breaks compilation of any of the tests (I mean, it shouldn't, but you never know...), as I've never been able to compile the full CTS without errors, with any SYCL implementation.

cc @keryell

### DIFF
--- a/tests/common/common.h
+++ b/tests/common/common.h
@@ -12,20 +12,18 @@
 // include our proxy to the real sycl header
 #include "sycl.h"
 
-// test framework specific device selector
-#include "../common/cts_selector.h"
-#include "../common/cts_async_handler.h"
-#include "../common/get_cts_object.h"
+#include "../../util/math_vector.h"
 #include "../../util/proxy.h"
+#include "../../util/test_base.h"
+#include "../common/cts_async_handler.h"
+#include "../common/cts_selector.h"
+#include "../common/get_cts_object.h"
 #include "macros.h"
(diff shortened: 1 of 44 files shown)


tests/common/common.h
```diff
@@ -13,19 +13,19 @@
 #include "sycl.h"
 
 // test framework specific device selector
-#include "../common/cts_selector.h"
+#include "../../util/proxy.h"
 #include "../common/cts_async_handler.h"
+#include "../common/cts_selector.h"
 #include "../common/get_cts_object.h"
-#include "../../util/proxy.h"
 #include "macros.h"
```
bader, line 21: We should probably remove this empty line, so that clang-format will sort both groups together.

tests/common/common_vec.h has the same issue.

bader (approved):

ProGTX (approved):

tests/common/common.h
psalz: Done. I've also removed the comment line in those two files regarding the device selector, which was misplaced due to the sorting and doesn't strike me as too important.

bader: Thanks!

PR #75, opened 2021-03-10
Changing the conformance tests to use const kernel functions

kopylovanat:
SYCL 2020 is changing kernel functions to be const.

### DIFF
--- a/tests/accessor/accessor_api_common_buffer_local.h
+++ b/tests/accessor/accessor_api_common_buffer_local.h
@@ -50,7 +50,7 @@ template <typename T, cl::sycl::access::mode mode,
           cl::sycl::access::target target,
           cl::sycl::access::placeholder placeholder>
 T multidim_subscript_read(
-    cl::sycl::accessor<T, 1, mode, target, placeholder>& acc,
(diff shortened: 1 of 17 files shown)


ProGTX (approved):

bader:
@keryell, @illuhad, could you take a look at this change, please?

tests/accessor/accessor_api_common_buffer_local.h
```diff
@@ -77,7 +77,7 @@ template <typename T, cl::sycl::access::mode mode,
           cl::sycl::access::target target,
           cl::sycl::access::placeholder placeholder>
 void multidim_subscript_write(
-    cl::sycl::accessor<T, 1, mode, target, placeholder>& acc,
+    const cl::sycl::accessor<T, 1, mode, target, placeholder>& acc,
```
keryell, line 80: Just double checking that you want `const` when writing into `acc`.
I think this is right, since I had to use some `mutable` in the triSYCL implementation.

kopylovanat: Yes, we still need it to be const and accessor operator[] that are used here is const according to SYCL-2020 spec p. 160

keryell (approved):
Thanks.

PR #77, opened 2021-03-16
Allow small ULP errors for vec operator/

kopylovanat:
Changing the verification of vec division operator for floating point types to allow
errors within 2.5 ulp (according to OpenCL spec, section 7.4)

### DIFF
--- a/util/accuracy.h
+++ b/util/accuracy.h
@@ -1,0 +1,34 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Support for accuracy check
+//
+*******************************************************************************/
+
+#ifndef __SYCLCTS_UTIL_ACCURACY_H
+#define __SYCLCTS_UTIL_ACCURACY_H
+
+#include <cmath>
+#include <limits>
+
+/**
+ * @brief Provides ulp(x) by definition given in OpenCL 1.2 rev. 19, 7.4
+ *        See Jean-Michel Muller "On the definition of ulp (x)", definition 7
+ *        Using sycl functions.
+ */
+template <typename T> T get_ulp_sycl(T x) {
+  const T inf = std::numeric_limits<T>::infinity();
+  const T negative = cl::sycl::fabs(cl::sycl::nextafter(x, -1.0f * inf) - x);
+  const T positive = cl::sycl::fabs(cl::sycl::nextafter(x, inf) - x);
+  return cl::sycl::fmin(negative, positive);
+}
+template <>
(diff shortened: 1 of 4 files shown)


kopylovanat:
Previously merged into SYCL-1.2.1 branch https://github.com/KhronosGroup/SYCL-CTS/pull/67

ProGTX (approved):
I believe this is just cherry-picking the changes that went into the 1.2.1 CTS already?

kopylovanat:
@ProGTX 
> I believe this is just cherry-picking the changes that went into the 1.2.1 CTS already?
> 
> Is the same true for these other PRs?
> 
> * #78
> * #79
> * #81

Yes, it's cherry-picking the change that went into the 1.2.1 CTS with PR https://github.com/KhronosGroup/SYCL-CTS/pull/67

Other mentioned PRs are not in 1.2.1 CTS branch.

util/accuracy.h
```diff
@@ -0,0 +1,34 @@
...
+#include <limits>
+
+/**
+ * @brief Provides ulp(x) by definition given in OpenCL 1.2 rev. 19, 7.4
+ *        See Jean-Michel Muller "On the definition of ulp (x)", definition 7
+ *        Using sycl functions.
+ */
+template <typename T> T get_ulp_sycl(T x) {
+  const T inf = std::numeric_limits<T>::infinity();
+  const T negative = cl::sycl::fabs(cl::sycl::nextafter(x, -1.0f * inf) - x);
```
keryell, line 22: Just curious... Would `-inf` just work instead of `-1.0f * inf`?

util/type_traits.h
```diff
@@ -0,0 +1,28 @@
...
+namespace {
+/**
+ * @brief Checks whether T is a floating-point sycl type
+ */
+template <typename T> struct is_cl_float_type {
+  static constexpr bool value = std::is_floating_point<T>::value ||
+                                std::is_same<cl::sycl::half, T>::value ||
+                                std::is_same<cl::sycl::cl_float, T>::value ||
+                                std::is_same<cl::sycl::cl_double, T>::value ||
+                                std::is_same<cl::sycl::cl_half, T>::value;
```
keryell, line 23: These exist only in the OpenCL back-end.
Is the CTS validating the OpenCL back-end when running on top of OpenCL?

kopylovanat: Since cl_float, cl_double, cl_half are still in the CTS right now, floating point detection still takes them into consideration too. Will be removed in the future with other coverage for these types.

keryell (approved):
Good!

PR #80, opened 2021-03-18
Provide std::array replacement for kernel code

yuriykoch:
Provide replacement type for fixed-width array to use in kernel functions. Implicitly uses pointer arithmetic for C++ ranged-for.

### DIFF
--- a/util/array.h
+++ b/util/array.h
@@ -36,7 +36,7 @@ struct array {
    */
   value_type values[N];
 
-  reference operator[](size_type pos) {
+  constexpr reference operator[](size_type pos) {
     return values[pos];
   }
   constexpr const_reference operator[](size_type pos) const {
@@ -47,13 +47,13 @@ struct array {
     return N;
   }
 
-  reference front() {
+  constexpr reference front() {
     return values[0];
   }
   constexpr const_reference front() const {
     return values[0];
   }
-  reference back() {
+  constexpr reference back() {
     return values[N-1];
   }
   constexpr const_reference back() const {
@@ -62,13 +62,13 @@ struct array {
 
   /** @brief Support for C++ ranged-for syntax
    */
-  iterator begin() noexcept {
+  constexpr iterator begin() noexcept {
     return &front();
   }
   constexpr const_iterator begin() const noexcept {
     return &front();
   }
-  iterator end() noexcept {
+  constexpr iterator end() noexcept {
     return &back();
   }
   constexpr const_iterator end() const noexcept {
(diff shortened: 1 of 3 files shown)


ProGTX:
Is there a specific reason for doing this? It generally looks good, I'm just curious what the motivation was.

util/array.h
```diff
@@ -0,0 +1,82 @@
...
+
+  /** @brief Internal data storage;
+   *         public for aggregate initialization availability
+   */
+  value_type values[N];
+
+  reference operator[](size_type pos) {
+    return values[pos];
+  }
+  const_reference operator[](size_type pos) const {
```
ProGTX, line 42: All `const` qualified functions can be marked `constexpr` if targetting SYCL 1.2.1. For SYCL 2020 all functions can be `constexpr`.

util/array.h
```diff
@@ -0,0 +1,82 @@
...
+    return values[pos];
+  }
+
+  inline constexpr size_type size() const noexcept {
```
ProGTX, line 46:
```suggestion
  constexpr size_type size() const noexcept {
```

`constexpr` implies `inline` when used on functions.

util/array.h
```diff
@@ -0,0 +1,82 @@
...
+  const_iterator begin() const noexcept {
+    return &front();
+  }
+  iterator end() noexcept {
+    return &back();
+  }
+  const_iterator end() const noexcept {
+    return &back();
+  }
+};
```
ProGTX, line 77: I think we're missing a `data()` function

bader:
> Is there a specific reason for doing this? It generally looks good, I'm just curious what the motivation was.

According to my understanding, it goal is provide exception-free implementation to avoid *potential* concerns regarding using std::array in SYCL device code.

util/array.h
yuriykoch: Do we really need it on the device side? I mean, we can always do
`for (const auto& item: arrayInstance) {
  /* access item */
}`

util/array.h
yuriykoch: Great catch, done

util/array.h
ProGTX: Not having it would probably be the only difference to a standard `array`, and I presume we want a drop-in replacement.

ProGTX (approved):
I think there are more concerns that just exceptions when it comes to the general STL, and we need to address them at some point. This change looks good, though.

keryell (approved):
That looks good!

PR #82, opened 2021-03-18
Add a test plan for spec constants

gmlueck:
We are starting to write test pans for SYCL 2020 features before
writing actual conformance tests.  This is at an example test plan,
which serves as an example plan and also describes the tests needed
for the specialization constant API.

### DIFF
--- a/test_plans/README.md
+++ b/test_plans/README.md
@@ -1,0 +1,13 @@
+This directory contains test plans for new features in SYCL 2020.  A test plan
+is simply a description of the tests that we plan to add to the CTS.  Since the
+test plans are more concise than the actual test, we can review and change them
+more quickly that reviewing and changing the actual tests.  Once a plan has
+been approved, we can write the tests according to the plan.
+
+Test plans do not have a strict format.  A plan should have enough information
+to tell which APIs will be tested and to explain how they will be tested.  The
+goal of a plan is to get feedback, so include enough detail for the reader to
+understand the core purpose of each test.  Most test plans have one section
+for each test, and each section has a bulleted list that describes that test.
+When writing a new test plan, it's often easiest to use an existing plan as a
+model.
--- a/test_plans/README.md
+++ b/test_plans/README.md
@@ -1,7 +1,7 @@
 This directory contains test plans for new features in SYCL 2020.  A test plan
 is simply a description of the tests that we plan to add to the CTS.  Since the
 test plans are more concise than the actual test, we can review and change them
-more quickly that reviewing and changing the actual tests.  Once a plan has
+more quickly than reviewing and changing the actual tests.  Once a plan has
 been approved, we can write the tests according to the plan.
 
 Test plans do not have a strict format.  A plan should have enough information
--- a/test_plans/spec-constants.asciidoc
+++ b/test_plans/spec-constants.asciidoc
@@ -219,43 +219,108 @@ constant's default value.
 [[sec:declarations]]
 === Spec constant defined in various ways
 
-Do the following test for a `specialization_id` variable defined in the
+Create an application with `specialization_id` variables defined in the
 following ways:
 
-* Defined in a non-global namespace.
-* Defined in the global namespace as `const`.
-* Defined in the global namespace as `constexpr`.
-* Defined in the global namespace as `inline`.
-* Defined in the global namespace as `inline const`.
+* Defined in a non-global named namespace.
+* Defined in an unnamed namespace.
 * Defined in the global namespace as `inline constexpr`.
+* Defined in the global namespace as `static constexpr`.
 * A static member variable of a struct in the global namespace.
 * A static member variable of a struct in a non-global namespace.
-* A static member variable declared `const` of a struct in the global namespace.
-* A static member variable declared `constexpr` of a struct in the global
-  namespace.
-* A static member variable declared `inline` of a struct in the global
-  namespace.
-* A static member variable declared `inline const` of a struct in the global
-  namespace.
+* A static member variable of a struct in an unnamed namespace.
 * A static member variable declared `inline constexpr` of a struct in the global
   namespace.
+* A static member variable of a templated struct in the global namespace.
 
-The test that is performed is:
+Perform the following test:
 
 * Create a `queue` from the tested device and call `queue::submit()`.
-* Set the value of the spec constant via
+* Set each spec constant to a different value via
   `handler::set_specialization_constant()`.
 * Submit a kernel via `handler::single_task()`.
-* Read the value of the spec constant via
-  `kernel_handler::get_specialization_constant()` and make sure we get the same
+* Read the value of each spec constant via
+  `kernel_handler::get_specialization_constant()` and make sure we get its
    value back.
 
 === Spec constant defined in various ways and set via `kernel_bundle`
 
-* Same test as in <<sec:declarations>>, except:
-  - Set spec constants in a `kernel_bundle`.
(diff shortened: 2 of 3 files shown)


test_plans/spec-constants.md
```diff
@@ -0,0 +1,540 @@
...
+* Set the value of the spec constant via
+  `kernel_bundle::set_specialization_constant()`.
+* Set the value again to a different value.
+* Call `kernel_bundle::get_specialization_constant()` and make sure we get the second
+   value back.
+* No kernel is submitted.
+
+#### Read a spec constant from a compiled kernel_bundle without writing its value
+
+* Compile kernel_bundle.
```
keryell, line 318: All these lines are unclear.
Do you mean just "Compile the bundle." or "Compile the bundle with `kernel_bundle::compile()`."?

test_plans/spec-constants.md
```diff
@@ -0,0 +1,540 @@
...
+* Create a `queue` from the tested device and call `queue::submit()`.
+* Set the values of some of the spec constants via
+  `handler::set_specialization_constant()` but do not set the values for all of
+  them.
+* Submit a kernel via `handler::single_task()`.
+* Read the values of the spec constant via
+  `kernel_handler::get_specialization_constant()` and make sure we get the
+  expected value from each (either the value we set or the default value).
+
+### Multiple spec constants with kernel_bundle
```
keryell, line 181:
```suggestion
### Multiple spec constants with `kernel_bundle`
```
everywhere

keryell (commented):
Quite some work to write a test plan!
In the same way that the tests might be written in intention using a lot of generic programming, is there a way to have the test plan also more in intention and less in extension? At the end it looks like some basic test sequence pattern in used on a Cartesian product of configurations which have been unrolled which make difficult to realize whether or not some cases are forgotten and obviously for the test writer to make sure that a compact implementation matches the test plan.

test_plans/README.md
```diff
@@ -0,0 +1,13 @@
+This directory contains test plans for new features in SYCL 2020.  A test plan
+is simply a description of the tests that we plan to add to the CTS.  Since the
+test plans are more concise than the actual test, we can review and change them
+more quickly that reviewing and changing the actual tests.  Once a plan has
```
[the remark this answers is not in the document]
gmlueck: Done.

test_plans/spec-constants.md
```diff
@@ -0,0 +1,540 @@
...
+that does not set a value should read the spec constant's default value.
+
+### Spec constant defined in various ways
+
+Do the following test for a `specialization_id` variable defined in the
+following ways:
+
+* Defined in a non-global namespace.
+* Defined in the global namespace as `const`.
+* Defined in the global namespace as `constexpr`.
```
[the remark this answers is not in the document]
gmlueck: Agreed.  My plan was to wait for https://github.com/KhronosGroup/SYCL-Docs/pull/133 to be merged, and then I would update this test plan.

test_plans/spec-constants.md
gmlueck: Yes, this was unclear.  I think 278702551abe5eb04b9cc53db60df01b63232693 addresses this.

test_plans/spec-constants.md
gmlueck: Done.

test_plans/spec-constants.md
```diff
@@ -0,0 +1,540 @@
...
+* Submit a kernel via `handler::single_task()`.
+* Call some helper function in the same translation unit, passing the
+  `kernel_handler` object by reference.
+* From the helper function, read the value of the spec constant via
+  `kernel_handler::get_specialization_constant()` and make sure we get the same
+   value back.
+
+#### Pass kernel handler object by value to another function
+
+* Same test as above, except pass the `kernel_handler` object by value.
```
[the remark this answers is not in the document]
gmlueck: I don't know how to enable section numbering and section cross references in markdown, so I converted the test plan to asciidoc.  References like this have now been changed to section references.

test_plans/spec-constants.asciidoc
```diff
@@ -0,0 +1,562 @@
...
+[[sec:declarations]]
+=== Spec constant defined in various ways
+
+Do the following test for a `specialization_id` variable defined in the
+following ways:
+
+* Defined in a non-global namespace.
+* Defined in the global namespace as `const`.
+* Defined in the global namespace as `constexpr`.
+* Defined in the global namespace as `inline`.
```
AlexeySachkov, line 228: This is probably illegal now, looking at KhronosGroup/SYCL-Docs#133

test_plans/spec-constants.asciidoc
```diff
@@ -0,0 +1,562 @@
...
+* Defined in the global namespace as `inline const`.
+* Defined in the global namespace as `inline constexpr`.
+* A static member variable of a struct in the global namespace.
+* A static member variable of a struct in a non-global namespace.
```
AlexeySachkov, line 232: Same here, `constexpr` is now required

test_plans/spec-constants.asciidoc
```diff
@@ -0,0 +1,562 @@
...
+
+=== Spec constant defined in various ways and set via `kernel_bundle`
+
+* Same test as in <<sec:declarations>>, except:
+  - Set spec constants in a `kernel_bundle`.
+  - Call `build()` to build the `kernel_bundle` into `executable` state.
+  - Register the bundle with a handler via `use_kernel_bundle()`.
+
+[[sec:other-tu]]
+=== Spec constant defined in another translation unit
```
AlexeySachkov, line 261: What if we have two non-`inline` `specialization_id` defined in different translation units with the same name? This should work and they should be treated as two different specialization constants - do we have this use case listed here somewhere?

yuriykoch: According to the C++ spec "every program shall contain exactly one definition of every non-inline function or variable that is odr-used in that program outside of a discarded statement (9.4.1); no diagnostic required". Could you possibly clarify the use case?

AlexeySachkov: `specialization_id` must be defined as `constexpr` so I assumed it in my comment: `constexpr` implies that the variable will have internal linkage which won't lead to ODR violation.

test_plans/spec-constants.md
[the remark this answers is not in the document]
gmlueck: I have updated the test plan to test only `constexpr` now, so I think this is resolved.

test_plans/spec-constants.asciidoc
gmlueck: Fixed in da04e6d5eef9f5c3f9bc0f803ab656a38e3e4744

test_plans/spec-constants.asciidoc
gmlueck: Fixed in da04e6d5eef9f5c3f9bc0f803ab656a38e3e4744

test_plans/spec-constants.asciidoc
gmlueck: > What if we have two non-inline specialization_id defined in different translation units with the same name? This should work and they should be treated as two different specialization constants - do we have this use case listed here somewhere?

Good idea.  I added this test in 5bbab0683fe2229a1ef27990e5887cf405c68a7f

keryell (approved):
This looks quite deeper than SYCL 1.2.1 tests!

fraggamuffin:
updated to match spec for const expr

fraggamuffin:
LGTM

PR #83, opened 2021-03-18
[WIP] [aspect_selector] Add test plan

ProGTX:
Initial test plan for the `aspect_selector` feature.

### DIFF
--- a/test_plans/SYCL-2020/aspect_selector.md
+++ b/test_plans/SYCL-2020/aspect_selector.md
@@ -1,0 +1,102 @@
+# Test plan for the aspect selector
+
+This is the test plan for the `aspect_selector` feature of SYCL 2020.
+
+## Testing scope
+
+### Device coverage
+
+All tests construct a test device (i.e. `testDevice`),
+representing the device where conformance will be passed on.
+Each test will construct an aspect selector object,
+which will try to select another device,
+this one based on the requested aspects (i.e. `selectedDevice`).
+`selectedDevice` can be any device in the system,
+not necessarily the same as `testDevice`.
+
+### Function overloads
+
+Each test that tests aspects `{asp1, ..., aspN}` must test all three overloads of the `aspect_selector` function:
+
+* Pass a vector of aspects: `aspect_selector(std::vector{asp1, ..., aspN})`
+* Pass aspects as function arguments: `aspect_selector(asp1, ..., aspN)`
+* Pass aspects as template parameters: `aspect_selector<asp1, ..., aspN>()`
+
+### Combinations of aspects
+
+All aspects listed in the `enum class aspect` will be tested,
+let's identify these aspects as `{asp1, ..., aspN}`.
+
+The CTS will cover these combinations:
+
+1. All combinations where no aspect is passed in.
+2. All combinations where a single aspect is used.
+3. All combinations of two aspects,
+including duplicates (e.g. `aspect_selector(aspX, aspX)`),
+and reversed order (e.g. `aspect_selector(aspX, aspY)` and `aspect_selector(aspY, aspX)`).
+
+The CTS does not have to test all combinations of 3 or more aspects.
+
+The CTS will test all the different numbers of aspects at least once, up to `N`,
+where the aspects are not duplicated,
+i.e. `aspect_selector(asp1), aspect_selector(asp1, asp2), ..., aspect_selector(asp1, asp2, ..., aspN)`.
+
+Additional testing will include 100 combinations
+of randomly selected aspect combinations,
+allowing aspect duplicates,
+with a random arity.
+
+> Note: it is recommended a Python generator is written to cover all required combinations.
+
+## Tests
+
+### No aspects
+
+1. `auto selector1 = aspect_selector()`
+2. `auto selector2 = aspect_selector<>()`
+
+In the test, both `selector1` and `selector2` are used
+to construct a SYCL device, `dev1` and `dev2` respectively.
+Both `dev1` and `dev2` must compare equal to the device obtained by `default_selector`
+(let's call this device `defaultDevice`).
+
+### Requested aspects using a device
+
+Assume requested aspects `{asp1, ..., aspN}` for each test.
+
+These steps are performed for each combination:
+
+1. Construct `testDevice`, which represents the device where conformance will be passed on.
+
+    * Also retrieve `testPlatform` SYCL platform from `testDevice` and construct `testQueue` SYCL queue from the device.
+2. Test using a vector of aspects
+
+    1. Construct `selectorVector` object obtained by calling `aspect_selector(std::vector{asp1, ..., aspN})`.
+    2. Try/catch block
+
+        1. Construct SYCL device (`selectedDevice`) using `selectorVector`.
+        2. If exception is throw:
+
+            * If `testDevice` contains all the requested aspects,
+              but no device was found using the aspect selector,
+              this is a failed test.
+            * Otherwise there is most likely no device in the system with the requested aspects, skip test.
+        3. If test passed so far, there should be no exceptions anymore with further testing.
+        4. The same checks for `testPlatform` (SYCL platform from aspect selector object)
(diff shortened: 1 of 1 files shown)


test_plans/SYCL-2020/aspect_selector.md
```diff
@@ -0,0 +1,102 @@
...
+The CTS does not have to test all combinations of 3 or more aspects.
+
+The CTS will test all the different numbers of aspects at least once, up to `N`,
+where the aspects are not duplicated,
+i.e. `aspect_selector(asp1), aspect_selector(asp1, asp2), ..., aspect_selector(asp1, asp2, ..., aspN)`.
+
+Additional testing will include 100 combinations
+of randomly selected aspect combinations,
+allowing aspect duplicates,
+with a random arity.
```
gmlueck, line 47: We had an internal discussion recently about adding tests with randomized inputs to the CTS, and we thought it was not a good idea.  We agree that such tests are useful, but we're not sure they belong in the CTS.  Vendors use the CTS to demonstrate conformance, so it seems potentially unfair for the CTS to behave differently for one vendor vs. another.  Consider two vendors that have the same bug.  One might run the CTS and get a pass while another runs the CTS and gets a fail.  That seems unfair.

test_plans/SYCL-2020/aspect_selector.md
```diff
@@ -0,0 +1,102 @@
...
+    1. Construct `selectorVector` object obtained by calling `aspect_selector(std::vector{asp1, ..., aspN})`.
+    2. Try/catch block
+
+        1. Construct SYCL device (`selectedDevice`) using `selectorVector`.
+        2. If exception is throw:
+
+            * If `testDevice` contains all the requested aspects,
+              but no device was found using the aspect selector,
+              this is a failed test.
+            * Otherwise there is most likely no device in the system with the requested aspects, skip test.
```
gmlueck, line 83: Why not iterate over all the devices in the system and see if there is one with all the requested aspects?

test_plans/SYCL-2020/aspect_selector.md
```diff
@@ -0,0 +1,102 @@
...
+        3. If test passed so far, there should be no exceptions anymore with further testing.
+        4. The same checks for `testPlatform` (SYCL platform from aspect selector object)
+        and `testQueue` (SYCL queue from aspect selector object).
+
+            * Return `selectedPlatform` and `selectedQueue`, respectively.
+    3. Check that `selectedDevice` contains all requested aspects.
+
+        * `selectedDevice.has(asp1), ..., selectedDevice.has(aspN)`.
+        * The same checks for `selectedPlatform` and `selectedQueue`.
```
gmlueck, line 92: I think the test for `selectedPlatform` is not be correct.  Let's say the test does this:

```
auto selectVector = aspect_selector(std::vector{asp1});
platform p{selectVector};
```

There is **no** guarantee that `p.has(asp1)` will be true.  The `selectVector` will select some device `d` which has that aspect.  However, the platform will only have that aspect is **all** of its devices have it.  If there is some other device `d2` in that platform which does not have `asp1`, then `p.has(asp1)` will be false.

Note that there is no member function `queue::has()`.  I guess you will test `selectedQueue.get_device().has()`?

gmlueck (commented):
I think there should also be some tests for the `denyList` parameter to the overload that takes two  `std::vector` parameters.

test_plans/SYCL-2020/aspect_selector.md
keryell: On the other hand we can imagine a test-driven implementation of SYCL implementing *only* the CTS... ;-)

test_plans/SYCL-2020/aspect_selector.md
keryell: So, testing for these behaviors?

test_plans/SYCL-2020/aspect_selector.md
ProGTX: How about requiring that the generated tests must always produce the same combinations (by using a predefined seed)? I've updated that in the latest wording.

test_plans/SYCL-2020/aspect_selector.md
ProGTX: Changed it to that, separated the platform.

test_plans/SYCL-2020/aspect_selector.md
ProGTX: You're correct, the latest wording should reflect that.

test_plans/SYCL-2020/aspect_selector.md
```diff
@@ -0,0 +1,127 @@
...
+Assume requested aspects `{asp1, ..., aspN}` for each test,
+and forbidden aspects `denyList = {noasp1, ..., noaspM}`.
+
+These steps are performed before testing combinations:
+
+1. Iterate over all devices to find out if there's one that contains all requested aspects
+    but doesn't contain any forbidden ones,
+    we call this `suitableDevice`.
+
+    * If a device like this cannot be found, skip test.
```
gmlueck, line 89: Why should the test be skipped in this case?  Instead, I think you could verify that the selector throws an `errc::runtime` exception as the spec mandates.

test_plans/SYCL-2020/aspect_selector.md
```diff
@@ -0,0 +1,127 @@
...
+    1. Check that `selectedDevice` contains all requested aspects
+        and none of the forbidden ones.
+
+        * `selectedDevice.has(asp1), ..., selectedDevice.has(aspN)`.
+        * `!selectedDevice.has(noasp1), ..., !selectedDevice.has(noaspN)`.
+
+            * If `denyList` is not empty.
+        * The same checks for `selectedQueue.get_device()`.
+    1. If `suitablePlatform` was found, construct SYCL platform (`selectedPlatform`) using `selectorVector`.
+    1. Check that `selectedPlatform` contains all requested aspects.
```
gmlueck, line 114: I think this is not correct.  Here is the definition of the platform constructor you are testing:

```
template <typename DeviceSelector>
explicit platform(const DeviceSelector &deviceSelector);
```

> Constructs a SYCL platform instance that is a copy of the platform which **contains the device** returned by the device selector parameter.

(Emphasis mine.)

The returned platform is merely guaranteed to contain **a** device that has all the requested aspects.  However, this does not mean that the platform itself has all of those aspects.  This is because `platform::has(asp)` is only `true` if **all** devices in the platform have aspect `asp`.

Note that this also affects your pre-condition where you check `suitablePlatform`.

test_plans/SYCL-2020/aspect_selector.md
```diff
@@ -0,0 +1,127 @@
...
+
+For each of the combinations described so far,
+add new combinations which also use forbidden aspects `{noasp1, ..., noaspM}` -
+`M` is independent of `N`, can be smaller, same size, or larger.
+These forbidden aspects will be passed to the `denyList` argument.
+Use the same rules as above for generating aspects,
+following these extra restrictions:
+
+* `noaspY` is not a member of the set `{asp1, ..., aspN}`
+    for each `noaspY` from the set of forbidden aspects.
```
gmlueck, line 60: I think this is your intent, but you should make it clear that we will also test the case where the `denyList` is empty.

test_plans/SYCL-2020/aspect_selector.md
gmlueck: Yes, I think this is reasonable.  As long at the test runs the same for each vendor, I'm happy.

fraggamuffin:
contractor?

bader:
https://github.com/KhronosGroup/SYCL-CTS/pull/422 is merged, so I think we can close this one. @ProGTX, are you okay with that?

bader:
> #422 is merged, so I think we can close this one. @ProGTX, are you okay with that?

@ProGTX, any objections to close this PR?

ProGTX:
Closing this PR since it's been replaced.

PR #87, opened 2021-03-19
Set CMake project language to CXX

psalz:
This prevents CMake from looking for a C Compiler, and manually setting `CMAKE_C_COMPILER` for different SYCL implementations is no longer necessary.

I've targeted this PR against the 1.2.1 branch, but it obviously also applies to 2020. What is the preferred workflow here going forward? Cherry-pick this into 2020? Target 2020 and cherry-pick into 1.2.1? Should I create two separate PRs..?


ProGTX (approved):
I think it's probably best to merge this into 1.2.1 and then cherry-pick it into 2020, I believe that's what we're doing with some of the other PRs as well.

illuhad (approved):

psalz:
Pinging to friendly remind whoever is responsible that this should also be cherry-picked into `SYCL-2020`! (Alternatively I can open another PR).

bader:
> Pinging to friendly remind whoever is responsible that this should also be cherry-picked into `SYCL-2020`! (Alternatively I can open another PR).

@psalz, I don't think we have a dedicated person with such responsibility, so your help would be much appreciated.

psalz:
Done, in #107.

PR #92, opened 2021-03-22
Improve test coverage for nd_item and group async copy and wait_for

yuriykoch:
Fixed gaps:
  - added result verification for async_work_group_copy tests
  - added result verification for wait_for tests
  - added custom number of events support for wait_for tests

Extra gaps fixed:
  - cover dimensions 2 and 3 for wait_for tests

Not covered by this commit:
  - nd_item and group range, id and subscript gaps
  - nd_item and group constructors and equality operators gaps
  - nd_item and group mem_fence() and barrier() gaps
  - group::parallel_for_work_item()

This PR depends on:
- #78 - 39af5da - cherry-picked once_per_unit.h
- #80 - 68ce9a2 - merge; depends on util/array.h
- #90 - fcb2b18 - merge; depends on get_cts_object::range<>::get_fixed_size() from common/get_cts_object.h
- #91 - 78f78da - merge; depends on common/type_coverage.h

Please, review changes starting from 96591f9 - "Improve test coverage for nd_item and group async copy and wait_for"


ProGTX (approved):

keryell (commented):
I have the feeling that the more I read in the PR and the more I see again the same code.
Probably because it includes some older PR not yet merged. And I am the culprit for this since I am to slow to review.
But perhaps it is time to rebase once the oldest PR have landed?

tests/common/get_cts_object.h
```diff
@@ -192,6 +201,17 @@ struct get_cts_object::range<2> {
...
+  /**
+   * @brief Constructs a range<2> by only using first component and the
+   *        total size given
+   * @tparam totalSize Value the size() call should return for the range
+   * @param r0 Value of the first component of the range
+   * @return range<2>
+   */
+  template <size_t totalSize>
+  static cl::sycl::range<2> get_fixed_size(size_t r0, size_t) {
+    return cl::sycl::range<2>(r0, totalSize / r0);
```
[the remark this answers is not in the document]
yuriykoch: That's from PR #90, still we can address it here. 
You are right, we do assume such restriction, so I've added an assert to make this explicit. Thanks for pointing this out.
Please, review a solution.

tests/common/get_cts_object.h
```diff
@@ -209,6 +229,18 @@ struct get_cts_object::range<3> {
...
+   * @brief Constructs a range<3> by only using two components and the
+   *        total size given
+   * @tparam totalSize Value the size() call should return for the range
+   * @param r0 Value of the first component of the range
+   * @param r0 Value of the second component of the range
+   * @return range<3>
+   */
+  template <size_t totalSize>
+  static cl::sycl::range<3> get_fixed_size(size_t r0, size_t r1) {
+    return cl::sycl::range<3>(r0, r1, totalSize / r0 / r1);
```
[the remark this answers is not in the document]
yuriykoch: Updated, see above

tests/common/async_work_group_copy.h
```diff
@@ -0,0 +1,513 @@
...
+template <size_t bufferSize, typename T,
+          cl::sycl::access::address_space addressSpace>
+sycl_cts::util::array<T, bufferSize> create_async_wg_copy_input(
+    cl::sycl::multi_ptr<T, addressSpace> ptr) {
+
+  sycl_cts::util::array<T, bufferSize> result;
+
+  /** We use bool values because bool can be converted to the integral
+   *  and to the floating point types
+   */
```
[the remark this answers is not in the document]
yuriykoch: I'm using bool value because it can be implicitly converted to any other type. To use static_cast we should provide some other function so that scalar T and vector T will go through different code path. Say, for `T=vec<bool, 2>` this implementation does work without explicit `T::element_type` usage.
Should I provide some integer sequence for non-boolean types? Or may be should I use some different binary sequence, `0-1-1` for example?

tests/common/async_work_group_copy.h
```diff
@@ -0,0 +1,513 @@
...
+  return result;
+}
+
+/**
+ * @brief Retrieve stride for async_work_group_copy call
+ *        Stride is always 1 for the local_ptr
+ */
+template <size_t stride, typename T>
+inline size_t get_async_wg_copy_stride(cl::sycl::local_ptr<T> ptr) {
+  static_cast<void>(ptr);
```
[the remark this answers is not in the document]
yuriykoch: Fixed, thanks

tests/common/async_work_group_copy.h
```diff
@@ -0,0 +1,513 @@
...
+  return 1;
+}
+/**
+ * @brief Retrieve stride for async_work_group_copy call
+ *        Stride is always the given one for the global_ptr
+ */
+template <size_t stride, typename T>
+inline size_t get_async_wg_copy_stride(cl::sycl::global_ptr<T> ptr) {
+  static_cast<void>(ptr);
```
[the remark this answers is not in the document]
yuriykoch: Fixed

tests/common/async_work_group_copy.h
```diff
@@ -0,0 +1,513 @@
...
+  // Use async_work_group_copy
+  auto event = constInstance.async_work_group_copy(dstPtr, srcPtr,
+                                                   numElements, callArgs...);
+  constInstance.wait_for(event);
+
+  bool succeed = true;
+  auto input = srcPtr;
+  auto output = dstPtr;
+
+  // Verify source not changed
```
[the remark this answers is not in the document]
yuriykoch: Why not?) That's async copy, so we can check if implementation doesn't use an overlapping buffers at some point.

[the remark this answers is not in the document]
keryell: Indeed!

tests/group/group_async_work_group_copy_fp16.cpp
```diff
@@ -23,17 +26,26 @@ class TEST_NAME : public util::test_base {
...
+      }
+      // Test using queue constructed already
+      for_type_and_vectors<check_type, cl::sycl::half>(queue, log,
+          "cl::sycl::half");
+      for_type_and_vectors<check_type, cl::sycl::cl_half>(queue, log,
+          "cl::sycl::cl_half");
+
+    } catch (const cl::sycl::exception &e) {
+      log_exception(log, e);
+      cl::sycl::string_class errorMsg =
```
keryell, line 45: Just use normal strings everywhere

util/variadic.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include <cstddef>
+#include <type_traits>
+
+/**
+ * @brief Can be used to run functors with parameter pack and custom parameters
+ * @tparam functorT Functor to use
+ */
+template <class functorT>
+struct run_variadic {
```
keryell, line 20: I have to think a little bit further about what is going on here with all this SFINAE.
Adding an example of use in the comment above, instead of grep-ing for it through the code.
Could it be clearer with some `if constexpr` and `...` interpolation instead of this recursion?

yuriykoch: Sorry, I'm not quite sure I've got your idea. We are unpacking single parameter into the comma-delimited list of parameters here.
I've provided an usage example using doxygen markdown for code blocks.

tests/group/group_async_work_group_copy_fp16.cpp
yuriykoch: Yes, but the codebase still uses deprecated SYCL aliases. Wouldn't it be better to replace all of them in a separate PR?

util/variadic.h
```diff
@@ -0,0 +1,95 @@
...
+   * @tparam T Type of params storage
+   * @tparam argsT Type of arguments to forward into the functor call
+   * @param params Parameters storage used for recursion
+   * @param args Arguments to forward into the functor call
+   */
+  template <size_t nParams, typename T, typename ... argsT>
+  typename std::enable_if<(nParams == 0), typename functorT::returnT>::type
+    static inline with(const T& params, argsT ... args) {
+      static_cast<void>(params);
+      return functorT{}(args...);
```
keryell, line 62:
```suggestion
  typename std::enable_if_t<nParams == 0, typename functorT::returnT>
    static inline with(const T&, argsT ... args) {
      return functorT{}(args...);
```
Not clear whether you need `std::forward` on the `args`.

util/variadic.h
```diff
@@ -0,0 +1,95 @@
...
+   * @brief Extract strict number of parameters from storage into the parameter
+   *        pack and call functor with parameter pack.
+   * @tparam nParams Number of params to use as parameter pack
+   * @tparam T Type of params storage
+   * @param params Storage to extract parameters from
+   */
+  template <size_t nParams, typename T>
+  typename std::enable_if<(nParams > 0), typename functorT::returnT>::type
+    static inline with(T&& params) {
+      return with<nParams-1>(params, params[nParams-1]);
```
keryell, line 91: I see.
Now with your documentation I understand better. :-)
Not fundamental but I have the feeling that you can do this without recursion and in 2 functions instead of 3 by using `std::make_index_sequence<nParams>` like in https://github.com/triSYCL/triSYCL/blob/master/include/triSYCL/detail/array_tuple_helpers.hpp

keryell (approved):
Thanks.

util/variadic.h
yuriykoch: @keryell You are right, `make_index_sequence` fits great. Thanks for awesome catch.
I was able to add some sugar actually, see a93bfd4

util/variadic.h
yuriykoch: Done, thanks

util/variadic.h
yuriykoch: Discussion continued and resolved below

tests/group/group_async_work_group_copy_fp16.cpp
yuriykoch: Done for all files touched by this change. See f1df84b

keryell:
Thanks for the clarification!

util/variadic.h
```diff
@@ -0,0 +1,92 @@
...
+ *
+ *              run_variadic<run_wait_for<args>>::with<32>(events, a, b)
+ *
+ *          Also we are able to use a selected subset of events
+ *
+ *              run_wait_for{}(events[1], events[3], events[7])
+ *
+ *          by using std::integer_sequence like
+ *
+ *              std::integer_sequence<std::size_t, 1, 3, 7> selected;
```
keryell, line 54: You are going beyond my expectations with such an API :-)

PR #101, opened 2021-03-29
Add atomic64 type traits

yuriykoch:
Provides has_atomic_support and bits_eq type traits to make possible separate code path for the atomic64 types at the compile time.

This PR depends on:
#77 - 5805017 - util/type_traits.h was added

Please, review changes starting from b1af951 - "Add has_atomic_support and bits_eq type traits"


util/type_traits.h
```diff
@@ -0,0 +1,69 @@
...
+                     contains<T, tailT...>>::type {};
+
+/**
+ * @brief Verify type has the given number of bits
+ */
+template <typename T, size_t bits>
+struct bits_eq :
+    std::conditional<(sizeof(T) * CHAR_BIT) == bits,
+                     std::true_type,
+                     std::false_type>::type {};
```
ProGTX, line 40:
```suggestion
template <typename T, size_t bits>
using bits_eq = std::bool_constant<(sizeof(T) * CHAR_BIT) == bits>;
```

Prefer writing type traits in terms of aliases wherever possible - it's clearer and helps with compilation times.

util/type_traits.h
```diff
@@ -0,0 +1,69 @@
...
+    contains<T,
+             int, unsigned int,
+             long, unsigned long,
+             long long, unsigned long long,
+             float>;
+
+/**
+ * @brief Checks whether T is a floating-point sycl type
+ */
+template <typename T> struct is_cl_float_type {
```
ProGTX, line 59: This could be `bool_constant` as well

ProGTX (approved):

util/type_traits.h
yuriykoch: Great catch, thanks

util/type_traits.h
yuriykoch: Fixed

yuriykoch:
All dependencies are merged. PR contains only relevant changes now.

PR #108, opened 2021-04-13
Add sycl = ::cl::sycl namespace alias

psalz:
SYCL 2020 introduces a new canonical namespace, `::sycl`, which is defined within the new `<sycl/sycl.hpp>` header. To ease the transition of the CTS to the new namespace, this creates an alias `sycl = cl::sycl`, while keeping the SYCL 1.2.1 header path `<CL/sycl.hpp>` for the time being.

For the record, I also tried going the reverse route, i.e., including `<sycl/sycl.hpp>`, and defining `namespace cl { namespace sycl = ::sycl; }`. However that didn't work, as DPC++ already exports that alias (while e.g. hipSYCL doesn't - the spec isn't super clear on this), and ComputeCpp 2.4 doesn't seem to support the new header at all.

### DIFF
--- a/tests/common/sycl.h
+++ b/tests/common/sycl.h
@@ -21,4 +21,8 @@
 #include <CL/sycl.hpp>
 #endif
 
+// TODO: Switch to canonical <sycl/sycl.hpp> header
+// once CTS has been fully converted to ::sycl namespace
+namespace sycl = ::cl::sycl;
+
 #endif  // __SYCLCTS_TESTS_COMMON_H


ProGTX (approved):

bader:
To clarify: I've added "Agenda" tag to discuss this description section:

> For the record, I also tried going the reverse route, i.e., including `<sycl/sycl.hpp>`, and defining `namespace cl { namespace sycl = ::sycl; }`. However that didn't work, as DPC++ already exports that alias (while e.g. hipSYCL doesn't - the spec isn't super clear on this), and ComputeCpp 2.4 doesn't seem to support the new header at all.

bader (approved):

keryell:
This would work with triSYCL.

bader:
> This would work with triSYCL.

@keryell, can we interpret this comment as "Approved" from your side? :)

keryell:
> > This would work with triSYCL.
> 
> @keryell, can we interpret this comment as "Approved" from your side? :)

Not really.
In theory yes for careful implementation with small footprint not trying to pollute all the namespaces around. :-)
Do you plan to update DPC++ accordingly to reduce its namespace pollution?
Perhaps we can discuss this inside the SYCL committee meeting.

keryell:
I have the feeling that a CTS should test exactly what is in the spec and not define some aliases or new indirect namespaces because it might have some side effect, like noticed by @psalz.

keryell:
Do we need to validate in the CTS that `cl::sycl` works as a deprecated compatibility mode?
I do not think we need.
Then, why not just changing in the CTS 2020 all the `cl::sycl::` by `sycl::` and do not set up an alias?

psalz:
> Then, why not just changing in the CTS 2020 all the `cl::sycl::` by `sycl::` and do not set up an alias?

That is of course also an option, I just thought that a more gradual change would reduce noise in the repo history. But it's all the same to me, really.

keryell:
By looking at it again:

> For the record, I also tried going the reverse route, i.e., including `<sycl/sycl.hpp>`, and defining `namespace cl { namespace sycl = ::sycl; }`. However that didn't work, as DPC++ already exports that alias (while e.g. hipSYCL doesn't - the spec isn't super clear on this), and ComputeCpp 2.4 doesn't seem to support the new header at all.

is the right hack to use for now, before changing globally `cl::sycl::` to `sycl::`.
DPC++ should not pollute the global namespace and ComputeCpp should support `<sycl/sycl.hpp>`, right?

keryell (approved):

PR #109, opened 2021-04-14
Improve event_info tests

psalz:
This improves the tests for `event::get_info` and `event::get_profiling_info`:

- Remove query for deprecated `sycl::info::event::reference_count`.
- Check that command execution status is updated correctly without prior call to `event::wait`. This was the initial reason for taking this up, as the DPC++ CUDA backend currently has this bug.
- Update return types of `event::get_profiling_info` for SYCL 2020.
- Do basic sanity checks on profiling timestamps.
- Transition everything to the new `::sycl` namespace (for which I added a local alias, as to not depend on #108).

### DIFF
--- a/tests/event/event_info.cpp
+++ b/tests/event/event_info.cpp
@@ -13,9 +13,10 @@
 namespace TEST_NAMESPACE {
 
 using namespace sycl_cts;
+namespace sycl = cl::sycl;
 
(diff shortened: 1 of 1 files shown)


bader (approved):

ProGTX (approved):

keryell (approved):
That looks better

PR #111, opened 2021-04-22
Extend type coverage for address_space tests

yuriykoch:
Also fixes false positives for test_duplication


ProGTX (approved):

tests/address_space/address_space_common.h
```diff
@@ -0,0 +1,198 @@
...
+public:
+  template <cl::sycl::access::address_space kAs>
+  struct AddrSpace {
+    constexpr T get_value() const { return static_cast<T>(to_integral(kAs)); }
+  };
+
+  // the parameter has no address space specification, so it'll
+  // get duplicated based on the argument's address space
+  // it should call the appropiate overload of f(), and that will
+  // tell us the address space the parameter was in
```
keryell, line 44: By looking at the uncommented following code I have no idea about how it works...

keryell (changes requested):

keryell (changes requested):
Please comment the code, specially for this kind of non trivial code about the most complex things of the SYCL specification.

tests/address_space/address_space_common.h
```diff
@@ -0,0 +1,198 @@
...
+    T readValue(T *p) { return *p; }
+
+    T readValue(T &p) { return p; }
+
+    T operator[](T *p) { return *p; }
+
+    T operator[](T &p) { return p; }
+  };
+
+  template <cl::sycl::access::address_space AspSpace>
```
keryell, line 60: Not clear to me the meaning of `AspSpace`. What `Asp` is for? Something more precise than `Addr`?

fraggamuffin:
Implementers should have a look for next week to decide
Need more comments to justify or explain

keryell:
@Naghasan any inspirational comment?

keryell:
Meeting 2021/07/08:

Still need to dive into this...

Naghasan:
@keryell sorry, finally got around to review. This test was added to test the multi_ptr API and some of the type deduction of 1.2.1. The type deduction is what make things confusing, some test are only there to trigger an inference of the type.. IIRC, this morphed quite a few times before being release publicly.

The code could be simplified to check for the interface only IMO.

tests/address_space/address_space_common.h
```diff
@@ -0,0 +1,198 @@
...
+    return result;
+  }
+
+  template <cl::sycl::access::address_space AspSpace>
+  static T *id(cl::sycl::multi_ptr<T, AspSpace> p) {
+    return p;
+  }
+
+  static bool test_return_type_deduction(cl::sycl::global_ptr<T> globalPtr,
+                                  cl::sycl::local_ptr<T> localPtr,
```
keryell, line 94: Indent

tests/address_space/address_space_common.h
```diff
@@ -0,0 +1,198 @@
...
+    EXPECT_ADDRSPACE_EQUALS(d.readValue(ptr), AspSpace);
+    EXPECT_ADDRSPACE_EQUALS(d.readValue(*ptr), AspSpace);
+    EXPECT_ADDRSPACE_EQUALS(d[ptr], AspSpace);
+    EXPECT_ADDRSPACE_EQUALS(d[*ptr], AspSpace);
+
+    return true;
+  }
+
+  static bool test_duplication(cl::sycl::global_ptr<T> globalPtr,
+                        cl::sycl::local_ptr<T> localPtr,
```
keryell, line 75: Indent.
It looks like most of the function declaration have weird indent.

keryell (approved):
OK, I think I start understanding what is going on here.
What is super obtuse is that there are tests about address space deduction for variables of various types and the value of the address space itself is returned as a value of the tested type.
For example `cl::sycl::access::address_space::private_space` is cast to a `double` or a `char` or a...
Not clear why this obfuscation instead of just returning an `cl::sycl::access::address_space` aside...
The problem is that it should be clarified & restructured some day, but outside of this PR.

bader (commented):
@yuriykoch, please, resolve merge conflicts in addition to applying suggestions from Ronan.

yuriykoch:
> Not clear why this obfuscation instead of just returning an cl::sycl::access::address_space aside...

Agree, looks like the logic of checks in `address_space.cpp` I've moved to the `address_space_common.h` was broken due to historical reasons. Sorry for missing it anyway. We may have type coverage enabled though..

All these tests do the following:
•	Initialize array with 4 integer values: [#L123](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/address_space/address_space.cpp#L123)
•	Create 4 multi_ptr instances pointing to 4 array elements: [#L160](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/address_space/address_space.cpp#L160)
•	Verify value in multi_ptr is the same one used for initialization: [#L19](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/address_space/address_space.cpp#L19)
in different ways

__test_initialization__
•	`int *p = multi_ptr;`
o	call “operator pointer() const;” for multi_ptr instance
•	`compare (*p) == int`

__test_return_type_deduction__
•	`int *p = multi_ptr`
o	Same as above, but using function to call this method
•	`compare (*p) == int`
o	Same as above, but using temporary instance
This test has the same logic as test_initialization. It’s guaranteed by C++ spec.

__test_duplication__
•	`int *p = multi_ptr (same as above)`
o	Using int readValue(int *p) { return *p; }
o	Using int readValue(int &p) { return p; }
o	Using same readValue methods, but within a struct  
o	Using same methods with different names
•	`compare (*p) == int`
o	Same way as above
This test also has the same logic as test_initialization. It’s guaranteed by C++ spec.

@keryell You can take a look into [this](https://github.com/KhronosGroup/SYCL-CTS/blob/a585e2cdfa3264235a3842e28b089484eee8f32d/tests/address_space/address_space.cpp) and [later](https://github.com/KhronosGroup/SYCL-CTS/blob/6492c211e343dcd7a38a443bd0d7ee26f672be7b/tests/address_space/address_space.cpp) commits for historic perspective of this code.

tests/address_space/address_space_common.h
yuriykoch: Sorry, applied clang-format in b1aec96c504d3f4de726ba75404ba498e418b8ef

tests/address_space/address_space_common.h
yuriykoch: Fixed by b1aec96c504d3f4de726ba75404ba498e418b8ef

keryell:
> All these tests do the following:

Since you have been courageous enough to write it, add the descriptions to their respective tests as comments before we land this. :-)

keryell (approved):

bader:
> > All these tests do the following:
> 
> Since you have been courageous enough to write it, add the descriptions to their respective tests as comments before we land this. :-)

I'll wait for that...

PR #112, opened 2021-04-27
Fix atomic tests for 'long' types

yuriykoch:
Both 'long' and 'unsigned long' types may have different number of bits for different data models


ProGTX (approved):

keryell (commented):
I have the feeling the code is obscured by a lot of SFINAE instead of just using some `if constexpr`.

yuriykoch:
> I have the feeling the code is obscured by a lot of SFINAE instead of just using some `if constexpr`.

@keryell Great! Thanks for advice. The code looks much better now.

keryell (approved):
Thanks!

PR #114, opened 2021-04-29
Add result verification for item tests

kovshikovser:


tests/item/item_1d.cpp
```diff
@@ -24,32 +24,42 @@ class kernel_item_1d {
...
     /* find the array id for this work item */
     size_t index = gid.get(0); /* x */
 
     /* get the linear id */
     const size_t glid = item.get_linear_id();
+    result &= m_x[int(glid)] == static_cast<int>(index);
 
     /* compare against the precomputed index */
-    m_o[int(glid)] = (m_x[int(glid)] == static_cast<int>(index));
+    m_o[int(glid)] = result;
```
ProGTX, line 62: I think a better way of doing this would be to store the results of each check separately, right now the information is lost as to which check failed. However I think that's outside of the scope of this PR, just something to keep in mind.

tests/item/item_1d.cpp
```diff
@@ -24,32 +24,42 @@ class kernel_item_1d {
...
+      : m_x(in_), m_o(out_), r_exp(r), offset_exp(cl::sycl::id<1>(0)) {}
 
   void operator()(cl::sycl::item<1> item) const {
+    bool result = true;
+
     cl::sycl::id<1> gid = item.get_id();
 
     size_t dim_a = item.get_id(0);
     size_t dim_b = item[0];
+    result &= gid.get(0) == dim_a && gid.get(0) == dim_b;
```
ProGTX, line 41:
```suggestion
    result &= (gid.get(0) == dim_a) && (gid.get(0) == dim_b);
```

Just to make it a bit easier to read

ProGTX (approved):

keryell (approved):

PR #117, opened 2021-05-12
Add vector functions as(), convert() correctness check

kopylovanat:


keryell (changes requested):
Simplify, now we switched to C++17.

tests/common/common_vec.h
```diff
@@ -136,28 +136,246 @@ bool check_vector_get_count_get_size(cl::sycl::vec<vecType, 1> inputVec) {
...
+      for (size_t i = 0; i < N; ++i) {
+        vecType elem = getElement(inputVec, i);
+        auto elemConvert = float_map_match(floats, vals, size, elem);
+        setElement<convertType, N>(resVec, i, elemConvert);
+      }
+      return resVec;
+    }
+  else {
+    return convert_vec<vecType, N, convertType>(inputVec);
+  }
```
ProGTX, line 193: This file might need some formatting

ProGTX (approved):

keryell (approved):
Thanks, that looks simpler.

tests/common/common_vec.h
ProGTX: I don't think they're useless, but we might be getting closer to a formatting holy war :) I'd say that we just need to keep the formatting consistent, and clang-format should take care of that.

kopylovanat: But that formatting is result of applying clang-format to this code (including constexpr on the different line from if) should it be changed?

keryell: Ah fine, if this is the output of `clang-format`. I just need to get used to this strange `if constexpr`.
By the way, someone should look some day at the top `.clang-format` to see how it fits with the recent `clang-format`.

bader: The formatting is definitely off here. 
Here is clang-format (trunk version) output: https://godbolt.org/z/8cM91KT3j

kopylovanat: I apologize, it was result of old version of `clang-format`. Updated with fixed formatting in 846a2bda1824821d3b295c69417eba7ff20da618

bader: > And also removing useless `{ }` like in the LLVM project?

To address this comment we can remove else-branch completely. if-branch ends with `return`, it safe to write like this:
```suggestion
  } // end of if-branch
  return convert_vec<vecType, N, convertType>(inputVec);
```

kopylovanat: Done.

bader (approved):
Thanks!

tests/common/common_vec.h
```diff
@@ -187,123 +168,123 @@ cl::sycl::vec<convertType, N> convert_vec(cl::sycl::vec<vecType, N> inputVec) {
...
-                            static_cast<convertType>(-2),
-                            static_cast<convertType>(-2)};
-  cl::sycl::vec<convertType, N> resVec;
-  for (size_t i = 0; i < N; ++i) {
-    vecType elem = getElement(inputVec, i);
-    auto elemConvert = float_map_match(floats, vals, size, elem);
-    setElement<convertType, N>(resVec, i, elemConvert);
+cl::sycl::vec<convertType, N> rte(cl::sycl::vec<vecType, N> inputVec) {
+  if
+    constexpr(if_FP_to_non_FP_conv_v<vecType, convertType>) {
```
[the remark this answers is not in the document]
keryell: OK, I understand better this science-fiction presentation: the obsolete `clang-format` had no idea of what is an `if constexpr`...

PR #118, opened 2021-05-12
Provide result verification for synchronous exceptions

yuriykoch:
Verify exception API calls and basic exception usage for synchronous errors

### DIFF
--- a/tests/exception_handling/synchronous_exceptions.cpp
+++ b/tests/exception_handling/synchronous_exceptions.cpp
@@ -8,10 +8,89 @@
 
 #include "../common/common.h"
 
+#include <limits>
+
 #define TEST_NAME synchronous_exceptions
 
 namespace TEST_NAMESPACE {
 using namespace sycl_cts;
+
+class custom_exception : public sycl::exception {
+ public:
+  struct reference {
+    static std::string what() {
+      return "custom_exception";
+    }
+    static std::error_code code() {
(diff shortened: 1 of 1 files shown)


tests/exception_handling/synchronous_exceptions.cpp
```diff
@@ -8,11 +8,90 @@
...
+  auto ctx = q.get_context();
+
+  try {
+    q.submit([&](cl::sycl::handler &cgh) {
+      cgh.single_task<class TEST_NAME>([=]() {});
+
+      throw custom_exception(ctx);
+    });
+    q.wait_and_throw();
+  } catch (const custom_exception &e) {
```
keryell, line 48: Just curious here...
Are you expecting that the exception is exiting the `submit` or is just captured and propagated to the `q.wait_and_throw()`?
My interpretation of the standard is that the exception should exit directly from the `submit`.
So, could we just replace `q.wait_and_throw()` by `q.wait()` or even remove the wait at all?

keryell (commented):
The synchronous exception testing looks too much like an asynchronous exception testing.

tests/exception_handling/synchronous_exceptions.cpp
yuriykoch: You are right, it's better to remove the blocking wait. Done by 214192da31bb

ProGTX (approved):

keryell (approved):

PR #122, opened 2021-05-18
Use size_t data for buffer size calculation

bader:
`unsigned` is not supposed to hold memory size, so using it that way
leads to incorrect memory size values due to overflow on certain
platforms where sizeof(unsigned) < sizeof(size_t).

Patch by Rajiv Deodhar.

### DIFF
--- a/tests/buffer/buffer_api_common.h
+++ b/tests/buffer/buffer_api_common.h
@@ -29,12 +29,12 @@ class empty_kernel {
 template <size_t dims>
 inline void precalculate(cl::sycl::range<dims> &rangeIn,
                          cl::sycl::range<dims> &rangeOut, size_t &elementsCount,
-                         unsigned elementsIn, unsigned elementsOut);
+                         size_t elementsIn, size_t elementsOut);
 
 template <>
 inline void precalculate<1>(cl::sycl::range<1> &rangeIn,
                             cl::sycl::range<1> &rangeOut, size_t &elementsCount,
-                            unsigned elementsIn, unsigned elementsOut) {
+                            size_t elementsIn, size_t elementsOut) {
   rangeIn = cl::sycl::range<1>(elementsIn);
   rangeOut = cl::sycl::range<1>(elementsOut);
   elementsCount = elementsOut;
@@ -43,7 +43,7 @@ inline void precalculate<1>(cl::sycl::range<1> &rangeIn,
 template <>
 inline void precalculate<2>(cl::sycl::range<2> &rangeIn,
                             cl::sycl::range<2> &rangeOut, size_t &elementsCount,
-                            unsigned elementsIn, unsigned elementsOut) {
+                            size_t elementsIn, size_t elementsOut) {
   rangeIn = cl::sycl::range<2>(elementsIn, elementsIn);
   rangeOut = cl::sycl::range<2>(elementsOut, elementsIn);
   elementsCount = (elementsOut * elementsIn);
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

tests/buffer/buffer_api_common.h
```diff
@@ -52,12 +52,33 @@ inline void precalculate<2>(cl::sycl::range<2> &rangeIn,
...
+template <>
+inline size_t calc_elem_count<2>(size_t numElemsPerDim) {
+  return numElemsPerDim * numElemsPerDim;
+}
+
+template <>
+inline size_t calc_elem_count<3>(size_t numElemsPerDim) {
+  return numElemsPerDim * numElemsPerDim * numElemsPerDim;
+}
+
```
keryell, line 81:
```suggestion
template <int dims>
inline size_t calc_elem_count(size_t numElemsPerDim) {
  return std::pow(numElemsPerDim, dims);
}
```
is painful enough.
I wonder whether we actually need a function template instead of a function...

keryell: OK remove this...

tests/buffer/buffer_api_common.h
```diff
@@ -436,12 +457,11 @@ void test_buffer(util::logger &log, cl::sycl::range<dims> &r,
 template <typename T, int numDims>
 void test_type_reinterpret(util::logger &log) {
   static constexpr size_t inputElemsPerDim = 4;
-  std::vector<uint8_t> reinterpretInputData(sizeof(T) * inputElemsPerDim *
-                                            numDims);
+  auto numElems = calc_elem_count<numDims>(inputElemsPerDim);
```
keryell, line 460: ... and just use:
```suggestion
  auto numElems = std::pow(inputElemsPerDim, dims);;
```

keryell (changes requested):
Think "simple".

tests/buffer/buffer_api_common.h
bader: `std::pow` uses one of floating point types, which doesn't seem to be a good fit here.

keryell: Ah yes, I have interpreted the case (7) of https://en.cppreference.com/w/cpp/numeric/math/pow in a too optimistic way. :-(
But since the template stuff is not even used to produce a `constexpr` value, just use a loop?
```c++
auto numElems = inputElemsPerDim;
for (int i = 1; i < numDims; ++i)
  numElems *= inputElemsPerDim;
```

bader: Done.

tests/buffer/buffer_api_common.h
```diff
@@ -436,12 +436,13 @@ void test_buffer(util::logger &log, cl::sycl::range<dims> &r,
 template <typename T, int numDims>
 void test_type_reinterpret(util::logger &log) {
   static constexpr size_t inputElemsPerDim = 4;
-  std::vector<uint8_t> reinterpretInputData(sizeof(T) * inputElemsPerDim *
-                                            numDims);
+  auto numElems = inputElemsPerDim;
+  for (int i = 1; i < numDims; ++i)
+    numElems *= inputElemsPerDim;
```
ProGTX, line 441:
```suggestion
  for (int i = 1; i < numDims; ++i) {
    numElems *= inputElemsPerDim;
  }
```

We could argue about what is better, but please keep the braces at least for consistency with other code

keryell (approved):

PR #124, opened 2021-05-20
Improve result verification for h_item API

yuriykoch:
Local id values are stored from each work-group asynchronously.
We are using the fact that all bitwise-equal values go into the same memory to simplify the test logic.

### DIFF
--- a/tests/h_item/h_item_api.cpp
+++ b/tests/h_item/h_item_api.cpp
@@ -6,11 +6,15 @@
 //
 *******************************************************************************/
 #include "../common/common.h"
+#include <string>
 
 #define TEST_NAME h_item_api
 
+namespace TEST_NAMESPACE {
+using namespace sycl_cts;
+
 template <int numDims>
-struct h_item_api_kernel_common;
+struct kernel_common;
 
 /**
  * @brief Kernel struct that tests retrieving a single value from a range or ID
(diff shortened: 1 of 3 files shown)


ProGTX (approved):

tests/h_item/h_item_api.cpp
```diff
@@ -19,26 +23,78 @@ struct h_item_api_kernel_common;
...
+            auto expected = item.get_physical_local_range()[currentDim];
+            success &= value == expected;
+          }
+          {
+            auto value = item.get_physical_local_id(currentDim);
+            auto expected = item.get_physical_local_id()[currentDim];
+            success &= value == expected;
+          }
+
+          if (!success) success_acc[0] = false;
```
keryell, line 78:
```suggestion
         *success_acc = success;
```

tests/h_item/h_item_api.cpp
```diff
@@ -19,26 +23,78 @@ struct h_item_api_kernel_common;
...
         });
   }
 };
 
-namespace TEST_NAME {
-using namespace sycl_cts;
+template <template <int> class T, int numDims>
```
keryell, line 83: I have the feeling that these functions could go into a global utility part of the infrastructure, if not existing already/

tests/h_item/h_item_api.cpp
```diff
@@ -56,94 +112,203 @@ class TEST_NAME : public util::test_base {
       auto testQueue = util::get_cts_object::queue();
 
       const auto kernelGroupRange =
-          util::get_cts_object::range<numDims>::get(8, 4, 2);
+          util::get_cts_object::range<numDims>::get(2, 2, 2);
```
keryell, line 115: I wonder whether having the same value everywhere could not hide some bugs with an implemention getting confused about x/y/z ordering...

yuriykoch: Thanks. Done by e16addc

tests/h_item/h_item_api.cpp
yuriykoch: We should change value from `true` to `false` only. It's required to avoid data race. I've added comment to make it explicit in e16addc. Please, review.

tests/h_item/h_item_api.cpp
yuriykoch: I've removed these functions in bde9441

tests/h_item/h_item_api.cpp
keryell: Oh, I see. Actually it is not even about a race condition but just about the algorithm that I completely got wrong... :-(
But since you are mentioning race conditions, I have the feeling that in OpenCL it is UB to have different work-item writing at the same place, even if they are writing the same value. I guess too that it will work anyway.
Good news, in SYCL 2020 we have reductions for this... :-)

yuriykoch: I guess this comment goes to the IDs storage logic also. Just to make sure - should I change algorithm?

keryell: Yes, that is a problem...
Let's discuss this during next SYCL committee meeting and look at where the discussion goes before changing anything.

bader: I think @keryell is right that this is UB. This is a race condition in C++ memory model.
@yuriykoch, let's allocate one result element per WI and reduce on the host.

Pennycook: I agree with @bader and @keryell.

It's a little surprising, but this is UB according to http://eel.is/c++draft/intro.multithread#intro.races-21:

> The execution of a program contains a data race if it contains two potentially concurrent conflicting actions, at least one of which is not atomic, and neither happens before the other, except for the special case for signal handlers described below. Any such data race results in undefined behavior.

This [paper](https://www.usenix.org/legacy/events/hotpar11/tech/final_files/Boehm.pdf) by Hans Boehm also calls out the issue being discussed here explicitly as being problematic:

> ...all kinds of C or C++ source-level "benign" races discussed in the literature can in fact lead to incorrect execution as a result of perfectly reasonable compiler transformations, or when the program is moved to a different hardware platform. ... Perhaps most surprisingly, this includes even the case of potentially concurrent writes of the same value by different threads.

keryell: Yes, this is an interesting (scary) paper from Hans Boehm. For more citation context it comes from https://www.usenix.org/conference/hotpar-11/how-miscompile-programs-benign-data-races

yuriykoch: Just to make sure - couldn't we just use atomic access mode to ensure there would be no UB nevertheless if the tests succeeds or fails due - for example - of some duplicated global or logical IDs?
I mean:
  - we do also have the same problem for global_acc, logical_acc and physical_acc
  - for a case of invalid global linear id there is no way to tell how to avoid UB due to writing at the same offset from within WI
  - strictly speaking UB is not limited to the broken data values or abort() call or anything, so we cannot expect test to have any valid output in case of duplicated IDs
  - there were a plenty of tests using something like "if (!success) accessor[0] = false" originally, so may be we should just use atomic when we see it to quick-fix without test logic changes

What do you think?

keryell: We could but atomics are an optional feature...

yuriykoch: Is it optional for the int/uint though? We could fit into these types instead of bool and size_t with some additional checks on ID values due possible truncation of invalid ones.

bader: I don't think it's a good idea to use features that are not supposed to be covered by the test and if they can be avoided.
Adding dependency on supporting atomics here seems unnecessary to me.
@yuriykoch, what problems do you see with the approach proposed above?

yuriykoch: @bader I agree, that's a huge cons against atomic usage.
The main issue is that we don't remove UB completely. We are testing h_item API, so we would expect test to provide some output in case there are invalid IDs. For example, in case all work items return zero ID there still will be UB by writing at the same zero location. So we may not expect any proper failure report in such case.
If that's not critical I would use logical local ID and group ID to provide unique offset for every execution to avoid failure silencing at least.

bader: The main point of this discussion is to remove UB from **the test code**. If UB exists in `h_item` **implementation**, there is nothing we can do about it and using `atomics` doesn't help neither.

yuriykoch: Done by 555af6886abed93af319a979a794e42db370268f. I've tried my best to describe all details and caveats of test implementation within the commit message

keryell (approved):
Good commit message!

tests/h_item/h_item_api.cpp
```diff
@@ -13,293 +13,396 @@
...
+  size_t global;
+  size_t physical;
+  size_t logical;
+};
+
+/** @brief Wrapper for global ID support in generic code
+ */
+struct global_id {
+  static const auto& value(const work_item_ids& ids) { return ids.global; }
+  static std::string description() { return "global ID"; };
```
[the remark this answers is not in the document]
yuriykoch: That' a great advice in general, thanks)
Still we do use lazy initialization paradigm in this test, therefore there would be no performance gain from such change. Meanwhile such change would require explicit cast to the `std::string` each time we get the `description`. Therefore I would like to avoid `string_view` usage in this specific case

PR #127, opened 2021-06-09
Adding new test for USM

vasilytric:
Provide verification for functions/APIs from Unified Shared Memory (USM) fill, memcpy and memset

### DIFF
--- a/tests/usm/CMakeLists.txt
+++ b/tests/usm/CMakeLists.txt
@@ -1,0 +1,3 @@
+file(GLOB test_cases_list *.cpp)
+
+add_cts_test(${test_cases_list})
--- a/tests/usm/usm_fill_memset_memcpy.cpp
+++ b/tests/usm/usm_fill_memset_memcpy.cpp
@@ -1,0 +1,85 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Checks functions fill, memset, and memcpy for Unified Shared Memory API.
+//
+*******************************************************************************/
+
+#include "../common/common.h"
+#include <cstring>
+
+#define TEST_NAME usm_fill_memset_memcpy
+
+namespace TEST_NAMESPACE {
+using namespace sycl_cts;
+
+/** Test instance
+ */
+class TEST_NAME : public sycl_cts::util::test_base {
+ public:
+  /** return information about this test
+   */
+  void get_info(test_base::info &out) const override {
+    set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
+  }
+
+  /** execute the test
+   */
+  void run(util::logger &log) override {
+    try {
+      auto q = util::get_cts_object::queue();
+      if (!q.get_device().has(sycl::aspect::usm_shared_allocations)) {
+        log.note(
+            "Device does not support accessing to unified shared memory "
+            "allocation");
+        return;
+      }
+
+      constexpr auto count{2};
+      constexpr auto size{sizeof(int) * count};
+      constexpr auto value_for_filling{1};
+      constexpr auto value_for_first_element_overwriting{10};
+
+      auto deleter = [&q](int *data) { sycl::free(data, q.get_context()); };
+      std::unique_ptr<int, decltype(deleter)> src(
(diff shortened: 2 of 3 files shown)


tests/usm/usm_fill_memset_memcpy.cpp
```diff
@@ -0,0 +1,85 @@
...
+      constexpr auto count{2};
+      constexpr auto size{sizeof(int) * count};
+      constexpr auto value_for_filling{1};
+      constexpr auto value_for_first_element_overwriting{10};
+
+      auto deleter = [&q](int *data) { sycl::free(data, q.get_context()); };
+      std::unique_ptr<int, decltype(deleter)> src(
+          sycl::malloc_shared<int>(size, q.get_device(), q.get_context()),
+          deleter);
+      std::unique_ptr<int, decltype(deleter)> output(
```
keryell, line 48:
```suggestion
      std::unique_ptr src(
          sycl::malloc_shared<int>(size, q.get_device(), q.get_context()),
          deleter);
      std::unique_ptr output(
```
thanks to C++17 CTAD

yuriykoch: @keryell How will it know that's an `unique_ptr<int>` or `unique_ptr<int[]>` though?

vasilytric: @keryell  I've tried to apply your suggestion but got a compilation error: "usm_fill_memset_memcpy.cpp(45,23): error: no viable constructor or deduction guide for deduction of template arguments of 'unique_ptr'
      std::unique_ptr src("

tests/usm/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+file(GLOB test_cases_list *.cpp)
```
ProGTX, line 1: It's generally bad CMake practice to glob for source files, it's better to list them explicitly in order to be able to properly track file dependencies.

tests/usm/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
...
+
+add_cts_test(${test_cases_list})
```
ProGTX, line 3: USM is a new test category, could you please add it to the standard filter as well?

ProGTX (approved):

tests/usm/CMakeLists.txt
vasilytric: This decision implemented across all repository. Fixing this task out of scope this pull request

tests/usm/CMakeLists.txt
vasilytric: i've applied this comment in https://github.com/KhronosGroup/SYCL-CTS/pull/127/commits/83da3e852fb18ead362208f0f9a64915b65c304c

tests/usm/usm_fill_memset_memcpy.cpp
vasilytric: Hi Ronan. I've tried to apply your suggestion but got a compilation error. I have no other ideas for changing this part of code.

tests/usm/CMakeLists.txt
ProGTX: We could start doing it properly just here for now, but I'm happy either way

tests/usm/usm_fill_memset_memcpy.cpp
keryell: Curious. It looks like this is by design according to https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr : "The program is ill-formed if this constructor is selected by class template argument deduction."
This does not work even for simple cases: https://godbolt.org/z/9GeqbTd8h
So, keep it like you did. Thanks for having tried.

vasilytric: @keryell  Do I understand correctly that you approve this PR?

keryell (approved):

tests/usm/usm_fill_memset_memcpy.cpp
keryell: Oh yes! Thank you for reminding this. Sorry for the distraction...

PR #130, opened 2021-06-22
Re-work vector tests for SYCL2020

vasilytric:
Refactor code since now C++17 features can be used and remove cl_* types
from type coverage.


conformance_filters/core.csv
```diff
@@ -41,5 +41,3 @@ vector_load_store
 vector_operators
 vector_swizzle_assignment
 vector_swizzles
-vector_swizzles_opencl
-usm
```
ProGTX, line 45: Why removing usm?

util/type_traits.h
```diff
@@ -53,13 +53,13 @@ using has_atomic_support =
...
  * @brief Checks whether T is a floating-point sycl type
  */
-template <typename T>
-using is_cl_float_type =
-    std::bool_constant<std::is_floating_point<T>::value ||
-                       std::is_same<cl::sycl::half, T>::value ||
-                       std::is_same<cl::sycl::cl_float, T>::value ||
-                       std::is_same<cl::sycl::cl_double, T>::value ||
-                       std::is_same<cl::sycl::cl_half, T>::value>;
+template <typename T> struct is_cl_float_type {
```
ProGTX, line 56: I don't know it from the diffs, but is `is_cl_float_type` still being used?

ProGTX (approved):

util/type_traits.h
vasilytric: Great catch, thank you. Yes, `is_cl_float_type` is still being used in SYCL-CTS code base along with deprecated `cl_*` types. After removing this typed we will still need this type trait for the `sycl::half` type. Probably we will rename this type trait from `is_cl_float_type` to `is_sycl_float_type`.

conformance_filters/core.csv
vasilytric: I've returned "usm" in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/130/commits/16974d619d4543c9b1c0e2793bce2402e78a2d7f

ProGTX (approved):

PR #136, opened 2021-07-13
Update SYCL API from 1.2.1 to 2020 version

kopylovanat:
This patch makes following SYCL API updates across the whole code base:
1. `cl::sycl` namespace is replaced with `sycl` namespace.
2. `access::target` -> `target`
3. `access::mode` -> `access_mode`
4. Removes aliases to standard C++ classes: `*_class` -> `std::*`
5. Updates extension querying for fp16/fp64: `has_extension("cl_khr_fp16/64")` -> `has(sycl::aspect::fp16/fp64)`


keryell (approved):
Quite massive! :-)
Thanks.

bader (approved):

ProGTX (approved):

psalz:
Well, I guess #108 was moot then (and should be reverted now -- in the current state the alias causes a compilation error for hipSYCL!).

bader:
I think we should to discuss the details of SYCL API transition from 1.2.1 to 2020 version with all the implementers. This patch upgrades only part of the 1.2.1 APIs and we need more work in this direction.

psalz:
One thing to consider with find-and-replace changes like this is that it often breaks formatting. As far as I can tell, the following files used to be fully formatted according to `clang-format`, but are no longer after this PR:

```
tests/address_space/address_space.cpp
tests/atomic/atomic_api_32.cpp
tests/atomic/atomic_api_common.h
tests/atomic/atomic_constructors_common.h
tests/buffer/buffer_destructors.cpp
tests/device_event/device_event_api.cpp
tests/device_selector/device_selector_api.cpp
tests/handler/handler_copy.cpp
tests/handler/handler_invoke_api.cpp
tests/h_item/h_item_constructors.cpp
tests/h_item/h_item_equality.cpp
tests/host_task/host_task_invoke_api.cpp
tests/invoke/invoke_kernel_params.cpp
tests/invoke/invoke_kernel_param_sizes.cpp
tests/invoke/invoke_template_kernels.cpp
tests/item/item_2d.cpp
tests/item/item_constructors.cpp
tests/item/item_equality.cpp
tests/kernel/kernel_api.cpp
tests/kernel/kernel_constructors.cpp
tests/kernel/kernel_info.cpp
tests/opencl_interop/opencl_interop_get.cpp
tests/platform/platform_info.cpp
tests/pointers/kernel_pointers.cpp
tests/program/program_api.cpp
tests/program/program_constructors.cpp
util/type_names.h
```

PR #137, opened 2021-07-19
Remove default platform selection

ProGTX:
Hardcoding vendor names makes it difficult to run the CTS on custom devices.
This patch removes default platform selection code altogether.
It's still possible to specify the device type to run on,
but implementations should provide their own ways
of performing selection between multiple platforms.

This includes completely removing the `--platform` argument,
the `ctsplat` enum value, and the `get/set_default_platform` functions.

Changed argument parsing to `parse_known_args`
to allow existing script calls to still work.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,7 +99,7 @@ The SYCL test suite can be launched via the following command::
     $ python runtests.py --help
 
     usage: runtests.py [-h] [-b BINPATH] [--csvpath CSVPATH] [--list]
-                       [-j JUNIT] [-p {host,intel,amd}]
+                       [-j JUNIT]
                        [-d {host,opencl_cpu,opencl_gpu,opencl_accelerator}]
 
     Khronos SYCL CTS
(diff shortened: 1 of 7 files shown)


bader:
> Hardcoding vendor names makes it difficult to run the CTS on custom devices.
> This patch removes default platform selection code altogether.
> It's still possible to specify the device type to run on,
> but implementations should provide their own ways
> of performing selection between multiple platforms.

I'm okay with the change, but doesn't the same logic applies to `device` as well? Shouldn't we remove `--device` option for consistency and let implementation to select tested device?

ProGTX:
I was considering that, but I'm not sure about it for SYCL 1.2.1. We definitely need to improve it, but removing the `--device` flag would get rid of the feature of running both host and OpenCL at the same time, which has implications for the conformance script, hence my hesitation for larger changes. Additionally, `--device` is about controlling the device type, which is clearly defined by the SYCL spec, so there's no vendor hardcoding.

SYCL 2020 is a whole new story altogether that I think is going to be addressed separately.

bader (approved):

bader:
> SYCL 2020 is a whole new story altogether that I think is going to be addressed separately.

Do you plan to propagate this change to the `SYCL-2020` branch?

ProGTX:
> Do you plan to propagate this change to the `SYCL-2020` branch?

It might be worth just applying it directly on the 2020 branch and make a better patch later, or I might try properly changing it there, not sure what the best approach is.

PR #138, opened 2021-07-19
Skip half and double builtin tests if not enabled

ProGTX:
`SYCL_CTS_ENABLE_HALF_TESTS` and `SYCL_CTS_ENABLE_DOUBLE_TESTS`
were applied to most tests, but not to `math_builtin_api`.

### DIFF
--- a/tests/math_builtin_api/CMakeLists.txt
+++ b/tests/math_builtin_api/CMakeLists.txt
@@ -1,9 +1,16 @@
 set(TEST_CASES_LIST "")
 
 set(MATH_CAT_WITH_VARIANT common float relational)
-set(MATH_VARIANT base double half)
+set(MATH_VARIANT base)
 
 set(MATH_CAT integer geometric native half)
+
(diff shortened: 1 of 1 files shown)


tests/math_builtin_api/CMakeLists.txt
```diff
@@ -1,9 +1,17 @@
 set(TEST_CASES_LIST "")
 
 set(MATH_CAT_WITH_VARIANT common float relational)
-set(MATH_VARIANT base double half)
+set(MATH_VARIANT base)
 
-set(MATH_CAT integer geometric native half)
+set(MATH_CAT integer geometric native)
```
kopylovanat, line 6: Just want to point out that half in MATH_CAT is for Half precision math functions (Table 173) and they operate on float. So maybe they shouldn't be excluded even if half tests are not enabled.

ProGTX: Good point, fixed

bader (approved):

PR #148, opened 2021-08-06
Provide debug level for logs

yuriykoch:
Sometimes it's important to have more detailed output from the tests up to the trace of actions made.

Providing such information on the regular basis would pull the performance down. Therefore compile-time switch is required to avoid all message construction routines entirely. Lambda usage makes possible to avoid heavy message construction in case compile-time switch is off. String, string view and string literal instances can be used without overhead.

As a side note, currently tests within `tests/accessors` are using undocumented `#ifdef VERBOSE_LOG` for the purpose of trace output.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,14 @@ endif()
 # ------------------
 
 # ------------------
+# Debug-level verbose logging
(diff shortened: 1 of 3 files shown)


keryell (approved):
Good idea!

ProGTX (approved):

PR #150, opened 2021-08-09
add_device_compiler_definitions [SYCL 2020]

ProGTX:
Bringing https://github.com/KhronosGroup/SYCL-CTS/pull/139 into the SYCL 2020 branch.

I also noticed that `SYCL_CTS_INTEL_PI_CUDA` isn't set like it is in the 1.2.1 branch, so I included that change, I can remove it if needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,10 +52,10 @@ endif()
 option(SYCL_CTS_ENABLE_DOUBLE_TESTS "Enable Double tests." ON)
 option(SYCL_CTS_ENABLE_HALF_TESTS "Enable Half tests." ON)
 if(SYCL_CTS_ENABLE_DOUBLE_TESTS)
-    add_definitions(-DSYCL_CTS_TEST_DOUBLE)
+    add_host_and_device_compiler_definitions(-DSYCL_CTS_TEST_DOUBLE)
 endif()
 if(SYCL_CTS_ENABLE_HALF_TESTS)
-    add_definitions(-DSYCL_CTS_TEST_HALF)
+    add_host_and_device_compiler_definitions(-DSYCL_CTS_TEST_HALF)
 endif()
 # ------------------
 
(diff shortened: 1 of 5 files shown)


bader:
> I also noticed that `SYCL_CTS_INTEL_PI_CUDA` isn't set like it is in the 1.2.1 branch, so I included that change, I can remove it if needed.

IIRC, this macro was set because SYCL-1.2.1 used OpenCL API directly assuming that implementation is based on OpeCL API.
SYCL 2020 doesn't have such assumptions, so I think eventually we should remove all customizations under `SYCL_CTS_INTEL_PI_CUDA` macro, but I don't know if we are ready for that.
@AerialMantis, do you know if this macro is still needed for SYCL-2020 branch?

bader (approved):
LGTM.

keryell:
I was just looking at this this morning comparing the SYCL 1.2.1 CTS branch and the SYCL 2020 branch and realized that there was some divergence.
Would it make sense instead to just merge the SYCL 1.2.1 branch into the SYCL 2020 branch on a regular basis and after review, to benefit from the SYCL 1.2.1 too directly?

keryell (approved):
Approving, even if nobody answered my question.

bader:
> Would it make sense instead to just merge the SYCL 1.2.1 branch into the SYCL 2020 branch on a regular basis and after review, to benefit from the SYCL 1.2.1 too directly?

I think it might be difficult as SYCL 2020 is not compatible with SYCL 1.2.1, so there might be conflicts. Some SYCL 1.2.1 APIs were removed from SYCL 2020 tests and there are more changes like this pending review.
What probably make sense is to have simplified process to cherry-pick changes from SYCL 1.2.1 branch, which are compatible with SYCL 2020 and patch applies cleanly to both branches.

> IIRC, this macro was set because SYCL-1.2.1 used OpenCL API directly assuming that implementation is based on OpeCL API.
> SYCL 2020 doesn't have such assumptions, so I think eventually we should remove all customizations under `SYCL_CTS_INTEL_PI_CUDA` macro, but I don't know if we are ready for that.
> @AerialMantis, do you know if this macro is still needed for SYCL-2020 branch?

I'd like to get an answer to this question as well, but I'm okay to resolve this in follow-up commits. This should not block the merge.

PR #154, opened 2021-08-20
Remove use of half from global namespace

vladimirlaz:
Recent SYCL 1.2.1 and SYCL 2020 specification put half type to sycl
namespace. This change replaces use of the half type from global
namespace with sycl::half.


bader (approved):
Thanks!

ProGTX (approved):

PR #158, opened 2021-08-30
Add tests for atomic access for USM allocations

vasilytric:
In this test implementation we use counter and increase it value on host and device. After increasing we expected counter value that equal `2 * (number of repetitions * value counters incrementing on one increment cycle)`

Due to current tests infrastructure we can't cover the concurrent access from multiple devices with `usm_atomic_shared_allocations` from the same `context`


tests/usm/usm_atomic_access_atomic64.cpp
```diff
@@ -0,0 +1,49 @@
...
+    } catch (const std::exception &e) {
+      auto errorMsg = "a std exception was caught: " + std::string(e.what());
+      FAIL(log, errorMsg);
+    }
+  }
+};
+
+// construction of this proxy will register the above test
+util::test_proxy<TEST_NAME> proxy;
+}  // namespace TEST_NAMESPACE
```
keryell, line 49: Most of these 49 lines should be factorized out

tests/usm/usm_atomic_access.h
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+using with_atomic64 = std::true_type;
+using without_atomic64 = std::false_type;
+/** @brief Dummy functor that use in type coverage
+ *  @tparam CounterT Variable type from type coverage
+ *  @tparam UseAtomic64Flag std::integral_constant type
+ */
+template <typename CounterT, typename UseAtomic64Flag>
```
keryell, line 54:
```suggestion
template <typename CounterT, bool UseAtomic64Flag>
```
looks clearer, with
```
auto constexpr witth_atomic64 = true;
```
and so on?

keryell (changes requested):
The PR description comes with a good design description that deserves to be in a header file comment somewhere

tests/usm/usm_atomic_access.h
```diff
@@ -0,0 +1,173 @@
...
+struct run_all_tests {
+  void operator()(sycl::queue &queue, sycl_cts::util::logger &log,
+                  const std::string &typeName) {
+    if constexpr ((sizeof(CounterT) * CHAR_BIT) == 64) {
+      if (!UseAtomic64Flag::value) {
```
ProGTX, line 59:
```suggestion
      if constexpr (!UseAtomic64Flag::value) {
```

If I understand this correctly

ProGTX (approved):

tests/usm/usm_atomic_access_atomic64.cpp
vasilytric: @keryell Ronan, You tell about uniting two test files (*.cpp) to one?

from this:
```
for_all_types<usm_atomic_access::run_all_tests,
                      usm_atomic_access::with_atomic64>(
      usm_atomic_access::get_scalar_types(), queue, log);
```
to this:
```
for_all_types<usm_atomic_access::run_all_tests,
                    usm_atomic_access::with_atomic64>(
          usm_atomic_access::get_scalar_types(), queue, log);
for_all_types<usm_atomic_access::run_all_tests,
                    usm_atomic_access::without_atomic64>(
          usm_atomic_access::get_scalar_types(), queue, log);
```

Or I wrong about Your suggestion?
Also `try-catch` block did not remove due to now [PR with moving try-catch block to test base](https://github.com/KhronosGroup/SYCL-CTS/pull/167) open yet

tests/usm/usm_atomic_access.h
vasilytric: `bool` type cannot be provided to the `for_all_types` function. by this function I call `run_all_tests` in 
[usm_atomic_access_core.cpp](https://github.com/KhronosGroup/SYCL-CTS/pull/158/files/7d316d63dd0cdd9b111c7f41a0386db167502b02#diff-29697b92aebee3800e595593638621294b090a5f860c125ba20694b76373188a) and [usm_atomic_access_atomic64.cpp](https://github.com/KhronosGroup/SYCL-CTS/pull/158/files/7d316d63dd0cdd9b111c7f41a0386db167502b02#diff-dc3e9b7188bf1db340a2b583845b1da7e8bdb2266fa0eb6c67e6258a139a75be)

tests/usm/usm_atomic_access.h
vasilytric: I've applied this comment in this commit: 4e8f956ce1f4490d5d5f2379f0f26d7342c8376c

tests/usm/usm_atomic_access.h
```diff
@@ -56,15 +56,15 @@ struct run_all_tests {
...
+      if constexpr (!UseAtomic64Flag::value) {
         return;
       }
       if (!queue.get_device().has(sycl::aspect::atomic64)) {
         log.note("Device does not perform 64-bit atomic operations");
         return;
       }
     } else {
-      if (UseAtomic64Flag::value) {
+      if constexpr (UseAtomic64Flag::value) {
```
keryell, line 67: What is the goal of this test?
OK, this is for the case we have a counter which has not 64 bits but we still have `UseAtomic64Flag` we cannot use here...
Should we have a logging message here too or just ignoring is good?

tests/usm/usm_atomic_access.h
keryell: I guess you would have to propagate the API change in the usage and the callers too.
Or is there another issue I do not see?

tests/usm/usm_atomic_access_atomic64.cpp
keryell: Do not worry: solved by https://github.com/KhronosGroup/SYCL-CTS/pull/167 :-)

tests/usm/usm_atomic_access.h
vasilytric: `for_all_types` used across all project and this changes requir do some changes in many files and then verification that changed tests works correctly
I suggest did not do this changes and just provide this argument in common parameters, as implemented in this commit: a73c49c988826254f402716a6e909aef9c041ad5

tests/usm/usm_atomic_access.h
vasilytric: I think that need logging in case, when we skip test due to some reasons, if all happened as expected log messages is useless

tests/usm/usm_atomic_access.h
```diff
@@ -0,0 +1,173 @@
...
+      if (!queue.get_device().has(sycl::aspect::atomic64)) {
+        log.note("Device does not perform 64-bit atomic operations");
+        return;
+      }
+    } else {
+      if (use_atomic_flag) {
+        return;
+      }
+    }
+    run_test_with_chosen_mem_type<CounterT>(queue, log, typeName);
```
keryell, line 71: Actually I was thinking about keeping the `constexpr` approach you had in your first version but just in a more modern way
```suggestion
template <typename CounterT, bool use_atomic_flag>
struct run_all_tests {
  void operator()(sycl::queue &queue, sycl_cts::util::logger &log,
                  const std::string &typeName) {
    if constexpr ((sizeof(CounterT) * CHAR_BIT) == 64) {
      if constexpr (!use_atomic_flag) {
        return;
      }
      if (!queue.get_device().has(sycl::aspect::atomic64)) {
        log.note("Device does not perform 64-bit atomic operations");
        return;
      }
    } else {
      if constexpr (use_atomic_flag) {
        return;
      }
    }
    run_test_with_chosen_mem_type<CounterT>(queue, log, typeName);
```

vasilytric: I can't provide `use_atomic_flag` as template parameter due to `for_all_types` (this function calls `run_all_tests`) does not support template parameters, that not equal to `typename` type

I suggest to not change `for_all_types` due to this function used across all project and this changes require do changes in many files and then verification that changed tests works correctly

keryell: OK. As you wish.

vasilytric: @keryell Thank You for understanding. Do You have any comments for this PR?

keryell (approved):
Thanks!

PR #160, opened 2021-09-01
Remove deprecated sycl::exception related APIs

kopylovanat:
remove tests for error_types
    enum class errc is used in SYCL 2020 instead of error types
    tests for errc have been developed separately

change get_cl_code to sycl::exception.code()


tests/exception_handling/asynchronous_exceptions.cpp
```diff
@@ -21,15 +21,27 @@ void check_exception_list_types() {
   static_assert(std::is_same<value_type, std::exception_ptr>::value,
                 "exception_list::value_type is of wrong type");
 
-  { check_type_existence<sycl::exception_list::reference> typeCheck; }
-  { check_type_existence<sycl::exception_list::const_reference> typeCheck; }
+  {
+    check_type_existence<sycl::exception_list::reference> typeCheck;
+    (void)typeCheck;
```
keryell, line 26: Just curious: do we need these `(void)` to silent something here?
Executing the constructor and destructor is not enough?

tests/exception_handling/asynchronous_exceptions.cpp
```diff
@@ -83,8 +95,10 @@ class TEST_NAME : public util::test_base {
         std::string sc = e.what();
         if (e.has_context()) {
           sycl::context c = e.get_context();
+          (void)c;
```
keryell, line 98: These ones seem more useful to me.

tests/exception_handling/asynchronous_exceptions.cpp
kopylovanat: Yes, it's here to silent warnings for unused variable.
Do you think it's better to add comment clarifying it?

tests/error/error_types.cpp
```diff
@@ -1,61 +0,0 @@
...
-
-  /** execute this test
-   */
-  void run(util::logger &log) override {
-#define CHECK_EXISTS(EXCEPTION_NAME)                       \
-  if (!std::is_class<EXCEPTION_NAME>::value) {             \
-    FAIL(log, "EXCEPTION_NAME is not defined as a class"); \
-  }
-
-    CHECK_EXISTS(sycl::exception);
```
ProGTX, line 34: I would expect these checks to be replaced with checks for different error codes existing, is that already somewhere or is it planned for a later PR?

tests/error/error_types.cpp
```diff
@@ -1,61 +0,0 @@
...
-    CHECK_EXISTS(sycl::link_program_error);
-    CHECK_EXISTS(sycl::invalid_object_error);
-    CHECK_EXISTS(sycl::memory_allocation_error);
-    CHECK_EXISTS(sycl::platform_error);
-    CHECK_EXISTS(sycl::profiling_error);
-    CHECK_EXISTS(sycl::feature_not_supported);
-#undef CHECK_EXISTS
-
-    /* Check that exception_list exists */
-    if (!std::is_class<sycl::exception_list>::value) {
```
ProGTX, line 52: `exception_list` has not been deprecated, shouldn't we keep this?

tests/error/error_types.cpp
```diff
@@ -1,61 +0,0 @@
...
-      FAIL(log, "Exception list is not defined as a class");
-    }
-  }
```
ProGTX, line 55: Another check that could be done in this test is that `sycl::exception` inherits from `std::exception`

tests/error/error_types.cpp
kopylovanat: Yes, tests for `enum class errc` with error codes are going to be added in later PR on this week.

tests/error/error_types.cpp
kopylovanat: You are right, returned that check.

tests/exception_handling/asynchronous_exceptions.cpp
keryell: OK.
Then perhaps this API is too low level and we could have instead just a `constexpr` function
```c++
check_type_existence<sycl::exception_list::reference>();
```

kopylovanat: Thank you. Done in https://github.com/KhronosGroup/SYCL-CTS/pull/160/commits/3c163d7f918c9fa995a191ee52101e1a2aceb4ff

tests/common/common.h
```diff
@@ -511,7 +511,7 @@ struct image_access<3> {
  */
 template <typename T>
 struct check_type_existence {
-  check_type_existence() = default;
+  constexpr check_type_existence() = default;
 };
```
keryell, line 515: I was thinking to just a function instead of a class:
```c++
template <typename T>
void constexpr check_type_existence() {
};
```

keryell (commented):
Almost there.

tests/common/common.h
kopylovanat: Done in https://github.com/KhronosGroup/SYCL-CTS/pull/160/commits/cb1eff92dd00b931d5bc162b23124da7b4bb838a

keryell (approved):
Cool!

bader (approved):

tests/error/error_types.cpp
ProGTX: I'm happy for this to be resolved as part of a future PR as mentioned in https://github.com/KhronosGroup/SYCL-CTS/pull/160#discussion_r703458833

ProGTX (approved):

PR #162, opened 2021-09-01
Provide adapter for type usage in kernel names

yuriykoch:
There is a plenty of restrictions on types for kernel names.
Still we may need to cover some types that should not be used for kernel names and kernel names' template parameters. Such type
coverage still require some generic code, so to avoid passing specific kernel names all the way through generic code we may just use some adapter to construct safe kernel name right in place.

### DIFF
--- a/util/kernel_names.h
+++ b/util/kernel_names.h
@@ -1,0 +1,139 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provide adapter for safely using different types with kernel names in
+//  generic code
+//
+*******************************************************************************/
+
+#ifndef __SYCLCTS_UTIL_KERNEL_NAMES_H
+#define __SYCLCTS_UTIL_KERNEL_NAMES_H
+
+#include <string>
+#include <utility>
+
+namespace sycl_cts {
+namespace util {
+namespace kernel_name {
+
+namespace details {
+
+template <typename T>
+struct adapter_core {};
+
+struct adapter_int8_t {};
+struct adapter_int16_t {};
+struct adapter_int32_t {};
+struct adapter_int64_t {};
+struct adapter_uint8_t {};
+struct adapter_uint16_t {};
+struct adapter_uint32_t {};
+struct adapter_uint64_t {};
+
+struct adapter_byte {};
+struct adapter_size_t {};
+
+/** @brief Provides type that is safe to use in kernel names based on type given
+ *  @details There is a plenty of restrictions on types we may use for kernel
+ *           naming. The most obvious is non-forward-declarable classes, but
+ *           there are more of them.
+ *           This function uses discarded statements to map input type into the
+ *           return type we may use for the kernel name template parameters.
+ *           Note: it does not provide different kernel names for the type
+ *           aliases, so we may have the same kernel name
+ *             - for `int` and `int32_t`, or
+ *             - for `size_t` and `uint64_t`, or
+ *             - for any other possible type alias
+ *           on any specific SYCL implementation. We may say it provides adapter
+ *           aliases for the type aliases.
+ */
+template <typename T>
+constexpr auto adapter_helper() {
+  // Fixed-width types are optional and implementation-defined
+#ifdef INT8_MAX
+  if constexpr (std::is_same_v<T, std::int8_t>) {
+    return adapter_int8_t{};
+  } else
+#endif
+#ifdef INT16_MAX
+      if constexpr (std::is_same_v<T, std::int16_t>) {
+    return adapter_int16_t{};
+  } else
+#endif
+#ifdef INT32_MAX
+      if constexpr (std::is_same_v<T, std::int32_t>) {
+    return adapter_int32_t{};
+  } else
+#endif
+#ifdef INT64_MAX
+      if constexpr (std::is_same_v<T, std::int64_t>) {
+    return adapter_int64_t{};
+  } else
+#endif
+#ifdef UINT8_MAX
+      if constexpr (std::is_same_v<T, std::uint8_t>) {
+    return adapter_uint8_t{};
+  } else
+#endif
+#ifdef UINT16_MAX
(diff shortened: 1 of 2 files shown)


util/kernel_names.h
```diff
@@ -0,0 +1,139 @@
...
+ *           aliases, so we may have the same kernel name
+ *             - for `int` and `int32_t`, or
+ *             - for `size_t` and `uint64_t`, or
+ *             - for any other possible type alias
+ *           on any specific SYCL implementation. We may say it provides adapter
+ *           aliases for the type aliases.
+ */
+template <typename T>
+constexpr auto adapter_helper() {
+  // Fixed-width types are optional and implementation-defined
```
keryell, line 53: That is true for general C/C++ but what will happens then elsewhere in the CTS, where such types might be used without any guard?
And in the SYCL 2020 specification *4.14. Data types* & *5.5. Built-in scalar data types* it looks like we assume that they do exist, or at least the 8-bit `char` does exist, so `std::int8_t`...

util/kernel_names.h
```diff
@@ -0,0 +1,139 @@
...
+ *           we are unable to have
+ *               using T = outer::inner;
+ *               single_task<kernel_name<T>>,
+ *           but we are able to have
+ *               using T = typename kernel_name::adapter<outer::inner>::type
+ *               single_task<kernel_name<T>>
+ *           in case we provide a proper template specialization for adapter.
+ *
+ *           Test developer is free to provide template specialization for any
+ *           test-specific type.
```
keryell, line 123: This is an interesting design document that deserves to appear close to the top of this file.

util/kernel_names.h
```diff
@@ -0,0 +1,139 @@
...
+ *           for kernel names directly. For example, for any type we cannot
+ *           forward-declare:
+ *               struct outer {
+ *                 struct inner {};
+ *               };
+ *           we are unable to have
+ *               using T = outer::inner;
+ *               single_task<kernel_name<T>>,
+ *           but we are able to have
+ *               using T = typename kernel_name::adapter<outer::inner>::type
```
keryell, line 118: Interesting.
I am curious whether this works on some pure integration-header based implementation like ComputeCpp, because I have the feeling that you need to have `outer::inner` somehow declared in the integration header to be referenced as a template name.
@ProGTX?
@bader I guess in Intel implementation all the useful code might be in the integration header/footer and thus `outer::inner` is found? But in that case, there might not be this constraint on forward-declarability?

ProGTX: I'm not sure. We're not able to test this at the moment, but from what I understand the adapter is supposed to work around this limitation?

keryell: It is unclear whether this should not be forbidden in the specification too. @gmlueck?

gmlueck: Yes, I agree that we should clarify the spec about legal kernel names.  I think this CTS code is assuming that the tool which generates the integration header resolves all type aliases and writes the resolved type to the integration header.  To illustrate with a simple example:

```
int main() {
  sycl::queue q;
  using myname = int;

  q.single_task<myname>([=] {/*...*/});
};
```

I think we expect this to work because the integration header should resolve the `myname` alias to `int` and use that in the integration header:

```
template <> struct KernelInfo<int> {/*...*/};
```

Of course, this wouldn't work if the integration header referenced the `myname` alias directly without resolving the type:

```
template <> struct KernelInfo<myname> {/*...*/};  // Doesn't work because "myname" is declared in function scope
```

I think that this CTS code is the same concept, except that the type alias is more complicated.

If others agree with my interpretation, perhaps we need to clarify in the spec that type aliases are allowed as kernel names, and the restrictions we list apply only to the resolved type (not to the alias itself).

yuriykoch: Just to make sure - the original intent is to have something like this:

```
struct outer {
  struct inner {
  };
};

template <typename T>
struct name {};

int main() {
  sycl::queue q;
  using Foo = typename kernel_name::adapter_t<outer::inner>;
  using Boo = name<kernel_name::adapter_t<std::byte>>;
  q.single_task<Foo>([=] {/*...*/});
  q.single_task<Boo>([=] {/*...*/});
};
```

util/kernel_names.h
yuriykoch: I was thinking about it in terms of doxygen documentation section for `struct adapter`. So every time we modify it or every time we end up near the `adapter_t` declaration using IDE or full-text search we would see the appropriate documentation nearby.
AFAIK I cannot move a `struct adapter` up because of C++ restrictions. @keryell Should I duplicate design notice in the header comment or move it entirely?

util/kernel_names.h
gmlueck: There is an unresolved issue against the SYCL 2020 spec that relates to this topic:

https://gitlab.khronos.org/sycl/Specification/-/issues/544

That issue basically says that the SYCL spec has several APIs that use the fixed-width integral types (e.g. `std::uin32_t`), but the core C++ specs says these types are optional.  We need to decide if the SYCL spec requires the underlying C++ compiler to define these types, or we need to change the SYCL spec to stop using them.

keryell: C++20 adopts 2-complement integer representation, so I guess future version of SYCL will follow this too.

What to do if there is no `std::int16_t`, etc? But alternatively how to know the CTS works if there is no SYCL implementation for such a device without a `std::int16_t`, etc? How to test the CTS implementation itself?

util/kernel_names.h
keryell: What is the `struct name` used for?

keryell (approved):
OK, I have the feeling that the problem you are trying to solve is not the one I was thinking of.
If it works, let's go for it.

util/kernel_names.h
yuriykoch: Done by 28e77dc09abc162c641cf412f084239d32410c0f

ProGTX (approved):

PR #163, opened 2021-09-01
(Deprecated API) Remove has_extension() and replace it with .has() method

kovshikovser:
- All has_extension(string) calls replaced with has(aspect)
- Removed has_extension(string) calls with extension without corresponding sycl::aspect (according to Table 189. "SYCL support for OpenCL 1.2 extensions" from the SYCL 2020 Spec)

### DIFF
--- a/tests/atomic/atomic_api_64_base.cpp
+++ b/tests/atomic/atomic_api_64_base.cpp
@@ -105,7 +105,7 @@ class TEST_NAME : public util::test_base {
 
       /** Check atomics for supported types
        */
-      if (testDevice.has_extension("cl_khr_int64_base_atomics")) {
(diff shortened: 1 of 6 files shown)


keryell (approved):

AlexeySachkov:
I wonder if there are any plans to have some separate test category for deprecated APIs. Deprecation doesn't mean removal and it still means that the functionality must be supported. What is CTS development strategy in that regard?

ProGTX (approved):

bader:
> I wonder if there are any plans to have some separate test category for deprecated APIs. Deprecation doesn't mean removal and it still means that the functionality must be supported. What is CTS development strategy in that regard?

Good question. I've created https://github.com/KhronosGroup/SYCL-CTS/issues/172 to discuss it separately.

PR #166, opened 2021-09-02
Update opencl_interop tests for SYCL 2020 style

kopylovanat:
Added check if SYCL_BACKEND_OPENCL is defined
Replaced method get() with function get_native()
Replaced removed constructors with make_*()
Removed checks for program's get_compile_options and get_link_options in opencl_interop_kernel.cpp as program is removed and there are no clear counterparts in kernel_bundle for it.


bader (approved):

ProGTX (approved):

keryell (approved):
That looks good!

PR #167, opened 2021-09-02
Move try catch to test base

vasilytric:
Moving `try-catch` common blocks from test to tests base
I did not move this blocks from `opencl` tests or non common blocks

This PR contain many changes due to code was reformatted after removing `try-catch` blocks. 
Ronan Keryell offered to do it in [this](https://github.com/KhronosGroup/SYCL-CTS/pull/140#discussion_r681945924) comment


ProGTX (approved):

util/test_base.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+void test_base::run_test(class logger &log) {
+  try {
+    this->run(log);
+  } catch (const sycl::exception &e) {
+    log_exception(log, e);
+    auto errorMsg = "a SYCL exception was caught: " + std::string(e.what());
+    FAIL(log, errorMsg);
+  } catch (const std::exception &e) {
+    auto errorMsg = "an exception was caught: " + std::string(e.what());
```
keryell, line 26: As I suggested in another PR, what is important is that an exception was thrown.
Of course to know it we need to catch the exception, but this is an implementation detail.
```suggestion
    auto errorMsg = "a SYCL exception was thrown: " + std::string(e.what());
    FAIL(log, errorMsg);
  } catch (const std::exception &e) {
    auto errorMsg = "an exception was thrown: " + std::string(e.what());
```

util/test_base.h
```diff
@@ -41,16 +43,18 @@ class test_base {
...
    *  @param log for emitting test notes and results
    */
-  virtual bool setup(class logger &) {
-    // stub
-    return true;
-  }
+  virtual bool setup(class logger &) { return true; }
 
-  /** execute this test
+  /** method that will be overriden in test file
```
keryell, line 48:
```suggestion
  /** member function that will be overridden in test file
```

util/test_base.h
```diff
@@ -41,16 +43,18 @@ class test_base {
...
    *  @param log for emitting test notes and results
    */
   virtual void run(class logger &log) = 0;
 
+  /** run overriden method with try-catch block
```
keryell, line 53:
```suggestion
  /** run overridden member function with try-catch block
```

keryell (changes requested):
This is quite an achievement! \o/
You have removed +8,330 −9,655 = 1325 lines of code.
Go on like that! :-)
Just a few remarks at the end.

psalz:
While I agree that tests absolutely shouldn't have to be manually wrapped in `try-catch` blocks, I'm not sure whether it is worth retroactively changing this for all of the existing tests, obliterating `git blame` for most of the CTS. Couldn't we just wrap ` test_base::run` in a `try-catch` internally, so that new tests don't have to do it manually, while existing tests simply continue to catch the exceptions themselves?

fraggamuffin:
Sept 9:
* ok to merge when it is ready

ProGTX:
> 
> 
> While I agree that tests absolutely shouldn't have to be manually wrapped in `try-catch` blocks, I'm not sure whether it is worth retroactively changing this for all of the existing tests, obliterating `git blame` for most of the CTS. Couldn't we just wrap ` test_base::run` in a `try-catch` internally, so that new tests don't have to do it manually, while existing tests simply continue to catch the exceptions themselves?

As mentioned, if we remove the `try-catch` blocks while still keeping the original scope (curly braces) of the `try` block, none of that gets changed, so `git blame` still shows the proper history.

psalz:
> As mentioned, if we remove the `try-catch` blocks while still keeping the original scope (curly braces) of the `try` block, none of that gets changed, so `git blame` still shows the proper history.

Ah, I misunderstood you then. That's actually a neat idea!

keryell:
> As mentioned, if we remove the `try-catch` blocks while still keeping the original scope (curly braces) of the `try` block, none of that gets changed, so `git blame` still shows the proper history.

Otherwise we can just postpone the `clang-format` phase up to the future, so with a local indent of 4 instead of 2.

psalz:
> > As mentioned, if we remove the `try-catch` blocks while still keeping the original scope (curly braces) of the `try` block, none of that gets changed, so `git blame` still shows the proper history.
> 
> Otherwise we can just postpone the `clang-format` phase up to the future, so with a local indent of 4 instead of 2.

This is what I originally though @ProGTX was proposing. It would suddenly introduce a lot more unformatted files into the codebase, which is something I think we should always try to avoid.

util/test_base.cpp
keryell: This was done in https://github.com/KhronosGroup/SYCL-CTS/pull/143 which is now merged in. So there is just some homogenizing work here.

keryell (commented):
When it was discussed during a recent committee meeting, the consensus was to merge this once the merge conflicts are solved.
Thanks for the cleaning!

bader:
@vasilytric, could you resolve merge conflicts, please?

bader:
@psalz, FYI. CI in works.

vasilytric:
> @vasilytric, could you resolve merge conflicts, please?

@bader I've done it in this merge-commit: 8c87087f537260069b018b82c34989b07350e351 
but got unexpected error, looks like this error is not related to code from this PR, but related to different library (`hipsycl`)

bader:
> > @vasilytric, could you resolve merge conflicts, please?
> 
> @bader I've done it in this merge-commit: [8c87087](https://github.com/KhronosGroup/SYCL-CTS/commit/8c87087f537260069b018b82c34989b07350e351)
> but got unexpected error, looks like this error is not related to code from this PR, but related to different library (`hipsycl`)

Great, it means the notification works as expected. :)
Please, ignore CTS build failures with hipsycl implementation. @psalz is planning to address them in a separate pull request.

vasilytric:
> Great, it means the notification system works as expected. :)

Thanks for the information :)

util/test_base.h
vasilytric: I've applied this comment in this commit: 4a4aa32e58ee565190fd5c4ff444133660b1ea7b

util/test_base.h
vasilytric: I've applied this comment in this two commits: 4a4aa32e58ee565190fd5c4ff444133660b1ea7b and 54f529b51c7864445834a0e62af510481bf755cf

keryell (approved):
Great!

tests/invoke/invoke_kernel_params.cpp
```diff
@@ -72,8 +72,12 @@ class test_kernel_name;
 class TEST_NAME : public util::test_base {
  public:
   typedef sycl::accessor<uint32_t, 1, sycl::access_mode::write,
-                             sycl::target::device>
-      acc_uint32_t;
+<<<<<<< HEAD
```
psalz, line 75: Merge conflict marker

tests/invoke/invoke_kernel_params.cpp
```diff
@@ -148,10 +151,12 @@ class TEST_NAME : public util::test_base {
 
           // access the output
           auto acc_pass =
-              buf_result
-                  .template get_access<sycl::access_mode::write,
-                                       sycl::target::device>(
-                      cgh);
+<<<<<<< HEAD
```
psalz, line 154: Merge conflict marker

psalz (changes requested):
Case in point that this patch may be a bit too large to review: Currently it contains two merge conflict markers.

tests/invoke/invoke_kernel_params.cpp
vasilytric: Thanks for the pay attention, I've removed this in this commit: f6cd54a48b76cbfb3cd2ad739c05ed42d43dcd3c

tests/invoke/invoke_kernel_params.cpp
vasilytric: Thanks for the pay attention, I've removed this in this commit: f6cd54a48b76cbfb3cd2ad739c05ed42d43dcd3c

bader:
@vasilytric, please, resolve merge conflicts with recent commits to the target branch.

psalz:
I'd like to again entertain @ProGTX's idea of keeping the current scope to avoid excessive reformatting. This line gets us there 99% of the way: `find tests -type f | xargs gawk -i inplace '/try {/ { print "{"; next } /} catch \(/ { skip = 1; print "}"; next }; skip == 1 && /}$/ { skip = 0; next } skip == 1 { next } 1'; git diff -U0 --no-color | clang-format-diff-10 -p1 -i`.

I'd be willing to file a PR containing this change plus some cherry-picked patches by @vasilytric such as 531159c.

vasilytric:
> @vasilytric, please, resolve merge conflicts with recent commits to the target branch.

@bader merge conflicts was resolved in this commit: 377e4cb84efb2b8bf573ff82d19a90e90d92c6eb

bader:
Superseded by #190.

PR #170, opened 2021-09-06
Fix reference values for sampled access with linear filtration

yuriykoch:
Upper coordinate in case of floating-point CoordT should use different texels for linear filtration.
See ["Filter Mode CLK_FILTER_LINEAR"](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#addressing-and-filter-modes) in OpenCL C Specification we are using according to the SYCL 1.2.1 par.3.

### DIFF
--- a/tests/accessor/accessor_api_image_common.h
+++ b/tests/accessor/accessor_api_image_common.h
@@ -945,7 +945,7 @@ class image_accessor_api_sampled_r {
       return get_expected_value_nearest(idx);
     }
     constexpr bool useUpper =
-        std::is_same<coordTag, acc_coord_tag::use_normalized_upper>::value;
+        std::is_same_v<coordTag, acc_coord_tag::use_normalized_upper>;
     if constexpr (!useUpper) {
       // Use simplified equation for lower coordinate values
       return get_expected_value_linear(idx);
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

tests/accessor/accessor_api_image_common.h
```diff
@@ -945,14 +936,56 @@ class image_accessor_api_sampled_r {
...
   T get_expected_value(image_id_t<dim, target> idx) const {
     const bool useLinear =
         m_sampler.filtering_mode == sycl::filtering_mode::linear;
 
     if (!useLinear) {
       return get_expected_value_nearest(idx);
     }
-    return get_expected_value_linear(idx);
+    constexpr bool useUpper =
+        std::is_same<coordTag, acc_coord_tag::use_normalized_upper>::value;
```
keryell, line 948: Oh no! You have started to use `std::is_same_v` everywhere instead. Go on! :-)

keryell (changes requested):

tests/accessor/accessor_api_image_common.h
yuriykoch: Fixed by 69a4cf9461f7b71c16a475d7335cb88c86869b63 :-)

keryell (approved):
Good!

PR #171, opened 2021-09-06
Remove deprecated checks for info::device

kopylovanat:
for max_constant_buffer_size, max_constant_args


ProGTX (approved):

psalz:
As mentioned in #172, I don't think we should be actively removing tests for features that are still required for 2020 spec conformance.

keryell (approved):

bader:
@kopylovanat, this API is not removed from the spec, so it's should be tested by SYCL CTS. I suggest closing this PR w/o merging.

kopylovanat:
@bader Done.

PR #173, opened 2021-09-08
Add test for sycl::make_error_code

vasilytric:
Provide verification for sycl::make_error code and verify same work std::error_code and sycl::make_error_code

### DIFF
--- a/tests/exceptions/exceptions.h
+++ b/tests/exceptions/exceptions.h
@@ -1,0 +1,36 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Common code for exceptions tests
+//
+*******************************************************************************/
+
+#ifndef __SYCL_CTS_TEST_EXCEPTIONS_H
+#define __SYCL_CTS_TEST_EXCEPTIONS_H
+
+#include "../common/common.h"
+#include <vector>
+
+namespace exceptions {
+
+static const std::vector<sycl::errc> all_err_codes{
+    sycl::errc::success,
+    sycl::errc::runtime,
+    sycl::errc::kernel,
+    sycl::errc::accessor,
+    sycl::errc::nd_range,
+    sycl::errc::event,
+    sycl::errc::kernel_argument,
+    sycl::errc::build,
+    sycl::errc::invalid,
+    sycl::errc::memory_allocation,
+    sycl::errc::platform,
+    sycl::errc::profiling,
+    sycl::errc::feature_not_supported,
+    sycl::errc::kernel_not_supported,
+    sycl::errc::backend_mismatch};
+
+}  // namespace exceptions
+
+#endif  // __SYCL_CTS_TEST_EXCEPTIONS_H
--- a/tests/exceptions/exceptions.h
+++ b/tests/exceptions/exceptions.h
@@ -12,6 +12,8 @@
 #include "../common/common.h"
 #include <vector>
 
+/** @brief Provide common code for all tests for exceptions
+ */
 namespace exceptions {
 
 static const std::vector<sycl::errc> all_err_codes{
--- a/tests/exceptions/exceptions.h
+++ b/tests/exceptions/exceptions.h
@@ -14,9 +14,9 @@
 
 /** @brief Provide common code for all tests for exceptions
  */
-namespace exceptions {
+namespace {
 
-static const std::vector<sycl::errc> all_err_codes{
+inline const std::vector<sycl::errc> all_err_codes{
     sycl::errc::success,
     sycl::errc::runtime,
     sycl::errc::kernel,
@@ -33,6 +33,6 @@ static const std::vector<sycl::errc> all_err_codes{
     sycl::errc::kernel_not_supported,
     sycl::errc::backend_mismatch};
 
-}  // namespace exceptions
+}  // namespace
 
 #endif  // __SYCL_CTS_TEST_EXCEPTIONS_H
--- a/tests/exceptions/exceptions_make_error_code.cpp
+++ b/tests/exceptions/exceptions_make_error_code.cpp
@@ -21,6 +21,7 @@ using namespace sycl_cts;
 void check_sycl_working(sycl::errc err_c, util::logger &log) {
   std::error_code make_err_c_result{sycl::make_error_code(err_c)};
 
+  CHECK_TYPE(log, make_err_c_result, std::error_code());
   if (!noexcept(sycl::make_error_code(err_c))) {
     FAIL(log, "sycl::make_error_code function are not marked as \"noexcept\"");
   }
--- a/tests/exceptions/exceptions_make_error_code.cpp
+++ b/tests/exceptions/exceptions_make_error_code.cpp
@@ -1,0 +1,118 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for sycl::make_error_code function
+//
+*******************************************************************************/
+
+#include "exceptions.h"
+
+#define TEST_NAME exceptions_make_error_code
+
+namespace TEST_NAMESPACE {
+
+using namespace sycl_cts;
+
+/** @brief Provide verification for sycl::make_error_code function vorking
+ *  @param err_c Error code from sycl::errc enumeration
+ *  @param log sycl_cts::util::logger class object
+ */
+void check_sycl_working(sycl::errc err_c, util::logger &log) {
+  auto make_err_c_result{sycl::make_error_code(err_c)};
+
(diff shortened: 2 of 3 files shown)


tests/exceptions/exceptions.h
```diff
@@ -0,0 +1,36 @@
...
+//
+*******************************************************************************/
+
+#ifndef __SYCL_CTS_TEST_EXCEPTIONS_H
+#define __SYCL_CTS_TEST_EXCEPTIONS_H
+
+#include "../common/common.h"
+#include <vector>
+
+namespace exceptions {
```
bader, line 15: Could you clarify the purpose of `namespace exceptions`, please?

tests/exceptions/exceptions_make_error_code.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+#include "exceptions.h"
+
+#define TEST_NAME exceptions_make_error_code
+
+namespace TEST_NAMESPACE {
+
+using namespace sycl_cts;
+
+/** @brief Provide verification for sycl::make_error_code function vorking
```
bader, line 17:
```suggestion
/** @brief Provide verification for sycl::make_error_code function
```

tests/exceptions/exceptions_make_error_code.cpp
```diff
@@ -0,0 +1,118 @@
...
+ *  @param err_c Error code from sycl::errc enumeration
+ *  @param log sycl_cts::util::logger class object
+ */
+void check_sycl_working(sycl::errc err_c, util::logger &log) {
+  auto make_err_c_result{sycl::make_error_code(err_c)};
```
bader, line 22: Is it a good use for `auto`? I think the test must check the return type for the tested SYCL function.
```suggestion
  std::error_code make_err_c_result{sycl::make_error_code(err_c)};
```

tests/exceptions/exceptions_make_error_code.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  if (!noexcept(sycl::make_error_code(err_c))) {
+    FAIL(log, "sycl::make_error_code function are not marked as \"noexcept\"");
+  }
+  CHECK_TYPE(log, make_err_c_result, std::error_code());
```
bader, line 27: Ah... this is where we check the return type...

bader (changes requested):

tests/exceptions/exceptions.h
vasilytric: I've add clarifications for this `namespace` in this commit: 169e0b181a4e405853c0139fe67fc1fed264e596

tests/exceptions/exceptions_make_error_code.cpp
vasilytric: Alexey, can i skip this comment due to that this check is performed below?

tests/exceptions/exceptions.h
bader: The comment doesn't explain why this namespace is needed. Can we remove it?

tests/exceptions/exceptions_make_error_code.cpp
bader: I'm okay if you move the type check right after the variable definition. Otherwise it's hard to follow the logic of the test.

tests/exceptions/exceptions.h
vasilytric: We using this namespace to avoid some redefinition errors or situations where we have variable/function with the same name

yuriykoch: @bader Let's assume we doesn't provide any namespace for `tests/foo/foo.h` and `tests/boo/boo.h`. Let's then assume we have same variable/function name in both headers. Wouldn't it break ODR-rule for a `test_all` executable, sometimes without even compilation failure?

bader: The variable is declared as `static`, so it doesn't break ODR-rule.
As this header is included to a single test, I suggest we move the `all_err_codes` definition to tests/exceptions/exceptions_make_error_code.cpp and do not add tests/exceptions/exceptions.h file.

tests/exceptions/exceptions_make_error_code.cpp
vasilytric: I've remove using `CHECK_TYPE` macros and use determinated type for `make_err_c_result` variable
I've removed `CHECK_TYPE` macros due to duplicated verification unnecessary
I've do this in this commit: 958080e51bc6720db7a68642eee3f0038a1947d4

tests/exceptions/exceptions.h
vasilytric: We are going to use it in tests that we will provide soon. Should we remove header now and add it again later?

vasilytric:
> Technically, `CHECK_TYPE` is more strict check than just initialize `std::error_code` variable. Current test implementation allows `make_error_code` to return any type `std::error_code` can be initialized with. See `CHECK_TYPE` definition:
> https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/common/macros.h#L123-L138

Ok, I revert my changes: I will verify type using `CHECK_TYPE` and will don't use `auto` for this variable. I've done this in this commit: 59cbf8c207d39eb79de7619b60938bf73f79ee31

Also, this check is performed immediately after `make_err_c_result` variable initialization

tests/exceptions/exceptions.h
bader: > We using this namespace to avoid some redefinition errors or situations where we have variable/function with the same name

> We are going to use it in tests that we will provide soon. 

For this use case, I think something like this would work better:

```suggestion
namespace {

constexpr std::array<sycl::errc, 15> all_err_codes{
```

vasilytric: if we will use `unnamed namespace` with `all_err_codes` variable within a header file. ODR-using from multiple translation units (from future tests for exceptions) violates the one-definition rule because the definition of `all_err_codes` is not identical in all translation units.

bader: You can add `inline` to `all_err_codes` definition if future tests add ODR-uses.

I still don't see any value in using namespaces for `all_err_codes` definition, but it's your preference, please, avoid using generic name like exception as it doesn't solve "redefinition errors or situations where we have variable/function with the same name".
I suggest we either don't use namespace at all or use existing approaches: `sycl_cts` or anonymous namespace.

tests/exceptions/exceptions_make_error_code.cpp
```diff
@@ -0,0 +1,118 @@
...
+namespace TEST_NAMESPACE {
+
+using namespace sycl_cts;
+
+/** @brief Provide verification for sycl::make_error_code function
+ *  @param err_c Error code from sycl::errc enumeration
+ *  @param log sycl_cts::util::logger class object
+ */
+void check_sycl_working(sycl::errc err_c, util::logger &log) {
+  std::error_code make_err_c_result{sycl::make_error_code(err_c)};
```
bader, line 22:
```suggestion
  auto make_err_c_result{sycl::make_error_code(err_c)};
```
This way `make_err_c_result` type matches `sycl::make_error_code(err_c)` return type and `CHECK_TYPE` below does relevant check.

tests/exceptions/exceptions_make_error_code.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/** @brief Provide verification for same work std::error_code and
+ *         sycl::make_error_code
+ *  @param err_c Error code from sycl::errc enumeration
+ *  @param log sycl_cts::util::logger class object
+ */
+void compare_sycl_and_std_working(sycl::errc err_c, util::logger &log) {
+  auto err_c_result{
+      std::error_code(static_cast<int>(err_c), sycl::sycl_category())};
```
keryell, line 43: Do you mean
```suggestion
  std::error_code err_c_result { static_cast<int>(err_c), sycl::sycl_category{} };
```
?

ProGTX (approved):

tests/exceptions/exceptions_make_error_code.cpp
kopylovanat: Done.

tests/exceptions/exceptions_make_error_code.cpp
kopylovanat: Done.

tests/exceptions/exceptions.h
kopylovanat: Changed to anonymous namespace.

bader (approved):

keryell (approved):
Thanks.

PR #174, opened 2021-09-08
Introduce basic CI workflows

psalz:
During this phase of rapid evolution the developer experience surrounding the CTS tends to suffer quite a bit, as changes introduced with one SYCL implementation in mind often break compilation for other implementations. In particular, innocent looking changes to the CTS support libraries can have devastating effects, sometimes breaking compilation of *all* test cases. A recent example of this is the usage of `sycl::bit_cast` in the `CTS::util` library, which is supported by neither ComputeCpp nor hipSYCL at this point in time.

In an effort to improve upon this situation, I'm proposing to introduce basic CI workflows to ensure that the CTS still *compiles* for all supported SYCL implementations.

To this end, this PR introduces two workflows:

1) `build_ci_containers.yml`: Build Docker container images for each SYCL implementation

   The workflow definition file specifies the version of each SYCL implementation to be built. The idea here is that implementors can easily bump the version of their implementation whenever they see fit. This workflow runs periodically (nightly/daily, depending on where you are) and pushes the resulting container images to DockerHub (which will be a no-op most of the time, as new images only have to be built when a SYCL implementation's version changes).

2) `cts_ci.yml`: Run CI within Docker containers

   The actual CI part entails configuring CMake and compiling (parts of) the CTS. Crucially, **no tests are being executed**. This happens within the per-SYCL implementation Docker containers created in the previous workflow. This workflow runs on each pull request, highlighting compilation-breaking changes early.

In this first iteration, the workflows only build the two support libraries `oclmath` and `CTS::util`. I'm going to evaluate whether it makes sense to try and compile all test cases as well in a subsequent step (it may not be feasible because a full compilation can take quite a lot of time). Of course it does not make sense to expect every SYCL implementation to be able to compile all test
cases at all times, so per-implementation filters would have to be put into place (a feature that the CTS infrastructure already supports).

While the workflows won't be enabled until this PR is merged, you can see the results of the workflow runs in my fork of the CTS: Here's a [container image build](https://github.com/psalz/SYCL-CTS/actions/runs/1213279283), and here's a [CI run](https://github.com/psalz/SYCL-CTS/actions/runs/1213283033).

The CI run fails because the CTS support libraries currently do not not build correctly for hipSYCL (= the reason why I'm introducing this in the first place). I will address this in a subsequent PR.

Outstanding things to do:
- [x] Move container images to the [KhronosGroup namespace on DockerHub](https://hub.docker.com/u/khronosgroup), if       possible (currently they're hosted on my own account). Not sure who I have to talk to regarding that (cc @oddhack?)
  - I'm aware of the [KhronosGroup/DockerContainers](https://github.com/KhronosGroup/DockerContainers) repository, however I believe in this case it would be better to keep the Dockerfiles close to the CTS sources.
- [x] Currently there is no container for ComputeCpp, as there is no way of programmatically obtaining a copy of the ComputeCpp CE binaries. I'm in contact with Codeplay regarding that.

Sidenote: I've called the Intel SYCL implementation "DPC++" (or rather `dpcpp`), since I believe that is the term most people are familiar with (due to Intel's own marketing efforts). However so far the CTS has referred to it as "Intel SYCL". Any preferences from Intel's side here @bader @jbrodman @mkinsner?


bader:
> Sidenote: I've called the Intel SYCL implementation "DPC++" (or rather `dpcpp`), since I believe that is the term most people are familiar with (due to Intel's own marketing efforts). However so far the CTS has referred to it as "Intel SYCL". Any preferences from Intel's side here @bader @jbrodman @mkinsner?

I don't have strong preference regarding the name used for SYCL-CTS CI, but I should mentioned that correct name of "Intel SYCL implementation" is "Intel(R) oneAPI DPC++" and it's distributed in binary form as part of oneAPI toolchain and supports Intel platforms only (https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/dpc-compiler.html). 

There is also an open source implementation (https://github.com/intel/llvm) named "oneAPI DPC++" (i.e. w/o Intel in the name), which supports multiple platforms thanks to contributions made by the community.

We probably should update SYCL-CTS CMake files to use proper names as well.

oddhack:
There are only a couple of us who have push permissions on dockerhub khronosgroup. You could be added if the WG decides that's appropriate.

psalz:
> I don't have strong preference regarding the name used for SYCL-CTS CI, but I should mentioned that correct name of "Intel SYCL implementation" is "Intel(R) oneAPI DPC++" and it's distributed in binary form as part of oneAPI toolchain and supports Intel platforms only (https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/dpc-compiler.html).
> 
> There is also an open source implementation (https://github.com/intel/llvm) named "oneAPI DPC++" (i.e. w/o Intel in the name), which supports multiple platforms thanks to contributions made by the community.

Ah yes, not confusing at all ;-).
 
> We probably should update SYCL-CTS CMake files to use proper names as well.

Agreed!

psalz:
> There are only a couple of us who have push permissions on dockerhub khronosgroup. You could be added if the WG decides that's appropriate.

That's fine, actually. I don't need write access to the DockerHub repository, as long as someone who has can generate an access token and add it as a secret to this repository. Once added, it cannot be viewed through the GUI, and secrets are not passed to workflow runs triggered by pull requests -- so the risk should be minimal.

docker/dpcpp/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    git checkout $IMPL_VERSION
+
+RUN python3 /tmp/dpcpp/buildbot/configure.py \
+      --src-dir=/tmp/dpcpp \
+      --obj-dir=/tmp/build \
+      --build-type=Release \
+      --cmake-opt=-DCMAKE_INSTALL_PREFIX=/sycl && \
+    python3 /tmp/dpcpp/buildbot/compile.py \
+      --src-dir=/tmp/dpcpp \
+      --obj-dir=/tmp/build
```
bader, line 20:
```suggestion
      --obj-dir=/tmp/build \
      -j$NCPU
```
I think it make sense to set # of parallel executors.

bader (approved):
The main concern I have is that requiring to pass these checks for all changes introduces hard dependency on SYCL implementations i.e. effectively blocking some CTS changes unless certain features are supported by all implementations.

docker/dpcpp/Dockerfile
psalz: Ah, actually that environment variable no longer exists. I used to set it based on `getconf _NPROCESSORS_ONLN`, however I since then switched to building with Ninja, which automatically determines the number of available cores. I will remove it for hipSYCL as well.

psalz:
> The main concern I have is that requiring to pass these checks for all changes introduces hard dependency on SYCL implementations i.e. effectively blocking some CTS changes unless certain features are supported by all implementations.

I wouldn't call it a hard dependency. The main goal is to avoid the unintended breakage of tests that already compile for a given implementation. Adding/using newer features such as `sycl::bit_cast` is still possible, as long as it's not being done within one of those few big files (such as `math_reference.cpp`) that all tests depend on. For example a dedicated header file could be introduced that is included only by the tests that need that particular feature. Later down the line, once all implementations support the feature, the header can then be merged into one of the big files.

keryell (approved):

bader:
> > I don't have strong preference regarding the name used for SYCL-CTS CI, but I should mentioned that correct name of "Intel SYCL implementation" is "Intel(R) oneAPI DPC++" and it's distributed in binary form as part of oneAPI toolchain and supports Intel platforms only (https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/dpc-compiler.html).
> > There is also an open source implementation (https://github.com/intel/llvm) named "oneAPI DPC++" (i.e. w/o Intel in the name), which supports multiple platforms thanks to contributions made by the community.
> 
> Ah yes, not confusing at all ;-).
> 
> > We probably should update SYCL-CTS CMake files to use proper names as well.
> 
> Agreed!

Just merged https://github.com/KhronosGroup/SYCL-CTS/pull/175. Feel free to update build instructions for DPC++ implementation.

psalz:
> There are only a couple of us who have push permissions on dockerhub khronosgroup. You could be added if the WG decides that's appropriate.

@oddhack: The WG just greenlit this (cc @fraggamuffin). How do we proceed? Can you generate the token for us, should I, or someone else?

oddhack:
> > There are only a couple of us who have push permissions on dockerhub khronosgroup. You could be added if the WG decides that's appropriate.
> 
> @oddhack: The WG just greenlit this (cc @fraggamuffin). How do we proceed? Can you generate the token for us, should I, or someone else?

I believe @outofcontrol can set you up with dockerhub khronosgroup membership - I probably could in principle but in reality have no idea how to do it :-)

psalz:
Okay, so I've:
- Switched workflows to the `khronosgroup` DockerHub namespace and added an access token to this repository
- Added support for ComputeCpp by manually pushing the 2.6.0 binaries to DockerHub
- Changed `Intel_SYCL` to `DPCPP`
- Removed the superfluous `-j$NCPU`

I think we can merge this!

bader:
@illuhad, @ProGTX, does it look good to you?

illuhad (approved):
LGTM. In the future it might be interesting to also compile for a device backend with hipSYCL since the CPU backend is a bit more permissive in some respects. (And the hipSYCL cmake support in CTS should probably also be updated)

bader:
> LGTM. In the future it might be interesting to also compile for a device backend with hipSYCL since the CPU backend is a bit more permissive in some respects. (And the hipSYCL cmake support in CTS should probably also be updated)

Agree. I think the same applies to execution on CPU using OpenCL back-end with DPC++. I hope Intel's OpenCL CPU implementation can be successfully deployed to GitHub Action bots.

psalz:
I'd like to hold off on merging this until we've gotten confirmation from Codeplay that pushing the CCPP headers & binaries to DockerHub is okay with them (cc @rodburns).

I've also just found a bug regarding Docker's `--build-arg`, will push a fix shortly.

psalz:
It appears that sorting out the remaining legal questions surrounding ComputeCpp may take some more time. I'd like to move forward with this PR regardless, as I have several other things queued up that depend on CI being in place. I've therefore disabled CI for ComputeCpp for the time being in 627029d.

Quick summary of the other new changes:
- c8ef863 : Turns out I was not actually building the intended git revision for each implementation (but rather simply the current `HEAD`).
- ee1ec88  : Quite self-explanatory, reduce container images sizes, which should make CI runs (ever so slightly) faster and waste less bandwidth.
- 12507c5 : Forward additional arguments passed to the `configure.sh` script to CMake. This allows us to change build configurations from within the CI workflow files, without having to rebuilt the container images first.

@bader from my side this is ready to merge now!

bader:
@ProGTX, are you okay to merge this?

PR #179, opened 2021-09-16
(Removed API) Remove tests for sycl::program and its usage

kovshikovser:
Remove all tests for removed sycl::program
Replace sycl::program with sycl::kernel_bundle


bader:
> Remove all tests for deprecated sycl::program

FYI: `sycl::program` is not deprecated, it's removed!
Deprecated features still must be tested, so we can't remove them from the tests.

bader (approved):

ProGTX (approved):

tests/kernel/kernel_constructors.cpp
```diff
@@ -162,109 +157,121 @@ class TEST_NAME : public sycl_cts::util::test_base {
...
+        auto kernelC = kbC.get_kernel(sycl::get_kernel_id<k_name5>());
+        ctsQueue.submit([&](sycl::handler &cgh) {
+          cgh.single_task(k_name5());
+        });
+        kernelC = (kernelA);
+
+        using k_name6 = test_kernel<6>;
+        auto kbD = sycl::get_kernel_bundle<
+                      k_name6, sycl::bundle_state::executable>(ctx);
+        auto kernelD = kbD.get_kernel(sycl::get_kernel_id<k_name6>());
```
keryell, line 183: @gmlueck do you think it is interesting to provide in the specification a short-cut for this pattern?

keryell (commented):
Nice to see the difference between `program` and `kernel_bundle`

tests/kernel/kernel_constructors.cpp
gmlueck: Yes, I think that would be easy to do.  If we do this, we should add similar overloads for `has_kernel`.  I think the implementation would be like:

```
class kernel_bundle {
 public:
  template<KernelName>
  bool has_kernel() { return has_kernel(get_kernel_id<KernelName>()); }

  template<KernelName>
  bool has_kernel(const device &dev) { return has_kernel(get_kernel_id<KernelName>(), dev); }

  template<KernelName>
  bool get_kernel() { return get_kernel(get_kernel_id<KernelName>()); }
};
```

@bader: Do you see any reason why this would cause problems?

bader: > @bader: Do you see any reason why this would cause problems?

No. :)

bader (approved):

tests/kernel/kernel_constructors.cpp
gmlueck: Shortcuts added to the spec in KhronosGroup/SYCL-Docs#194

keryell (approved):

kovshikovser:
Added very small fix for check_type_existence usage in tests/kernel/kernel_info.cpp

psalz:
Unfortunately this [broke CI again](https://github.com/KhronosGroup/SYCL-CTS/runs/3789470508?check_suite_focus=true), ComputeCpp doesn't have `sycl::get_kernel_bundle`...

tests/common/common.h
```diff
@@ -591,9 +591,9 @@ inline bool kernel_supports_wg_size(sycl_cts::util::logger& log,
     return false;
   }
 
-  sycl::program program(context, devicesToCheck);
-  program.build_with_kernel_type<kernelT>("");
-  auto kernel = program.get_kernel<kernelT>();
+  auto kb =
+      sycl::get_kernel_bundle<kernelT, sycl::bundle_state::executable>(context);
```
bader, line 595: @kovshikovser, it seems like we should wrap this code with `#if !defined(__COMPUTECPP__)` and keep old API in `#else` branch to fix the build with ComputeCPP. 

@psalz. am I right?

psalz: I think so, yes!

tests/common/get_cts_object.h
```diff
@@ -92,68 +92,10 @@ struct get_cts_object {
...
-        cgh.single_task(dummy_functor<kernel_name>());
-      });
-      q.wait_and_throw();
-      program.build_with_kernel_type<dummy_functor<kernel_name>>(buildOptions);
-
-      return program;
+      auto ctx = queue.get_context();
+      auto kb_exe = sycl::get_kernel_bundle<
+                      kernel_name, sycl::bundle_state::executable>(ctx);
+      return kb_exe.get_kernel(sycl::get_kernel_id<kernel_name>());
```
psalz, line 98: Same here.

tests/common/common.h
psalz: Actually hipSYCL doesn't have `get_kernel_bundle` either yet. The reason why passes is that hipSYCL doesn't compile `test_base_opencl.cpp`, which includes these common headers. So the check should be `#if !defined(__COMPUTECPP__) && !defined(__HIPSYCL__)`!

psalz: Addressed in #189.

PR #186, opened 2021-09-23
Fix compilation of CTS support libraries for ComputeCpp and hipSYCL

psalz:
As discussed during the WG meeting, these are a few bandaid fixes to enable compilation of the two CTS support libraries `oclmath` and `CTS::util` with both ComputeCpp and hipSYCL. Since we don't have CI in place for ComputeCpp yet unfortunately, you'll just have to take my word for it ;-).

As part of this change I've gone ahead and removed the `common/sycl.h` wrapper header, as the namespace alias (introduced in #108) was breaking hipSYCL since ef7cd6ab5. This header mostly disabled some warnings for MSVC; given the comment ("Todo: Remove") I'm assuming this has been out of date for a while.


bader (approved):

ProGTX (approved):

keryell (approved):
Good.

PR #187, opened 2021-09-26
Do not rely on implementations to include <cmath>

alexbatashev:
Looks like implementations implicitly include `<cmath>` to implement builtins. However, that is not required by the spec.


ProGTX (approved):

keryell (approved):
Good point!

PR #190, opened 2021-10-04
Move try-catch to test_base w/o formatting changes

psalz:
This is essentially the same as #167, but by keeping the `try-catch` blocks' curly braces around, (almost) no formatting changes are required. This makes the patch easier to review and less error prone. Includes changes by @vasilytric that move the `try-catch` block to a new `test_base.cpp` file; I've taken the liberty of squashing some of the later fixes. I've confirmed that the set of tests that compiles for each of the three SYCL implementations remains unchanged.

Includes #189 to allow for CI to pass.

Supersedes #167.

### DIFF
--- a/tests/usm/usm_allocate_free.h
+++ b/tests/usm/usm_allocate_free.h
@@ -418,17 +418,7 @@ class check_usm_allocate_free {
  */
 template <typename T, typename op>
 static void run_usm_test(util::logger &log) {
-  try {
-    check_usm_allocate_free<T, op>{}(log);
-  } catch (const sycl::exception &e) {
-    log_exception(log, e);
-    std::string errorMsg =
-        "a SYCL exception was caught: " + std::string(e.what());
-    FAIL(log, errorMsg);
-  } catch (const std::exception &e) {
-    std::string errorMsg = "an exception was caught: " + std::string(e.what());
-    FAIL(log, errorMsg);
-  }
+  { check_usm_allocate_free<T, op>{}(log); }
 }
 
 }  // namespace usm_allocate_free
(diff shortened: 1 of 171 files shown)


tests/usm/usm_allocate_free.h
```diff
@@ -418,17 +418,7 @@ class check_usm_allocate_free {
...
-  } catch (const sycl::exception &e) {
-    log_exception(log, e);
-    std::string errorMsg =
-        "a SYCL exception was caught: " + std::string(e.what());
-    FAIL(log, errorMsg);
-  } catch (const std::exception &e) {
-    std::string errorMsg = "an exception was caught: " + std::string(e.what());
-    FAIL(log, errorMsg);
-  }
+  { check_usm_allocate_free<T, op>{}(log); }
```
bader, line 421:
```suggestion
  check_usm_allocate_free<T, op>{}(log);
```
We modify this line anyway, so we can drop extra braces here.

tests/usm/usm_allocator_api.cpp
```diff
@@ -27,18 +27,7 @@ class TEST_NAME : public sycl_cts::util::test_base {
...
-      log_exception(log, e);
-      std::string errorMsg =
-          "a SYCL exception was caught: " + std::string(e.what());
-      FAIL(log, errorMsg);
-    } catch (const std::exception &e) {
-      std::string errorMsg =
-          "an exception was caught: " + std::string(e.what());
-      FAIL(log, errorMsg);
-    }
+    { check_usm_allocator_api<TestType, allocate_general>{}(log); }
```
bader, line 30:
```suggestion
    check_usm_allocator_api<TestType, allocate_general>{}(log);
```
We should apply this change to all USM tests.

ProGTX (approved):

tests/usm/usm_allocator_api.cpp
psalz: Done!

PR #191, opened 2021-10-05
Overhaul device selection mechanism

### LINKED ISSUE #185: Simplify CLI device selection mechanism
The current mechanism for selecting a device to run the CTS on requires the user to specify a combination of SYCL device type and one out of a hard-coded list of vendor names for platform selection (e.g. `Intel` or `ARM`). This is both outdated (in that it distinguishes between a "host device" and "OpenCL device") and -- in my opinion -- needlessly cumbersome. 

I would like to propose a simpler yet more flexible alternative based on substring matching. Instead of the four existing CLI parameters `--platform` / `-p` / `--device` / `-d`, there would only be two: The existing `--device` and a new `--list-devices`. While the latter can be used to list all devices currently available (including their device type and platform name), the former is used for selecting one of the devices based on substring matching (and optionally regex syntax).

Examples:
```
> ./bin/test_all --list-devices --device "OpenCL.*Xe"
4 devices available (* = currently selected):
  Type         Platform / Device
  cpu          Intel(R) OpenCL / 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
* gpu          Intel(R) OpenCL HD Graphics / Intel(R) Iris(R) Xe Graphics [0x9a49]
  gpu          Intel(R) Level-Zero / Intel(R) Iris(R) Xe Graphics [0x9a49]
  host         SYCL host platform / SYCL host device
```

```
> ./bin/test_all --list-devices --device "Level-Zero.*Xe"
4 devices available (* = currently selected):
  Type         Platform / Device
  cpu          Intel(R) OpenCL / 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
  gpu          Intel(R) OpenCL HD Graphics / Intel(R) Iris(R) Xe Graphics [0x9a49]
* gpu          Intel(R) Level-Zero / Intel(R) Iris(R) Xe Graphics [0x9a49]
  host         SYCL host platform / SYCL host device
```

In this case I've chosen to match over not just the device name, but the platform name as well (as listed, i.e., `<platform name> / <device name>`). The `--device` flag could therefore also be called `--platform-device`  or something, but that may be a bit awkward.

psalz:
This changes the way of selecting the platform/device to run the CTS on from a hard-coded list of options to a mechanism based on substring matching.

The options `--platform`, `-p`, `--device` and `-d` are replaced by a single `--device` option that receives a string that is interpreted as an ECMAScript regular expression for matching against all available devices reported by the SYCL implementation. Additionally, a new `--list-devices` option is introduced that lists all aforementioned devices, additionally indicating the device currently matched against by `--device`.

Additional changes:
   - If no `--device` is given, fall back to the default selector.
   - Replace `[host|opencl]_[platform|device]_name` CMake configuration parameters by `SYCL_CTS_CTEST_DEVICE`. This is really only relevant to the conformance report generation script, or when running the tests manually through CTest.
   - Update runtests.py to receive single `--device` option.
   - Fix some additional issues in runtests.py script.

Example output:
```
> ./bin/test_all --list-devices --device "i7"
4 devices available (> = currently selected):
  Type         Platform / Device
> cpu          Intel(R) OpenCL / 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
  gpu          Intel(R) OpenCL HD Graphics / Intel(R) Iris(R) Xe Graphics [0x9a49]
  gpu          Intel(R) Level-Zero / Intel(R) Iris(R) Xe Graphics [0x9a49]
  host         SYCL host platform / SYCL host device
```

Notes:
   - I did not update the run_conformance_tests.py script, because it is currently broken in more than one way and in need of a general overhaul (which I *may* do in an upcoming PR).
   - As of ComputeCpp 2.6.0, the `sycl::device::operator==` seems to be broken, so `--list-devices` does not correctly indicate the currently selected device (cc @ProGTX).

Closes #185.


bader (approved):

util/test_manager.h
```diff
@@ -43,6 +46,12 @@ class test_manager : public singleton<test_manager> {
    */
   bool wimpy_mode_enabled() const;
 
+  /**
+   * @return The regex set by the `--device` CLI parameter, used for selecting
+   * the CTS device.
+   */
+  const std::optional<std::regex>& get_device_regex() { return device_regex; }
```
ProGTX, line 53:
```suggestion
  const std::optional<std::regex>& get_device_regex() const { return device_regex; }
```

tests/common/cts_selector.h
```diff
@@ -20,116 +22,34 @@ class cts_selector : public sycl::device_selector {
...
 
-    // Early exit for host device
-    if (dev.is_host()) {
-      if (ctsDevType == selector::ctsdevice::host) {
-        return 1000;
-      } else {
-        return -1;
-      }
+    if (!device_regex.has_value()) {
+      return sycl::default_selector{}(dev);
```
keryell, line 41: In the future we need to modernize the device selector API too instead of relying on deprecated SYCL 1.2.1 one.

keryell (approved):

runtests.py
```diff
@@ -494,21 +489,16 @@ def parse_args( ):
...
     parser = argparse.ArgumentParser( description="Khronos SYCL CTS" )
 
     parser.add_argument( "-b", "--binpath", help="specify path to the cts executable file" )
 
     parser.add_argument( "--csvpath", help="specify path to csv file for filtering tests" )
     parser.add_argument( "--list", help="list all tests in a test binary", action="store_true" )
     parser.add_argument( "-j", "--junit", help="specify output path for a junit xml file" )
-    parser.add_argument( "-p", "--platform", choices=platforms, help="The platform to run on " )
-    parser.add_argument( "-d", "--device", choices=devices, help="The device to run on " )
+    parser.add_argument( "--device", help="The device to run on" )
```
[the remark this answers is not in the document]
psalz: I could argue that people likely aren't running the CTS frequently enough that a shorthand is needed; however the actual reason is that Catch2's CLI already has a use for `-d` (show test durations), so I've removed the shorthand in anticipation of the migration -- I was trying to be sneaky.

tests/common/cts_selector.h
psalz: Right, ComputeCpp doesn't support `sycl::default_selector_v` yet, though.

ProGTX: ComputeCpp added support for `default_selector_v` in version 2.3, and in general should support all the new device selection mechanisms since 2.6.

PR #192, opened 2021-10-05
Rework Configuration-Time Test Category Filtering Mechanism

psalz:
I'm proposing to change the mechanism for filtering the test categories (= subfolders inside `tests/` containing a `CMakeLists.txt` file) during *configuration time*.

Whereas the previous approach (`SYCL_CTS_TEST_FILTER`) used to specify all test categories that are to be included in the compilation, I'm proposing to do the opposite, i.e., list all categories that are to be excluded. To underline this change, I've renamed the CMake config parameter to `SYCL_CTS_EXCLUDE_TEST_CATEGORIES`.

Switching from inclusion to exclusion has the benefit that (when a filter is used) newly added tests are being compiled by default, instead of having to be added to a list manually. If we end up compiling tests during CI (which I'm still very much planning on doing), we will have to limit the set of tests compiled by each implementation somehow (as no implementation can currently compile all tests). Using an exclusion filter will therefore highlight potential problems for different SYCL implementations early.

I've added filters for ComputeCpp, DPCPP and hipSYCL, reflecting their current state of CTS support. **It would be great if someone could confirm these** on their setup (either use them and check that everything else compiles, or ideally re-generate the filters -- see below).

Having these filters within the repository for each implementation also serves as a nice way of tracking the current state of CTS support, where an empty list indicates that all tests compile successfully. It also shows which test categories we may want to focus on most. For example, currently none of the implementations seem to be able to compile the `accessor` tests.

As tests and implementations improve over time, these filters should be regenerated and updated within the repository. To streamline this process, I'm introducing a new utility script called `generate_exclude_filter.py` that makes this rather simple. Basically, the script attempts to compile all targets for a given SYCL implementation, spitting out the failing ones, which can then be used as a filter.

I've removed the ability to specify a filter when generating a conformance report using the `run_conformance_tests.py` script. Frankly I never quite understood why this feature was there (and mandatory!) in the first place. We can re-introduce it should the need arise.

Lastly, I've removed the `conformance_filters` directory containing the `core.csv` and `codeplay.csv` filters.


ci/dpcpp.filter
```diff
@@ -0,0 +1,15 @@
...
+id
+invoke
+item
+math_builtin_api
+multi_ptr
+nd_item
+range
+reduction
+specialization_constants
+usm
```
bader, line 15: I can build all these tests with [yesterday build](https://github.com/intel/llvm/releases/tag/sycl-nightly%2F20211004) except:
buffer
reduction
specialization_constants 
usm

How can I reproduce your results?

psalz: I've tried it for the [version currently used in CI](https://github.com/KhronosGroup/SYCL-CTS/blob/23d2b9be80fcd08b48e5e4cb73d1efdd186e6b07/.github/workflows/build_ci_containers.yml#L44), i.e., `f5c838cd`. Sounds like we maybe should bump that! :-).

bader: Right. I think https://github.com/intel/llvm/pull/4587 fixes build of several tests.

ci/generate_exclude_filter.py
```diff
@@ -0,0 +1,218 @@
+#!/usr/bin/env python3
+
+# NOTE: This script requires Python >= 3.7 and CMake >= 3.15
```
bader, line 3: Should we bump CMake requirements?
https://github.com/KhronosGroup/SYCL-CTS#compilation-and-execution-prerequisites requires CMake 3.13.

bader (approved):

ci/dpcpp.filter
bader: Let's update it in a separate change.

run_conformance_tests.py
```diff
@@ -317,14 +316,13 @@ def main(argv=sys.argv[1:]):
 
     # Parse and gather all the script args
     (cmake_exe, build_system_name, build_system_call, full_conformance,
-     conformance_filter, implementation_name, additional_cmake_args, host_names,
+     implementation_name, additional_cmake_args, host_names,
```
ProGTX, line 319: We use this script in internal testing, could we please keep the filter functionality in some form?

ci/computecpp.filter
```diff
@@ -2,40 +2,30 @@ accessor
 address_space
```
ProGTX, line 2: Does this mean that all of the remaining tests are failing with ComputeCpp? That's a very long list, is that because of missing SYCL 2020 features?

ci/generate_exclude_filter.py
ProGTX: This script also requires ninja

run_conformance_tests.py
psalz: Okay sure. Would using the exclude filter work for you though? It should be simple enough to switch from a whitelist to a blacklist, right?

ci/generate_exclude_filter.py
bader: I'm okay to update prerequisites listed in README file with follow-up PR(s).

run_conformance_tests.py
psalz: Great, then I will address this in my follow-up PR that fixes the `run_conformance_tests` script, if that's alright with you!

ProGTX: That's fine, thank you

psalz:
Any more concerns with this, or can I go ahead and merge?

PR #194, opened 2021-10-14
Fix/improve conformance report generation script

psalz:
As promised in #192, this fixes various problems with the conformance report generation script and updates it to properly handle recently introduced mechanisms/changes:

- Add support for new exclude list filter (via the `--exclude-categories` parameter, which is optional)
- Add support for new `--device` regex parameter
- Retire notion of distinct "Host Device" and "OpenCL Device";  only a single device is tested for conformance
- Remove support for listing device extensions that are no longer reported by the `test_manager`
- Fix forwarding of multiple `--additional-cmake-args`
- Reformat stylesheet.xml to use LF line endings and space indentations

Here's what a report generated for hipSYCL looks like:
image

### DIFF
--- a/run_conformance_tests.py
+++ b/run_conformance_tests.py
@@ -29,14 +29,12 @@ def handle_args(argv):
     parser.add_argument(
         '-a',
         '--additional-cmake-args',
-        help=
-        'Additional args to hand to CMake required by the tested implementation.',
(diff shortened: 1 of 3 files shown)


illuhad (approved):

psalz:
I'm going to go ahead and merge this; we can always come back to it for further tweaks in future PRs.

PR #198, opened 2021-10-14
Unify naming of full-conformance macro and option, fix its usage, reduce compile times

psalz:
Following up on #197, I was looking for ways of reducing compile times. While CI is still running on #197 at the time of writing, the results of the per-translation unit timings are (spoiler alert!) mostly confirming what we already knew: The generated vector tests are taking up most of the compile time.

As you know, we already have the `SYCL_CTS_ENABLE_FULL_CONFORMANCE` option to rein in compile times somewhat. However, even with the option disabled (which it is for CI), compile times are arguably still too long. While looking into why vector tests take that long to compile even in non-full conformance mode, I've noticed that some tests were not correctly using the macro to guard portions of the code: Some tests erroneously used the CMake option's name, as opposed to the (confusingly similar) macro name (`SYCL_CTS_FULL_CONFORMANCE`). Some where even still using the outdated `SYCL_CTS_EXTENSIVE_MODE` macro.

To avoid this confusion in the future, I'm proposing to unify the option's and macro's name to both simply be `SYCL_CTS_ENABLE_FULL_CONFORMANCE`. I've adjusted all relevant locations accordingly.

Additionally, I'm proposing to reduce compile times in non-full conformance mode by drastically reducing the set of data types for which vector tests are being generated.

Note: This PR includes both #195 and #197, so those should be merged first.

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  compile-cts-support-libraries:
+  compile-cts:
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
@@ -13,6 +13,7 @@ jobs:
         sycl-impl: [computecpp, dpcpp, hipsycl]
     env:
       container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
+      parallel-build-jobs: 2
     container:
       # image: khronosgroup/sycl-cts-ci:${{ matrix.sycl-impl }}
       # ComputeCpp images are hosted by Codeplay
@@ -24,10 +25,15 @@ jobs:
           submodules: true
       - name: Configure CMake
         working-directory: ${{ env.container-workspace }}
-        run: bash /scripts/configure.sh
+        run: |
+          bash /scripts/configure.sh \
+            -DSYCL_CTS_EXCLUDE_TEST_CATEGORIES="${{ env.container-workspace }}/ci/${{ matrix.sycl-impl }}.filter"
       - name: Build 'oclmath'
         working-directory: ${{ env.container-workspace }}/build
         run: cmake --build . --target oclmath
       - name: Build 'util'
         working-directory: ${{ env.container-workspace }}/build
         run: cmake --build . --target util
+      - name: Build all supported test categories
+        working-directory: ${{ env.container-workspace }}/build
+        run: cmake --build . --target test_all -- -j${{ env.parallel-build-jobs }}
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -39,7 +39,7 @@ jobs:
         working-directory: ${{ env.container-workspace }}/build
         run: |
           TS_BEFORE=$(date +%s)
-          cmake --build . --target test_all -- -j${{ env.parallel-build-jobs }}
+          cmake --build . --target test_all --parallel ${{ env.parallel-build-jobs }}
           TS_AFTER=$(date +%s)
           ELAPSED=$(($TS_AFTER - $TS_BEFORE))
           sort -rn -o build_times.log build_times.log
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -27,7 +27,8 @@ jobs:
         working-directory: ${{ env.container-workspace }}
         run: |
           bash /scripts/configure.sh \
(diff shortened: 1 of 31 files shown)


.github/workflows/cts_ci.yml
```diff
@@ -24,10 +25,27 @@ jobs:
...
         working-directory: ${{ env.container-workspace }}/build
         run: cmake --build . --target util
+      - name: Build all supported test categories
+        working-directory: ${{ env.container-workspace }}/build
+        run: |
+          TS_BEFORE=$(date +%s)
+          cmake --build . --target test_all -- -j${{ env.parallel-build-jobs }}
+          TS_AFTER=$(date +%s)
+          ELAPSED=$(($TS_AFTER - $TS_BEFORE))
+          sort -rn -o build_times.log build_times.log
```
keryell, line 45: In general avoid abbreviations now we do no longer rely on punch tapes to exchange our open-source programs (which I did long time ago... :-) )
```suggestion
          sort --numeric-sort --reverse --output=build_times.log build_times.log
```

.github/workflows/cts_ci.yml
```diff
@@ -24,10 +25,27 @@ jobs:
...
+          echo "Total time: $( date -d@$ELAPSED -u '+%-Hh %-Mm %-Ss' )" >> build_times.log
```
keryell, line 46:
```suggestion
          echo "Total time: $( date --date=@$ELAPSED --utc '+%-Hh %-Mm %-Ss' )" >> build_times.log
```

keryell:
> (spoiler alert!) mostly confirming what we already knew: The generated vector tests are taking up most of the compile time.

Since I have moved the triSYCL CPU unit tests from Travis-CI to GitHub Action which seems quite more constrained and with only 1 core by default, the tests fail because of the time limit, mainly because of the vector unit test compilation time.
When run in a Docker image, there are 2 cores and it passes! \o/

.github/workflows/cts_ci.yml
psalz: Yes you are right, I've addressed this as part of #197, which is where this change is coming from!

psalz:
> Since I have moved the triSYCL CPU unit tests from Travis-CI to GitHub Action which seems quite more constrained and with only 1 core by default, the tests fail because of the time limit, mainly because of the vector unit test compilation time. When run in a Docker image, there are 2 cores and it passes! \o/

Interesting... Thankfully we are already running inside Docker. But yes, the machines are pretty bare bones unfortunately (although it is free so we shouldn't complain).

.github/workflows/cts_ci.yml
```diff
@@ -24,10 +25,27 @@ jobs:
...
         working-directory: ${{ env.container-workspace }}/build
         run: cmake --build . --target oclmath
       - name: Build 'util'
         working-directory: ${{ env.container-workspace }}/build
         run: cmake --build . --target util
+      - name: Build all supported test categories
+        working-directory: ${{ env.container-workspace }}/build
+        run: |
+          TS_BEFORE=$(date +%s)
+          cmake --build . --target test_all -- -j${{ env.parallel-build-jobs }}
```
keryell, line 42: While you are here :-)
```suggestion
          cmake --build . --target test_all -- --parallel  ${{ env.parallel-build-jobs }}
```
I am unsure about the `--` in the middle. It means that you push the option to be handled by the underlying build tool textually.
If you do not have `--`, this is actually an option understood by CMake which will be translated in the right syntax for the underlying tool.
https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project

psalz: You are getting quite pedantic on me here... Fixed

CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ option(SYCL_CTS_ENABLE_FULL_CONFORMANCE
        "Enable full conformance mode with extensive tests" OFF)
 if(SYCL_CTS_ENABLE_FULL_CONFORMANCE)
     message(STATUS "Full conformance mode: ON")
-    add_definitions(-DSYCL_CTS_FULL_CONFORMANCE)
+    add_definitions(-DSYCL_CTS_ENABLE_FULL_CONFORMANCE)
```
ProGTX, line 34:
```suggestion
    add_host_and_device_compiler_definitions(-DSYCL_CTS_ENABLE_FULL_CONFORMANCE)
```

ProGTX (approved):

.github/workflows/cts_ci.yml
keryell: I noticed with you some sense of deutsche Qualität so I take advantage of it. ;-)

PR #199, opened 2021-10-15
Add test for sycl::errc_for

vasilytric:
This test provide verification that `sycl::errc_for` implemented according to SYCL-2020 revision 3

### DIFF
--- a/tests/exceptions/exceptions_errc_for.cpp
+++ b/tests/exceptions/exceptions_errc_for.cpp
@@ -1,0 +1,102 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for sycl::errc_for
+//
+*******************************************************************************/
+
+#include "exceptions.h"
+
+#define TEST_NAME exceptions_errc_for
+
+namespace TEST_NAMESPACE {
+
+template <template <sycl::backend> class arg>
+struct check_template_exists {};
+
+bool check_opencl_supporting(const sycl::queue &q,
+                             sycl_cts::util::logger &log) {
+  bool opencl_supported{};
+#ifdef SYCL_BACKEND_OPENCL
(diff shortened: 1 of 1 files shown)


tests/exceptions/exceptions_errc_for.cpp
```diff
@@ -0,0 +1,102 @@
...
+        FAIL(log, "sycl::errc_for is a error condition enumeration");
+      }
+      if (sycl::error_category_for<sycl::backend::opencl>().name() !=
+          "opencl") {
+        FAIL(log,
+             "sycl::error_category_for<sycl::backend::opencl> name is not "
+             "equal to \"opencl\"");
+      }
+#endif  // SYCL_BACKEND_OPENCL
+    } catch (const sycl::exception &e) {
```
psalz, line 87: Wrapping tests in try-catch blocks should no longer be required since #190 has been merged!

keryell (approved):

tests/exceptions/exceptions_errc_for.cpp
```diff
@@ -0,0 +1,89 @@
...
+#define TEST_NAME exceptions_errc_for
+
+namespace TEST_NAMESPACE {
+
+template <template <sycl::backend> class arg>
+struct check_template_exists {};
+
+bool check_opencl_supporting(const sycl::queue &q,
+                             sycl_cts::util::logger &log) {
+  bool opencl_supported{};
```
ProGTX, line 20:
```suggestion
  bool opencl_supported{false};
```

ProGTX (approved):

tests/exceptions/exceptions_errc_for.cpp
vasilytric: I've applied this comment in this commit: 4cc40890ad0cab86cb55479a71fe92b096c2be18

PR #200, opened 2021-10-15
Add tests for sub_group_mask simple const member functions

vasilytric:
`sub_group_mask` it's an extension, documentations can be found here: [intel/llvm/doc/SubGroupMask](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/SubGroupMask/SubGroupMask.asciidoc)

`Makefiles` will not be generated for extension tests, to build them should be provided a flag to `CMake` command, also tests on `one API` can be found in `tests/extension_one_api` folder


tests/sub_group_mask/sub_group_mask_common.h
```diff
@@ -0,0 +1,142 @@
...
+                    unsigned long long>{
+        "char",           "signed char", "unsigned char",     "short",
+        "unsigned short", "int",         "unsigned int",      "long",
+        "unsigned long",  "long long",   "unsigned long long"};
+
+template <typename funT, typename PredT, typename T>
+class test_kernel;
+
+constexpr size_t globalSize = 128;
+constexpr size_t localSize = 32;
```
bader, line 30:
```suggestion
constexpr size_t globalSize = 8;
constexpr size_t localSize = 4;
```
Why these values? Shouldn't they depend on sub-group size supported by HW?

tests/sub_group_mask/sub_group_mask_common.h
```diff
@@ -0,0 +1,142 @@
...
+#ifdef SYCL_EXT_ONEAPI_SUB_GROUP_MASK
+namespace {
+
+static const auto types =
+    named_type_pack<char, signed char, unsigned char, short, unsigned short,
+                    int, unsigned int, long, unsigned long, long long,
+                    unsigned long long>{
+        "char",           "signed char", "unsigned char",     "short",
+        "unsigned short", "int",         "unsigned int",      "long",
+        "unsigned long",  "long long",   "unsigned long long"};
```
bader, line 24: I suggest we reduce the list of tested types by default and test the rest in "full mode".
Something like this: https://github.com/KhronosGroup/SYCL-CTS/pull/198/files#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0d

tests/sub_group_mask/sub_group_mask_common.h
```diff
@@ -0,0 +1,142 @@
...
+  bool operator()(const sycl::sub_group &sub_group) {
+    return sub_group.get_local_id().get(0) <
+           sub_group.get_local_range().get(0) / 2;
+  }
+};
+
+struct second_half_predicate {
+  bool operator()(const sycl::sub_group &sub_group) {
+    return sub_group.get_local_id().get(0) >
+           sub_group.get_local_range().get(0) / 2 - 1;
```
bader, line 62:
```suggestion
    return sub_group.get_local_id().get(0) >=
           sub_group.get_local_range().get(0) / 2;
```
If local range is `1`, it definitely gives wrong result.

tests/sub_group_mask/sub_group_mask_common.h
```diff
@@ -0,0 +1,142 @@
...
+void check_mask_api(sycl_cts::util::logger &log) {
+  sycl::range<1> globalRange(globalSize);
+  sycl::range<1> localRange(localSize);
+  sycl::nd_range<1> dataRange(globalRange, localRange);
+
+  auto resultArr = get_result_array();
+  auto resultTypeArr = get_result_array();
+
+  auto testQueue = sycl_cts::util::get_cts_object::queue();
+  try {
```
bader, line 91: I think try-catch is not necessary here.
```suggestion
```

tests/sub_group_mask/sub_group_mask_common.h
```diff
@@ -0,0 +1,142 @@
...
+  resultArr.fill(true);
+  return resultArr;
+}
+
+inline bool if_check(const sycl::sub_group &sub_group) {
+#ifndef SYCL_CTS_ENABLE_FULL_CONFORMANCE
+  return sub_group.leader();
+#else
+  return true;
+#endif
```
bader, line 77: Avoiding negation makes code easier to read.
```suggestion
#ifdef SYCL_CTS_ENABLE_FULL_CONFORMANCE
  return true;
#else
  return sub_group.leader();
#endif
```

tests/sub_group_mask/sub_group_mask_find_high.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace TEST_NAMESPACE {
+
+using namespace sycl_cts;
+#ifdef SYCL_EXT_ONEAPI_SUB_GROUP_MASK
+
+struct check_result_find_high {
+  bool operator()(const sycl::ext::oneapi::sub_group_mask &sub_group_mask,
+                  const sycl::sub_group &sub_group) {
+    return sub_group_mask.find_high() ==
+           sycl::id(sub_group.get_local_range().get(0) / 2 - 1);
```
bader, line 22: Does it work for range `1`? Please, make sure that computed id stays within the legal range.
```suggestion
           sycl::id((sub_group.get_local_range().get(0) + 1) / 2 - 1);
```

tests/sub_group_mask/sub_group_mask_subscript.cpp
```diff
@@ -0,0 +1,60 @@
...
+using namespace sycl_cts;
+#ifdef SYCL_EXT_ONEAPI_SUB_GROUP_MASK
+
+struct check_result_subscript {
+  bool operator()(const sycl::ext::oneapi::sub_group_mask &sub_group_mask,
+                  const sycl::sub_group &) {
+    for (size_t N = 0; N < sub_group_mask.size(); N++) {
+      if (sub_group_mask[sycl::id(N)] != (N % 2 == 0)) return false;
+    }
+    return true;
```
bader, line 24: Does it make sense to scatter the test code among 9 files? Just 4-5 of 60 lines in this file are specific to the test, the rest are duplicated.

keryell (commented):
I am puzzled by the goal of these tests since there is no comments about what is actually tested.
Since there are a lot of  `#ifdef SYCL_EXT_ONEAPI_SUB_GROUP_MASK` all over the place, do we test here the concept of an extension in a SYCL implementation and `sycl::ext::oneapi::sub_group_mask` is just an example used to test the extension mechanism?

tests/sub_group_mask/sub_group_mask_all.cpp
```diff
@@ -0,0 +1,69 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Extension Conformance Test
```
ProGTX, line 3: I'm a bit confused as well, this is an Intel extension, not a Khronos one, why would it be in the CTS? In any case the CTS would probably need some extra mechanism to enable/disable SYCL extensions, here's it's just added in full.

ProGTX (changes requested):

tests/sub_group_mask/sub_group_mask_common.h
vasilytric: I've applied interactions with different `sub_group_size` in this commit: eb0ad1d745453a6be2666cc4e162a173debfb675
If this fix is OK, then I'll apply this change to another PRs

tests/sub_group_mask/sub_group_mask_common.h
vasilytric: this `try-catch` block was deleted [here](https://github.com/KhronosGroup/SYCL-CTS/pull/200/commits/b5c4eac065218bb997972bfe3914eb3c8bc3305a#diff-1eaa5e112ff44918530cc1ca6a8df13d6fb5379bf7045aaf9a2e8ec2e5262531L91) and [here](https://github.com/KhronosGroup/SYCL-CTS/pull/200/commits/b5c4eac065218bb997972bfe3914eb3c8bc3305a#diff-1eaa5e112ff44918530cc1ca6a8df13d6fb5379bf7045aaf9a2e8ec2e5262531L117-L121)

tests/sub_group_mask/sub_group_mask_all.cpp
bader: According to my understanding this is not "an Intel extension", but rather `oneAPI` extension, which is a cross-vendor extension targeting future core specification versions. What would be the right process to add tests for such extensions?

keryell: Let's discuss this during the next SYCL committee.

tests/sub_group_mask/sub_group_mask_common.h
vasilytric: @bader this fix (commit: eb0ad1d745453a6be2666cc4e162a173debfb675) is ok? can I apply this change to another tests on `sub_group_mask`?

bader: @vasilytric, it doesn't really answers my question. I see that now test tries sub-group sizes 8, 16 and 32, so having I think I get why we set `localSize` to 32, but it's not clear why global size should be larger than the `localSize`.

vasilytric: Yes, you are right we can have `global_size` the same as `local_size` and in this case we are going to have fewer parallel processes
This fix could be found in this commit: 96661372cff541955c3fa44fdaa97b535636de97

CMakeLists.txt
```diff
@@ -39,6 +39,15 @@ else()
 endif()
 # ------------------
 
+# ------------------
+# Extensive mode for running extension one API tests
```
bader, line 43:
```suggestion
# Extensive mode for running extension oneAPI tests
```
It's "oneAPI" in one word, so please, update it everywhere including name directory.
https://www.oneapi.io/

CMakeLists.txt
```diff
@@ -39,6 +39,15 @@ else()
...
+option(SYCL_CTS_ENABLE_EXTENSION_ONE_API_TESTS
+       "Enable extension one API tests" OFF)
```
bader, line 45: I suppose there will be more than one "oneAPI" extension and some implementation might want to validate some, but not all of them. We probably should think about more flexible way to configure tests for extensions (e.g. re-use filters enabled for core feature tests).

In addition to that, it would be great to build a common infrastructure for adding tests for other vendor extensions: `intel`, `codeplay`, etc.

CMakeLists.txt
```diff
@@ -39,6 +39,15 @@ else()
...
+if(SYCL_CTS_ENABLE_EXTENSION_ONE_API_TESTS)
+    message(WARNING "Enable extension tests mode: ON")
```
bader, line 47:
```suggestion
    message(STATUS "oneAPI extension tests mode: ON")
```

tests/extension_one_api/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+if(SYCL_CTS_ENABLE_EXTENSION_ONE_API_TESTS)
```
bader, line 1: How about following directory structure: 
`tests/extension/<vendor>/<short_ext_name>/...`, where short_ext_name is the name w/o `ext_<vendorstring>_` [prefix](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_names_for_extensions_to_existing_classes_or_enumerations).
(E.g. `tests/extension/oneapi/sub_group_mask/...`)

Another option:
 or `tests/extension/<ext_name>/...`, where `ext_name` can be full extension name as defined by [the spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_names_for_extensions_to_existing_classes_or_enumerations), but w/o `ext_` prefix - it's redundant for the tests located in extension directory.
(E.g. `tests/extension/oneapi_sub_group_mask/...`)

bader (changes requested):

CMakeLists.txt
vasilytric: > In addition to that, it would be great to build a common infrastructure for adding tests for other vendor extensions: intel, codeplay, etc.

@bader could you clarify what this infrastructure should do?
it should be a common flag to enable to make possible to build all extensions tests from all vendors or something different?

bader: Give developers control over enabled tests. This switch provides just two options: enable or disable all oneAPI extension tests at once.
I think we should have an interface to enable or disable each particular extension test.

vasilytric: > I think we should have an interface to enable or disable each particular extension test.

@bader ok, `SYCL_CTS_ENABLE_EXTENSION_ONE_API_TESTS` will be switched into separate flag for each extension group

CMakeLists.txt
```diff
@@ -39,6 +39,15 @@ else()
 endif()
 # ------------------
 
+# ------------------
+# Extensive mode for running extension oneAPI sub_group_mask tests
+option(SYCL_EXT_ONE_API_ENABLE_SUB_GROUP_MASK_TESTS
```
bader:
```suggestion
option(SYCL_ENABLE_EXT_ONEAPI_SUB_GROUP_MASK_TESTS
```
- ONE_API -> ONEAPI 
- Don't put ENABLE in the middle of the extension name.

CMakeLists.txt
vasilytric: this comment was applied in this commit: ca0b61742388852edfb295a4733924e19be9733c

CMakeLists.txt
vasilytric: the flag `SYCL_CTS_ENABLE_EXTENSION_ONE_API_TESTS` was replaced with `SYCL_EXT_ONEAPI_ENABLE_SUB_GROUP_MASK_TESTS` and new flag related to the tests on `sub_group_mask`
I've done this change [here](https://github.com/KhronosGroup/SYCL-CTS/pull/200/commits/e2188e3ff5458af4ef8a8f6bb2d9447088bc8846)

CMakeLists.txt
vasilytric: this comment was applied in this commit: ca0b61742388852edfb295a4733924e19be9733c

tests/extension_one_api/CMakeLists.txt
vasilytric: I think that the second option is better because we can see the vendor name in the prefix of the test group
I've applied this comment in this commit: ecc42021e507810dad1957f02176e380dfdea475

CMakeLists.txt
vasilytric: This comment was replaced in this commit: a294d1910b53c2fb50a3ded0e4aa10907dcd5cf2

vasilytric:
@keryell @ProGTX Do you have comment to these changes for moving and disabling by default tests on extensions
These changes can be found [here](https://github.com/KhronosGroup/SYCL-CTS/pull/200/files/6b186fce1d4f3f9dde8a983a9170b290d07f23f5..b89a27ee59600964bd6f4527a543b858c2ad82d5)

bader (approved):

keryell (approved):
Great!
I think having all the extensions in some subdirectories is quite clearer for feature control.

ProGTX (approved):

bader:
@psalz, do you know why required checks haven't been started?
GitHub UI requires all required checks to pass before unlocking merge button.

psalz:
> @psalz, do you know why required checks haven't been started? GitHub UI requires all required checks to pass before unlocking merge button.

Yes, the problem is that GitHub only lets you set required checks based on their name. So a run of `compile-cts (computecpp, 2.6.0)` will no longer satisfy the check if we update the ComputeCpp version to 2.8.0 (which is what happened here, same for the other implementations). Ideally we could say we require `compile-cts (computecpp, *)` to pass, however it looks like GitHub doesn't support any kind of wildcard syntax. So the current solution is either a back-merge to trigger a new run with the updated workflows, or we could simply get rid of that setting.

PR #203, opened 2021-10-18
Remove member function get() that are removed in SYCL 2020

kopylovanat:
In classes context, device, event, platform, queue

### DIFF
--- a/tests/common/common.h
+++ b/tests/common/common.h
@@ -167,7 +167,7 @@ void check_equality(sycl_cts::util::logger& log, T& a, T& b) {
   if (queue.get_backend() == sycl::backend::opencl) {
     if (sycl::get_native<sycl::backend::opencl>(a) !=
         sycl::get_native<sycl::backend::opencl>(b)) {
-      FAIL(log, "two objects are not equal (get_native)");
+      FAIL(log, "two objects are not equal");
     }
   }
 #endif  // SYCL_BACKEND_OPENCL
--- a/tests/common/common.h
+++ b/tests/common/common.h
@@ -160,13 +160,19 @@ void check_equality(sycl_cts::util::logger& log, T& a, T& b, bool checkGet) {
     FAIL(log, "two objects are not equal (is_host)");
   }
 
-  /** check get
+#ifdef SYCL_BACKEND_OPENCL
+  /** check get_native
    */
   if (checkGet) {
-    if (a.get() != b.get()) {
-      FAIL(log, "two objects are not equal (get)");
+    auto queue = util::get_cts_object::queue();
+    if (queue.get_backend() == sycl::backend::opencl) {
+      if ((sycl::get_native<sycl::backend::opencl>(a) !=
+          sycl::get_native<sycl::backend::opencl>(b))) {
+        FAIL(log, "two objects are not equal (get_native)");
+      }
     }
   }
+#endif  // SYCL_BACKEND_OPENCL
 };
 
 /**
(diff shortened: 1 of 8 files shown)


tests/context/context_constructors.cpp
```diff
@@ -128,9 +128,13 @@ class TEST_NAME : public util::test_base {
...
 
-#ifdef SYCL_CTS_TEST_OPENCL_INTEROP
-        if (!selector.is_host() && (contextA.get() != contextB.get())) {
-          FAIL(log, "context was not copied correctly (get)");
+#ifdef SYCL_BACKEND_OPENCL
+        auto queue = util::get_cts_object::queue();
+        if (queue.get_backend() == sycl::backend::opencl) {
+          if (sycl::get_native<sycl::backend::opencl>(contextA) !=
+              sycl::get_native<sycl::backend::opencl>(contextB)) {
+            FAIL(log, "context was not copied correctly (get)");
```
bader, line 136:
```suggestion
            FAIL(log, "context was not copied correctly");
```
What is the point of `(get)` (or `(get_native)`) part of the error message? There is no such part in tests/event/event_constructors.cpp.
I suggest we apply the same style to all similar messages and remove ` (get)` part of the message.

keryell:
It looks like you have removed a lot of `.get()` tests but they have been replaced by `get_native_mem()`, `get_native_queue()`, `get_native_device()`, `get_native_context()`...
I have the feeling this should be tested somewhere.
In a specific OpenCL-related specific tests?

tests/common/common.h
```diff
@@ -153,20 +153,24 @@ void check_get_profiling_info_param(sycl_cts::util::logger& log,
...
   }
 
-  /** check get
+#ifdef SYCL_BACKEND_OPENCL
+  /** check get_native
    */
-  if (checkGet) {
-    if (a.get() != b.get()) {
-      FAIL(log, "two objects are not equal (get)");
+  auto queue = sycl_cts::util::get_cts_object::queue();
```
ProGTX, line 166: I suggest moving the preprocessor check after the object, e.g.

```suggestion
  auto queue = sycl_cts::util::get_cts_object::queue();
  auto queueBackend = queue.get_backend();
#ifdef SYCL_BACKEND_OPENCL
```

The same elsewhere

kopylovanat: But if we move preprocessor check after the object we are going to get unused variables `queue` and `queueBackend` when SYCL_BACKEND_OPENCL is not defined, how code would benefit from such change?

kopylovanat:
> It looks like you have removed a lot of `.get()` tests but they have been replaced by some `get_native<>()` and it is unclear that all the cases are tested. For the buffers/images and events, it looks like the specification says some `std::vector` of `cl_mem` and `cl_event` are returned but it is unclear that these vectors are always ordered in the same way. Should we sort the vectors before comparing them? More generally, do you think everything is tested here or do you plan to have other OpenCL-related specific interoperability tests?

@keryell I'm removing checks for `get()` and not replacing them with `get_native` in these tests because `get_native` in SYCL 2020 unlike `get` in SYCL 1.2.1 is not member function of these classes. Checks for `get_native` are in opencl_interop test folder. They are also modified from checks for `get` that existed before for classes context, device, event, platform, queue, kernel. But since there was no checks for `get` for buffer or for new classes sampled_image and unsampled_image, checks for get_native for these classes need to be added separately. There are no immediate plans for it.
Check for get_native<event> fixed in PR https://github.com/KhronosGroup/SYCL-CTS/pull/210/ but it doesn't check for `cl_event` order. I'm not sure if it should be added since specification seems to be unclear about that.

keryell (approved):

psalz:
It seems like this PR, in particular ab7202a broke compilation with ComputeCpp, which is why now all new PRs are failing CI.

PR #205, opened 2021-10-21
Remove closing brace

vasilytric:


ProGTX (approved):

psalz:
Good catch! I overlooked this because no implementation supported compiling tests for specialization constants at the time.

PR #206, opened 2021-10-22
Fix double definition and update function comments for reduction_without_identity tests

vasilytric:


tests/reduction/reduction_without_identity_param_common.h
```diff
@@ -318,40 +316,45 @@ struct run_tests_for_all_functors {
     // for functors that can be called by .combine() or overloaded operator()
     // test will be called twice using operator +, *, ^= e.t.c.  and .combine()
     run_test_for_all_reductions_types<
-        VariableT, reduction_get_lambda::with_combine, UsePropertyFlagT>(
+        VariableT, reduction_get_lambda::with_combine, UsePropertyFlagT::value>(
```
keryell, line 319: By looking at all these repetitions, you should compute these once and just use them later with
```
using reduction = reduction_get_lambda::with_combine;
auto constexpr use_property = UsePropertyFlagT::value;
```
or whatever with a better naming.

keryell (commented):
Good simplification and comment improvement.

ProGTX (approved):

tests/reduction/reduction_without_identity_param_common.h
vasilytric: I applied this comment in this commit: d2ba612be8ce0df974466c9db04151402a0c3c13
I think that `use_lambda_with_combine`, `use_lambda_without_combine` and `use_property_flag` it's not bad names

keryell (approved):

PR #209, opened 2021-10-27
Add possible to see which tests were skipped and which tests were passed

vasilytric:
Add message to the console with the information about number skipped tests, the message will look like:
```
--- some_test
  . **Check #1 skipped**
  - skip
  . **Check #2 passed**

  . Passed 0/1 tests (0%)
  . Skipped 1/1 tests (100%)
```

The test will be marked as `Skipped` after `log.skip()` called even if some other check in the same test passed


psalz (approved):
> Before using some other test framework we can use this improvement

Yes, sure!

keryell (approved):

ProGTX (approved):

PR #210, opened 2021-10-28
Fix tests for opencl_interop

kopylovanat:


util/test_base_opencl.cpp
```diff
@@ -38,6 +38,12 @@ test_base_opencl::test_base_opencl()
  */
 bool test_base_opencl::setup(logger &log) {
   /* get the OpenCLHelper object */
+  auto queue = util::get_cts_object::queue();
+  if (queue.get_backend() != sycl::backend::opencl) {
+    log.skip("Interop part is not supported on non-OpenCL backend types");
```
keryell, line 43:
```suggestion
    log.skip("OpenCL interoperability part is not supported on non-OpenCL backend types");
```

ProGTX (approved):

util/test_base_opencl.cpp
kopylovanat: Done in https://github.com/KhronosGroup/SYCL-CTS/pull/210/commits/4dc8e5a93577b954c110130eea389bb3c87ba005

PR #211, opened 2021-11-01
Fix spurious failure for `specialization_constants_external` tests

yuriykoch:
It's implementation-defined if a kernel is available in a bundle with `bundle_state::input`.
So if such bundle misses some kernels it shouldn't trigger a test failure.


keryell:
The description of this PR should appear as a comment in this PR so that the code is understandable (around line 139?).

yuriykoch:
> The description of this PR should appear as a comment in this PR so that the code is understandable (around line 139?).

Done by 26372269066864e73e2db9840f33f4c86cf2bb27

ProGTX (approved):

PR #216, opened 2021-11-10
Fix conformance script arguments parsing

yuriykoch:
Looks like we need to escape `additional_args` , so we can pass a path within `additional_args` on Windows systems without backslashes removed by `shlex.split()` call.

### DIFF
--- a/run_conformance_tests.py
+++ b/run_conformance_tests.py
@@ -92,7 +92,8 @@ def split_additional_args(additional_args):
     # shlex doesn't support None
     if additional_args is None:
         return []
-    return shlex.split(additional_args)
+    # shell-parse in posix mode the shell-escaped string with additional_args
+    return shlex.split(shlex.quote(additional_args))
 
 
 def generate_cmake_call(cmake_exe, build_system_name, full_conformance,
--- a/run_conformance_tests.py
+++ b/run_conformance_tests.py
@@ -92,8 +92,15 @@ def split_additional_args(additional_args):
     # shlex doesn't support None
     if additional_args is None:
         return []
-    # shell-parse in posix mode the shell-escaped string with additional_args
-    return shlex.split(shlex.quote(additional_args))
+    use_posix_mode = os.name != 'nt'
+    # shlex was not intended for Windows by design, with non-POSIX Unix shells
+    # supported. Still it provides a Windows-compilant solution up to the
+    # certain degree, details: https://bugs.python.org/issue1724822
+    # The rules for quoting and escaping in non-POSIX mode are different from
+    # the native Windows rules, defined for CommandLineToArgvW for example.
+    # Still it makes possible to use something like "%ENV_VAR%\subdir" within
+    # additional arguments
+    return shlex.split(additional_args, posix=use_posix_mode)
 
 
 def generate_cmake_call(cmake_exe, build_system_name, full_conformance,


psalz:
Hm, I just removed this exact line because it breaks passing multiple additional arguments, for example `--additional-cmake-args="-DSYCL_IMPLEMENTATION=hipSYCL -DCMAKE_PREFIX_PATH=/path/to/hipsycl/bin"`.

I did some experimenting with the different variants:
```python
#!/usr/bin/env python3
import shlex
import sys

args = sys.argv[1]
default = shlex.split(args)
quote = shlex.split(shlex.quote(args))
nonposix = shlex.split(args, posix=False)

print(f"default {len(default)}:\t{default}")
print(f"quote {len(quote)}:\t{quote}")
print(f"posix=False {len(nonposix)}:\t{nonposix}")

```
Output on Linux:
```
> python3 shlex_test.py "ARG=foo UNIX_PATH=/my/path WIN_PATH=C:\path\to QARG='foo' QUNIX_PATH='/my/path' QWIN_PATH='C:\path\to'"
default 6:      ['ARG=foo', 'UNIX_PATH=/my/path', 'WIN_PATH=C:pathto', 'QARG=foo', 'QUNIX_PATH=/my/path', 'QWIN_PATH=C:\\path\\to']
quote 1:        ["ARG=foo UNIX_PATH=/my/path WIN_PATH=C:\\path\\to QARG='foo' QUNIX_PATH='/my/path' QWIN_PATH='C:\\path\\to'"]
posix=False 6:  ['ARG=foo', 'UNIX_PATH=/my/path', 'WIN_PATH=C:\\path\\to', "QARG='foo'", "QUNIX_PATH='/my/path'", "QWIN_PATH='C:\\path\\to'"]
```
Output on Windows:
```
python.exe shlex_test.py "ARG=foo UNIX_PATH=/my/path WIN_PATH=C:\path\to QARG='foo' QUNIX_PATH='/my/path' QWIN_PATH='C:\path\to'"
default 6:      ['ARG=foo', 'UNIX_PATH=/my/path', 'WIN_PATH=C:pathto', 'QARG=foo', 'QUNIX_PATH=/my/path', 'QWIN_PATH=C:\\path\\to']
quote 1:        ["ARG=foo UNIX_PATH=/my/path WIN_PATH=C:\\path\\to QARG='foo' QUNIX_PATH='/my/path' QWIN_PATH='C:\\path\\to'"]
posix=False 6:  ['ARG=foo', 'UNIX_PATH=/my/path', 'WIN_PATH=C:\\path\\to', "QARG='foo'", "QUNIX_PATH='/my/path'", "QWIN_PATH='C:\\path\\to'"]
```

Given this, I would say the best course of action would be to stick with the existing (non-quoted) approach, and use single quotes around Windows paths. Would that work for you?

run_conformance_tests.py
```diff
@@ -92,7 +92,8 @@ def split_additional_args(additional_args):
     # shlex doesn't support None
     if additional_args is None:
         return []
-    return shlex.split(additional_args)
+    # shell-parse in posix mode the shell-escaped string with additional_args
```
keryell, line 95:
```suggestion
    # shell-parse in POSIX mode the shell-escaped string with additional_args
```

bader:
> Given this, I would say the best course of action would be to stick with the existing (non-quoted) approach, and use single quotes around Windows paths. Would that work for you?
> 
> Edit: Actually I had formed my conclusion based on an earlier version of the test script where I've set the argument string directly inside the script instead of passing it in through the command line (which led to the `\t` inside the `C:\path\to` string to be interpreted as a tab character, causing different results). This is not a problem when passing the string through the command line (as we do in the actual `run_conformance_tests.py` script).
> 
> Given this, **it looks like `posix=False` would be the better option**, since it does not require quoting. If we decide to go with it, we should just check whether this could have any other unintended consequences.

@yuriykoch, please, reply to this question and apply the suggestion from @keryell.

run_conformance_tests.py
yuriykoch: Thanks, applied by a2a7efab82090aa48ef4fbe9b778b65c95bbe859

yuriykoch:
> Given this, **it looks like `posix=False` would be the better option**, since it does not require quoting. If we decide to go with it, we should just check whether this could have any other unintended consequences.

There were two major considerations originally for the Windows use-case
- path delimiters
- environment variables to be used as a part of `additional_args`

Using non-POSIX mode works fine for me, though I assume it should be enabled for Windows only. @psalz Do you have any concerns? Looks like we would use different additional arguments on Unix and on Windows systems anyway.

psalz (approved):
> @psalz Do you have any concerns?

If everything works for you now on Windows, I think that's a good solution!

keryell (approved):
Thanks for the motivating comment!

PR #220, opened 2021-11-29
Add tests for has_kernel_bundle for zero kernels and zero devices tests

vasilytric:
This PR provides these tests for `sycl::has_kernel_bundle`:
- Tests files with name `has_kernel_bundle_zero_devices` provides verifications for case when provide to the `sycl::has_kernel_bundle` empty vector with sycl::device
- Tests files with name `has_kernel_bundle_zero_kernels` provides verifications for case when provide to the `sycl::has_kernel_bundle` empty vector with sycl::kernel_id or application has no user-defined kernels

This PR also provides common code for kernel bundle tests and some other generic functions
___________________________________________

The `SYCL CTS CI / compile-cts (dpcpp)` failed due to `add_independent_cts_tests` CMake function is not defined in this PR, it can be found [here](https://github.com/KhronosGroup/SYCL-CTS/pull/218/files#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0dR197-R214)

PR #218, opened 2021-11-17
Tests for sycl::get_kernel_id and sycl::get_kernel_ids

vasilytric:
Some of these test required containing in separate *.exe-file: `get_kernel_ids_multiple_kernels`, `get_kernel_ids_single_kernel` and `get_kernel_ids_no_kernels` because these files required stricted list of available kernels
for build this tests need call building process for each test separately: `make test_get_kernel_ids_multiple_kernels`, `make test_get_kernel_ids_single_kernel` and `make test_get_kernel_ids_no_kernels`
But if use `make` all tests will be builded

PR #226, opened 2021-12-01
Add tests for sycl::compile

vasilytric:
The functions from `kernel_bundle.h`-file should extend the [file with the same name](https://github.com/KhronosGroup/SYCL-CTS/pull/220/files#diff-4284e4cdc3ce2098c0a3f9ed3a535f3e40b93332227a36d4ea252d65e574b9b5), that mentioned in PR#220

These tests do these verifications:
- if compilation is possible all kernels without optional features should be represented in the result bundle
- if compilation is possible only compatible kernels should be represented in the result bundle
- exception was thrown for zero devices passed
- associated devices are equal to devices from obtained kernel bundle
- `property_list` existence and empty `property_list` can be passed or omitted 

______________________________________
This PR depend on these PRs:
- https://github.com/KhronosGroup/SYCL-CTS/pull/218
- https://github.com/KhronosGroup/SYCL-CTS/pull/220


ProGTX (approved):

tests/kernel_bundle/sycl_compile.h
```diff
@@ -0,0 +1,165 @@
...
+                  " contains it");
+  }
+
+  // Check that result object bundle has the same kernels as input bundle
+  bool same_kernels_in_obj_bundle = true;
+  for (const auto &in_id : input_ids) {
+    same_kernels_in_obj_bundle &= obj_kb.has_kernel(in_id);
+  }
+
+  if (!same_kernels_in_obj_bundle) {
```
keryell, line 98: Use an algorithm

tests/kernel_bundle/sycl_compile.h
```diff
@@ -0,0 +1,165 @@
...
+  auto obj_kb = compile_bundle<Overload>(input_kb, devices);
+
+  // Check that result kernel bundle contains all devices from passed vector
+  bool same_devs = true;
+  auto kb_devs = obj_kb.get_devices();
+  for (const auto &dev : devices) {
+    same_devs &=
+        (std::find(kb_devs.begin(), kb_devs.end(), dev) != kb_devs.end());
+  }
+  if (!same_devs) {
```
keryell, line 144: Use an algorithm

tests/kernel_bundle/sycl_compile.h
```diff
@@ -0,0 +1,165 @@
...
+         "Set of associated to obj_kb devices is not equal to list of "
+         "devices passed.");
+  }
+
+  // Check that result kernel bundle does not have duplicates
+  bool unique_devs = true;
+  for (const auto &dev : kb_devs) {
+    unique_devs &= (std::count(kb_devs.begin(), kb_devs.end(), dev) == 1);
+  }
+  if (!unique_devs) {
```
keryell, line 155: Use an algorithm

keryell (changes requested):
Good!
But you could make the code clearer with algorithms like `any_of` , `all_of`...

tests/kernel_bundle/sycl_compile.h
vasilytric: this comment was applied [here](https://github.com/KhronosGroup/SYCL-CTS/pull/226/commits/649690bea56daeb866eee43cdad87e5f196de7c0#diff-d7dd04b2c33a3f23629a9f87a1e1089058ce7e6a7fa093db890c2d2508b4d206L140-L144)

tests/kernel_bundle/sycl_compile.h
vasilytric: this comment was applied [here](https://github.com/KhronosGroup/SYCL-CTS/pull/226/commits/649690bea56daeb866eee43cdad87e5f196de7c0#diff-d7dd04b2c33a3f23629a9f87a1e1089058ce7e6a7fa093db890c2d2508b4d206L93-L97)

tests/kernel_bundle/sycl_compile.h
```diff
@@ -0,0 +1,161 @@
...
+                  " but compiled bundle"
+                  " contains it");
+  }
+
+  // Check that result object bundle has the same kernels as input bundle
+  bool same_kernels_in_obj_bundle =
+      std::any_of(input_ids.cbegin(), input_ids.cend(),
+                  [&](auto &in_id) { return obj_kb.has_kernel(in_id); });
+
+  if (!same_kernels_in_obj_bundle) {
```
keryell, line 97: The interest of the algorithm is that you do not need the variable to compute the intermediate result anymore
```suggestion
  if (!std::any_of(input_ids.cbegin(), input_ids.cend(),
                   [&](auto &in_id) { return obj_kb.has_kernel(in_id); })) {
```

tests/kernel_bundle/sycl_compile.h
```diff
@@ -0,0 +1,161 @@
...
+
+  auto input_kb =
+      sycl::get_kernel_bundle<sycl::bundle_state::input>(ctx, devices);
+  auto obj_kb = compile_bundle<Overload>(input_kb, devices);
+
+  // Check that result kernel bundle contains all devices from passed vector
+  bool same_devs = true;
+  auto kb_devs = obj_kb.get_devices();
+  same_devs = std::equal(kb_devs.begin(), kb_devs.end(), devices.begin());
+  if (!same_devs) {
```
keryell, line 140:
```suggestion
  auto kb_devs = obj_kb.get_devices();
  if (!same_devs = std::equal(kb_devs.begin(), kb_devs.end(), devices.begin())) {
```
and in C++20 it will be just 1 line :-)

tests/kernel_bundle/sycl_compile.h
keryell: Yes, this one is subtler.
Right now you have an O(n^2) implementation.
By looking at https://en.cppreference.com/w/cpp/algorithm, what about:
```c++
std::sort(kb_devs.begin(), kb_devs.end());
if (std::adjacent_find(kb_devs.begin(), kb_devs.end()) != kb_devs.end()) {
```
in O(n log n)?
Perhaps we could have the utility function to return always a sorted list at the first place?
@illuhad interesting to see all this code requiring comparison of devices. :-)

tests/kernel_bundle/sycl_compile_exception_for_zero_devices.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    if ( !sycl::has_kernel_bundle<sycl::bundle_state::input>(queue.get_context()) ) {
+      log.note("No kernel bundle with input state for test (skipped).");
+      return;
+    }
+
+    auto input_kb = sycl::get_kernel_bundle<sycl::bundle_state::input>(queue.get_context());
+
+    TestCaseDescription desc{"(kernel_bundle, {})"};
+    expect_throws<sycl::errc::invalid>(log, desc, [&]() {
```
keryell, line 43:
```suggestion
    expect_throws<sycl::errc::invalid>(log, desc, [&] {
```

tests/kernel_bundle/sycl_compile_exception_if_dev_not_in_set.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    if ( !sycl::has_kernel_bundle<sycl::bundle_state::input>(ctx) ) {
+      log.note("No kernel bundle with input state for test (skipped).");
+      return;
+    }
+
+    auto input_kb = sycl::get_kernel_bundle<sycl::bundle_state::input>(ctx);
+
+    TestCaseDescription desc{"(kernel_bundle, devices)"};
+    expect_throws<sycl::errc::invalid>(log, desc, [&]() {
```
keryell, line 49:
```suggestion
    expect_throws<sycl::errc::invalid>(log, desc, [&] {
```

keryell (changes requested):
Almost there!

vasilytric:
> Perhaps we could have the utility function to return always a sorted list at the first place?
@keryell In case when we have `std::vector<sycl::device>` we can't use `std::adjacent_find` because of I get the following error, while I try to call `std::sort(kb_devs.begin(), kb_devs.end())`: "error: no matching function for call to object of type 'std::less<>'"

tests/kernel_bundle/sycl_compile_exception_if_dev_not_in_set.cpp
vasilytric: I've applied this comment in this commit: 6fdf3ec404994d18aa35613bcf7165c9a2d59d4d

tests/kernel_bundle/sycl_compile.h
vasilytric: I've applied this comment [here](https://github.com/KhronosGroup/SYCL-CTS/pull/226/commits/269dcf45c5db46690f9568a15ec437558e76df3c#diff-d7dd04b2c33a3f23629a9f87a1e1089058ce7e6a7fa093db890c2d2508b4d206R134-R136)

tests/kernel_bundle/sycl_compile.h
vasilytric: I've applied this comment [here](https://github.com/KhronosGroup/SYCL-CTS/pull/226/commits/3b85821013e59bdb8af8b5f011b2105b599fd931#diff-d7dd04b2c33a3f23629a9f87a1e1089058ce7e6a7fa093db890c2d2508b4d206L97-L98)

tests/kernel_bundle/sycl_compile.h
keryell: This is either a bug in the implementation or a bug in the SYCL spec.
It should be possible to have a `std::set` or even a `std::unsorted_set` of `sycl::device`.

vasilytric: > It should be possible to have a `std::set` or even a `std::unsorted_set` of `sycl::device`.
@keryell We can use `std::unsorted_set<sycl::device>` without some errors, and we can use the `std::set<sycl::device>` with a comparing function.

But in this case we can't use `std::sort` for sorting by `std::hash` because of the fact that result of the `std::hash` is unexpected: two different objects that can have the same hash (according to C++17)

Maybe you have a suggestion?

keryell: That is the normal behavior of a hash function for not being injective https://en.wikipedia.org/wiki/Injective_function since the goal is to compress the definition domain. The hashing is not use to do the comparison, just as a heuristic to improve the data structure. At the end an equality comparison is always done.
But the real problem is that a SYCL implementations should have comparison operations working on most SYCL objects like `device`, `buffer`... and a `std::hash` specialization working.
Do we have such tests in the CTS at the first place?
Then, not using `std::sort`, `std::set` or `std::unordered_set` because of SYCL implementation bugs is futile since it is not conformant anyway. :-)
@bader My feeling is that we should implement each test the right way™ even if it does not compile today on some implementations because of some implementation limitations.

bader: Interesting...
How do see comparison of two sycl devices can be implemented? DPC++ implements SYCL device as a wrapper around a native handle. We could implement SYCL objects comparison operators if native handles can be compared, but I'm not sure if it's defined.
AFAIK, OpenCL device handle is just a pointer, so comparing pointer values might be unstable and we get different results sorting the same list of devices. Right?

keryell: I have created internal ticket https://gitlab.khronos.org/sycl/Specification/-/issues/586 and we can discuss it during the next SYCL committee meeting.

keryell: OK, from the SYCL workgroup meeting today this might be fixed in the specification.
In the meantime you can keep the O(n^2) algorithm and we can come back on this.
The problem is to remember where we will have to fix it. Adding some kind of TODO or adding a link to each test in https://gitlab.khronos.org/sycl/Specification/-/issues/586 ?

keryell (approved):
Thanks.

tests/kernel_bundle/sycl_compile.h
vasilytric: @keryell  TODO comment was added in this commit: a3afd778782dff127cabb8f71fe4d4036f3ee5c5
Provided issue requires authorization to see it. Therefore the link on the issue was not mentioned because of not all users can see this issue.

tests/kernel_bundle/sycl_compile.h
```diff
@@ -0,0 +1,163 @@
...
+  if (!std::equal(kb_devs.begin(), kb_devs.end(), devices.begin())) {
+    FAIL(log,
+         "Set of associated to obj_kb devices is not equal to list of "
+         "devices passed.");
+  }
+
+  // TODO: we can't use std::set<sycl::device> due to the fact that sycl::device
+  // has no comparative operators.
+  // When this issue will be resolved the following statement can be used insted
+  // of cycle with std::count calling:
```
keryell, line 144:
```suggestion
  // has no comparison operators.
  // Once this issue is resolved, the following statement can be used instead
  // of cycling with a std::count call:
```

keryell (changes requested):
Thanks for the comment.
Almost there!

tests/kernel_bundle/sycl_compile.h
vasilytric: I've applied this comment in this commit: d910c659140ff7236d0fdc6281a733d02b2b9a3c

keryell (approved):
Thanks!

PR #229, opened 2021-12-01
Add tests on sycl::join

vasilytric:
These tests do verifications for these scenarios:
 - join kernel bundle with empty one
 - join single kernel bundle (once, twice)
 - join zero kernel bundles
 - check ability to invoke kernel after join


ProGTX (approved):

tests/kernel_bundle/sycl_join.h
```diff
@@ -0,0 +1,36 @@
...
+template <sycl::bundle_state State>
+inline bool check_dev_images_equal(sycl_cts::util::logger &log,
+                                   sycl::kernel_bundle<State> &lhs_kb,
+                                   sycl::kernel_bundle<State> &rhs_kb) {
+  bool ok = (std::distance(lhs_kb.begin(), lhs_kb.end()) ==
+             std::distance(rhs_kb.begin(), rhs_kb.end()));
+  if (!ok) return ok;
+  for (const auto &dev_img : lhs_kb) {
+    ok &= (std::find(rhs_kb.begin(), rhs_kb.end(), dev_img) != rhs_kb.end());
+  }
```
keryell, line 29: This is another case where we would need https://gitlab.khronos.org/sycl/Specification/-/issues/586 to avoid this O(n^2) algorithm.

tests/kernel_bundle/sycl_join_check_kernel_execution.cpp
```diff
@@ -0,0 +1,76 @@
...
+                      sycl::kernel_bundle<sycl::bundle_state::executable> &kb) {
+  bool value_to_rewrite_in_kernel = false;
+  {
+    sycl::buffer<bool, 1> res_buffer(&value_to_rewrite_in_kernel,
+                                     sycl::range<1>(1));
+    q.submit([&](sycl::handler &cgh) {
+      auto res_acc =
+          res_buffer.template get_access<sycl::access_mode::write>(cgh);
+      cgh.use_kernel_bundle(kb);
+      cgh.single_task<KernelName>([=]() { res_acc[0] = true; });
```
keryell, line 39:
```suggestion
      cgh.single_task<KernelName>([=] { res_acc[0] = true; });
```

keryell (changes requested):
Good.

tests/kernel_bundle/sycl_join.h
keryell: OK, from the SYCL workgroup meeting today this might be fixed in the specification.
In the meantime you can keep the O(n^2) algorithm and we can come back on this.
The problem is to remember where we will have to fix it. Adding some kind of TODO or adding a link to each test in https://gitlab.khronos.org/sycl/Specification/-/issues/586 ?

vasilytric: @keryell  I am going to leave the same comment as [here](https://github.com/KhronosGroup/SYCL-CTS/commit/a3afd778782dff127cabb8f71fe4d4036f3ee5c5#diff-d7dd04b2c33a3f23629a9f87a1e1089058ce7e6a7fa093db890c2d2508b4d206R141-R146)

If this comment structure is ok, then I'll apply it to this case

tests/kernel_bundle/sycl_join_check_kernel_execution.cpp
vasilytric: I've applied this comment [here](https://github.com/KhronosGroup/SYCL-CTS/pull/229/commits/05b37d37e6f10c1afd5ea287aaa95be0d52d9259#diff-baef877460d6a2ce4e330d5cd9c11f26a0d498776535ee294b1f30540ce86a89R39)

tests/kernel_bundle/sycl_join.h
vasilytric: I've applied this comment [here](https://github.com/KhronosGroup/SYCL-CTS/pull/229/commits/05b37d37e6f10c1afd5ea287aaa95be0d52d9259#diff-6b84827637b50d648c97ff5af6c69a319ec4de83ebccd0b2ebe5b18d2774dbf6R27-R32)

keryell: Good. Can you also retrofit from https://github.com/KhronosGroup/SYCL-CTS/pull/226/commits/d910c659140ff7236d0fdc6281a733d02b2b9a3c ?

vasilytric: @keryell Ronan, if I correctly understood you then, I have already done it.
Updated comment can be found in [tests/kernel_bundle/sycl_join.h](https://github.com/KhronosGroup/SYCL-CTS/pull/229/files#diff-6b84827637b50d648c97ff5af6c69a319ec4de83ebccd0b2ebe5b18d2774dbf6R27-R32)

keryell (approved):
Great!

PR #230, opened 2021-12-06
Avoid tabs in command lines

hvdijk:
This commit updates cmake/FindComputeCpp.cmake to avoid the use of tabs in command lines. The tabs are valid but cause issues when copying and pasting commands to run them standalone, as even the pasting of a tab may trigger tab completion, depending on the shell and shell settings.

The tabs were a workaround for older CMake versions, but a sufficiently recent CMake is already required.

I have verified that with this MR applied, with the Ninja generator, the only change to the generated `build.ninja` is the removal of the tabs.

Note: this PR is against SYCL-2020 but also applies cleanly to SYCL-1.2.1/master. If this gets approved and merged to SYCL-2020, I would appreciate it if it could also be cherry-picked there, or I can create a new PR for that.

### DIFF
--- a/cmake/FindComputeCpp.cmake
+++ b/cmake/FindComputeCpp.cmake
@@ -89,11 +89,12 @@ function(build_spir exe_name spir_target_name source_file output_path)
             ${computecpp_device_compiler_defs}
             ${COMPUTECPP_USER_FLAGS}
             ${platform_specific_args}
-            $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\t-I\">\">
-            $<$<BOOL:${device_compile_definitions}>:-D$<JOIN:${device_compile_definitions},\t-D>>
-            $<JOIN:${device_compile_options},\t>
+            $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\;-I\">\">
+            $<$<BOOL:${device_compile_definitions}>:-D$<JOIN:${device_compile_definitions},\;-D>>
+            $<JOIN:${device_compile_options},\;>
             -o ${output_bc}
             -c ${source_file}
+    COMMAND_EXPAND_LISTS
     DEPENDS ${source_file}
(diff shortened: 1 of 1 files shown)


bader (approved):

ProGTX (approved):

cmake/FindComputeCpp.cmake
```diff
@@ -89,11 +89,12 @@ function(build_spir exe_name spir_target_name source_file output_path)
             ${computecpp_device_compiler_defs}
             ${COMPUTECPP_USER_FLAGS}
             ${platform_specific_args}
-            $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\t-I\">\">
-            $<$<BOOL:${device_compile_definitions}>:-D$<JOIN:${device_compile_definitions},\t-D>>
-            $<JOIN:${device_compile_options},\t>
+            $<$<BOOL:${include_directories}>:-I\"$<JOIN:${include_directories},\"\;-I\">\">
```
keryell, line 92: After being puzzled by the meaning of `\;` I looked at the CMake documentation. Ouch, it has really a meaning. :-)

keryell: Add a comment explaining how it works along https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#escape-sequences

ProGTX: Isn't this because `;` is used for lists in CMake, so you need to escape that? Do we need to document that? I guess it's a bit strange, but so is large part of CMake :)

keryell: Yes, that is fair.

keryell (approved):
Thanks!

bader:
Can someone with admin rights to merge this PR, please?
I can't merge because of missing status of required checks.

keryell:
> I can't merge because of missing status of required checks.

I have added you to the SYCL admin list anyway as you deserve it. :-)

PR #217, opened 2021-11-11
Migrate CTS to Catch2 Testing Framework

psalz:
As decided in the WG, we will be migrating the CTS away from the existing custom-built testing framework to [Catch2](https://github.com/catchorg/Catch2). This will allow us to write more conventional, fine-grained test cases with natural language descriptions, hopefully making it both easier to comprehend existing tests as well as write new ones.

Going forward, instead of looking like this...
```c++
class event_constructors : public util:test_base {
  void run(util::logger& log) override {
    // (... more tests)
    // check equality operator
    {
      cts_selector selector;
      auto queue = util::get_cts_object::queue(selector);

      auto eventA = get_queue_event<class event_constructors_6>(queue);
      cl::sycl::event eventB(eventA);
      auto eventC = get_queue_event<class event_constructors_7>(queue);
      eventC = eventA;
      auto eventD = get_queue_event<class event_constructors_8>(queue);

      if (!(eventA == eventB)) {
        FAIL(log, "event equality does not work correctly (copy constructed)");
      }
      if (!(eventA == eventC)) {
        FAIL(log, "event equality does not work correctly (copy assigned)");
      }
      // (... more checks)
      queue.wait_and_throw();
    }
    // (... more tests)
  }
}
```

...a test case will look more like this:
```c++
TEST_CASE("sycl::event supports equality operators") {
  cts_selector selector;
  auto queue = util::get_cts_object::queue(selector);

  auto eventA = get_queue_event<class event_constructors_6>(queue);
  cl::sycl::event eventB(eventA);
  auto eventC = get_queue_event<class event_constructors_7>(queue);
  eventC = eventA;
  auto eventD = get_queue_event<class event_constructors_8>(queue);

  CHECK(eventA == eventB);
  CHECK(eventA == eventC);
  CHECK_FALSE(eventA != eventB);
  CHECK_FALSE(eventA != eventC);
  CHECK_FALSE(eventC == eventD);
  CHECK(eventC != eventD);

  queue.wait_and_throw();
}
```

Other benefits that come with Catch2 include:
- Smart assertion macros with great error reporting
- `SECTION`s to elegantly express different combinations of testing logic that share common setup and/or teardown code
- Type parametrized (templated) test cases
- ...and more. See the [Catch2 documentation](https://github.com/catchorg/Catch2/tree/devel/docs) for a full overview of all available features.

## Migration Strategy

The CTS is already quite large (> 50k LoC) and re-writing all tests from scratch at once is infeasible; an incremental strategy is needed.

[...]

PR #231, opened 2021-12-07
Mark existing `multi_ptr` tests as legacy

yuriykoch:
Make existing `multi_ptr` tests legacy according to the SYCL-2020 updates:

- enforce legacy decoration usage
- rename existing files and test names
- add SYCL_CTS_ENABLE_LEGACY_TESTS flag
- and update the run_conformance_tests.py appropriately, to make it possible to disable legacy tests


psalz:
In #217 I'm referring to tests that use the current (soon to be old) way of writing tests as "legacy". To avoid confusion, should we simply call these something like "1.2.1-tests" instead?

psalz:
After looking some more into this; I originally thought you were guarding tests for SYCL 1.2.1 functionality that has been removed/made optional in 2020. However it is my understanding that `access::decorated::legacy` is a non-optional part of the 2020 specification, i.e. it is required for conformance, so why should there be an option for the CTS to disable testing for this?

yuriykoch:
> However it is my understanding that `access::decorated::legacy` is a non-optional part of the 2020 specification, i.e. it is required for conformance, so why should there be an option for the CTS to disable testing for this?

This indeed is a non-optional, but deprecated part of the SYCL-2020 standard. Therefore we might use it, for example, for deprecated accessor API. The option is enabled by default, so it doesn't affect the conformance. Still it provides a possibility to have tests for deprecated features disabled - and to state explicitly the tests are on the deprecated feature which may be gone someday.
May be we need to rename the `legacy` option/macro to the `deprecated` within this PR?

bader (approved):

run_conformance_tests.py
```diff
@@ -59,6 +59,11 @@ def handle_args(argv):
         help='Disable full conformance mode to avoid extensive tests.',
         required=False,
         action='store_true')
+    parser.add_argument(
+        '--enable-legacy',
+        help='Enable legacy tests.',
+        required=False,
+        action='store_true')
```
psalz, line 66: Shouldn't this also default to true for the conformance report generation?

psalz (approved):
Sorry for the late response. I agree that having a switch for disabling tests for deprecated features could be useful, as long as they are enabled by default. I'd still be in favor of renaming the option to something like `SYCL_CTS_TEST_DEPRECATED_FEATURES` (or `SYCL_CTS_ENABLE_DEPRECATED_TESTS`, although that sounds like the tests themselves are deprecated, which they are not) to avoid confusion with "legacy-style tests" once the Catch2 migration has been merged.

run_conformance_tests.py
yuriykoch: My bad. Fixed both the default value and argument naming by 97f90a65ce88610ee4af1a99d5ff183812b26ae0.

yuriykoch:
Thanks, the suggested `SYCL_CTS_TEST_DEPRECATED_FEATURES` looks good to me. Updated by bf9cdf6ba43771fb6802354cae2eacc6e37e554e

psalz (approved):
Thanks, LGTM!

PR #235, opened 2021-12-13
Fix accessor_api_image_core test for image arrays

psamolysov-intel:
When the acc_coord_tag::use_normalized_upper tag is used to compute
the expected values in the linear filtering mode, the value for idx + 1
is used as expected (see the return get_expected_value_linear(idx + 1)
line in the body of the T get_expected_value(image_id_t<dim, target> idx)
method of the image_accessor_api_sampled_r class). This fact is taken into
account in the image_access<dims>::get_normalized() function and works
fine for a single image but not for image arrays since for the image array,
a correct slice should be the next one when the linear filtering mode is
used and the test is run for acc_coord_tag::use_normalized_upper tag.

### DIFF
--- a/tests/accessor/accessor_api_image_common.h
+++ b/tests/accessor/accessor_api_image_common.h
@@ -390,30 +390,67 @@ std::vector<sycl::byte> get_image_input_data(
   return convert_image_data_to_bytes<T>(data, byteSize);
 }
 
+/**
+ * @brief Returns an image_array's index to read an image by
+ *        coordinates from
+ * @tparam dims Number of accessor dimensions
+ * @param idx Work-item ID
+ * @return Index of the image to read
+ */
+template <int dims>
+size_t get_slice_index(image_array_id_t<dims> idx) {
+  return idx[dims];
+}
+
+/**
+ * @brief Returns an image_array's index to read an image by
+ *        coordinates from
+ * @tparam dims Number of accessor dimensions
+ * @tparam coordT Type of the tag to mark the right index for test
+ * @param coordTag A tag to mark the right index for test
+ * @param filteringMode A filtering mode used by the sampler to read the image
+ * @param idx Work-item ID
+ * @return Index of the image to read. If the linear filtering mode is used
+ *         and coordT is equal to acc_coord_tag::use_normalized_upper, the
+ *         function returns the next index.
+ */
+template <int dims, typename coordT>
+size_t get_slice_index(const coordT& coordTag,
+                       sycl::filtering_mode filteringMode,
+                       image_array_id_t<dims> idx) {
+  if (std::is_same_v<coordT, acc_coord_tag::use_normalized_upper> &&
+      filteringMode == sycl::filtering_mode::linear) {
+    return idx[dims] + 1;
+  } else {
+    return idx[dims];
+  }
+}
+
 template <typename T, int dims, sycl::target target,
-          sycl::access_mode mode>
-T read_image_acc(const sycl::accessor<T, dims, mode, target> &acc,
+          sycl::access_mode access_mode>
+T read_image_acc(const sycl::accessor<T, dims, access_mode, target> &acc,
                  sycl::id<dims> idx) {
   return acc.read(image_access<dims>::get_int(idx));
 }
 
-template <typename T, int dims, sycl::access_mode mode>
-T read_image_acc(const sycl::accessor<T, dims, mode,
+template <typename T, int dims, sycl::access_mode access_mode>
+T read_image_acc(const sycl::accessor<T, dims, access_mode,
                                     sycl::target::image_array> &acc,
                  image_array_id_t<dims> idx) {
   // Verify __image_array_slice__ read
   using coordT = acc_coord_tag::use_int;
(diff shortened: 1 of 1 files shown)


bader (approved):

tests/accessor/accessor_api_image_common.h
```diff
@@ -427,16 +463,18 @@ T read_image_acc_sampled(const sycl::accessor<T, dims, mode, target> &acc,
...
                          sycl::sampler smpl,
                          image_array_id_t<dims> idx,
                          image_array_range_t<dims> range,
-                         const coordT& coordTag) {
+                         const coordT& coordTag,
+                         sycl::filtering_mode filteringMode) {
   // Verify __image_array_slice__ read
   const auto coords = image_array_coords<dims>::get(coordTag, idx, range);
-  return acc[idx[dims]].read(coords, smpl);
+  return acc[get_slice_index<dims>(coordTag, filteringMode, idx)].read(coords,
```
yuriykoch, line 476: Currently the range is defined by using the `image_accessor_sampler::supports_out_of_range` method, so for the linear filtering  mode and image with texels `0 .. N-1` we have verification for `acc[0].read() ... acc[N].read()`.

Please correct me if I'm wrong, but looks like with this patch we are going change verification range from `0 .. N` to `1 .. N+1`, switching off the verification of `acc[0].read()` in case of linear filtration and `acc_coord_tag::use_normalized_upper`, but using two out-of-bound indexes instead.

Is it intended, or may be it's better to tune the `get_expected_value<coordT>(idx)` used on the [L1108](https://github.com/KhronosGroup/SYCL-CTS/pull/235/files#diff-cffc0211756d6bcabcad4ed222118165250a78dc0187548ab21285c84258f2a4R1108)?

psamolysov-intel: Hm... It looks like you are right, the test after the change just skips `acc[0].read()`. This is similar how `get_expected_value` returns `get_expected_value_linear(idx + 1)` in case of linear filtration and `acc_coord_tag::use_normalized_upper`. So, if we have `idx` is equal to `0`, we calculate an expected value for `1` and for `target::image` we compare this expected value with the actual value read from `get_normalized(const pixel_tag::upper, ...)` coordinates (the coordinates for `0+1` with some manipulations). For `target::image_array`, the last number in `idx` is the image index, while the others are used to calculate the coordinates to read from. When we did `idx + 1` in the `get_expected_value` method, we incremented all the components of `idx` and we should increment the image index we are reading from to match this increment.

I'm not sure what do you mean when you said about to tune the `get_expected_value<coordT>(idx)`. I think we should increment all the components of `idx` but the last one if we are working with `target::image_array`. I've tried and it works, the code has got fewer changes that I proposed before. Thank you for the suggestion. I committed a new patch, please have a look.

tests/accessor/accessor_api_image_common.h
```diff
@@ -984,10 +984,23 @@ class image_accessor_api_sampled_r {
        *    ftp://ftp.icsi.berkeley.edu/pub/theory/priest-thesis.ps.Z
        *  for details
        */
-      return get_expected_value_linear(idx + 1);
+      return get_expected_value_linear(next_idx(idx));
     }
   }
 
+  template <int dims>
+  sycl::id<dims> next_idx(sycl::id<dims> idx) const {
```
bader, line 992:
```suggestion
  sycl::id<dims> next_id(sycl::id<dims> idx) const {
```
I think it's worth adding a comment clarifying what this function does and why.

psamolysov-intel: Thank you for the suggestion, I've renamed the method and added a comment why this method is required.

tests/accessor/accessor_api_image_common.h
psamolysov-intel: @yuriykoch Unfortunately the "Re-request review" button doesn't work. Could you review the latest commit?

tests/accessor/accessor_api_image_common.h
```diff
@@ -984,10 +984,28 @@ class image_accessor_api_sampled_r {
...
+  template <int dims>
+  sycl::id<dims> next_id(sycl::id<dims> idx) const {
+    if constexpr (target == sycl::target::image_array) {
+      auto res = idx;
+      for (int i = 0; i + 1 < dims; i++) {
+        res[i]++;
+      }
+      return res;
+    }
+    return idx + 1;
```
yuriykoch, line 1006:
```suggestion
    static_assert(dims > 0);
    ++idx;
    if constexpr (target == sycl::target::image_array) {
      --idx[dims-1];
    }
    return idx;
```
Just as a suggestion. We are not going to wrap on `size_t` anyway, as it would require quite huge images. Assert is going to have no effect for SYCL-2020 rev.4

psamolysov-intel: @yuriykoch Thank you for the suggestion, I've committed it.

ProGTX (approved):

psamolysov-intel:
@illuhad @keryell Could you review this PR please?

bader:
Can someone with admin rights to merge this PR, please?
I can't merge because of missing status of required checks.

PR #236, opened 2021-12-13
Fix usm_atomic_access_* tests

psamolysov-intel:
In accordance to the SYCL 2020 specification
(https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references),
the sycl::atomic_ref class template has no default values for the
DefaultOrder and DefaultScope template parameters, a default value is
defined for the last one (Space) only. So, the tests have been changed
to explicitly provide those parameters. Also, the test attempted to
create an atomic_ref value with a pointer instead of a value.

The short and unsigned short data types has been removed from the
test as unsupported, the SYCL 2020 specification states that the data
type T in the sycl::atomic_ref first template parameter is permitted to
be int, unsigned int, long, unsigned long, long long, unsigned long long,
float or double (see 4.15.3. Atomic references).

### DIFF
--- a/.github/workflows/build_ci_containers.yml
+++ b/.github/workflows/build_ci_containers.yml
@@ -41,7 +41,7 @@ jobs:
           # - sycl-impl: computecpp
           # version: 2.6.0
           - sycl-impl: dpcpp
-            version: f5c838cd
+            version: 6699a5d
           - sycl-impl: hipsycl
             version: 72a29fb9
     steps:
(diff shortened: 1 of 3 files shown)


bader:
I recently was trying to use atomic_ref and found SYCL spec confusing. I agree with the part that "the sycl::atomic_ref class template has no default values", but there are a couple of examples ([here](the sycl::atomic_ref class template has no default values) and [here](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:kernel.attributes) in listing 3) using atomic_ref class w/o explicitly specifying template parameters.
@AerialMantis, was the intention to use user-defined deduction guides to infer default scope and order values or it's a bug in the spec examples?

bader (approved):
USM tests are not supposed to test `atomic_ref` API, so this change looks reasonable to me.

jbrodman (approved):
LGTM

PR #237, opened 2021-12-14
Improve generic code

AndreyXRomanov:
Move utility functions and user defined data types into `tests/common`.


tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
...
+struct no_cnstr {
+  float a;
+  int b;
+  char c;
+
+  void operator=(const int &v) {
+    this->a = v;
+    this->b = v;
+    this->c = v;
+  }
```
keryell, line 48: Does this compiles without returning a reference to `*this`?
```suggestion
  auto& operator=(const int &v) {
    a = v;
    b = v;
    c = v;
    return *this;
  }
```
Use case:
```c++
no_cnstr a, b;
a = b = 2;
```

tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
...
+    a = val * 3.0;
+    b = val * 2;
+    c = val;
+  }
+
+  void operator=(const int &v) {
+    this->a = v * 3.0;
+    this->b = v * 2;
+    this->c = v;
+  }
```
keryell, line 76: Cf supra

tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
...
+  friend bool operator==(const no_def_cnstr &lhs, const no_def_cnstr &rhs) {
+    return ((lhs.a == rhs.a) && (lhs.b == rhs.b) && (lhs.c == rhs.c));
+  }
+
+  void operator=(const int &v) {
+    no_def_cnstr temp(v);
+    this->a = temp.a;
+    this->b = temp.b;
+    this->c = temp.c;
+  }
```
keryell, line 104:
```suggestion
  auto& operator=(const int &v) {
    return *this = no_def_cnstr { v };
  }
```
but it might be too terse?

keryell: Actually you do not even need this at all since you have `constexpr no_def_cnstr(int val) : a(val * 3.0), b(val * 2), c(val) {}` which is implicit, so the copy constructor will be called afterward.
There might need some copy constructor to be defaulted to have this working if this is not already the case.

tests/common/type_list.h
keryell, line 76:
```suggestion
  auto& operator=(const int &v) {
    assign(v);
    return *this;
  }
```

tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
...
+inline constexpr auto get_init_value_helper<bool>(int x) {
+  return (x%2 != 0);
+}
+
+template <>
+inline constexpr auto get_init_value_helper<no_cnstr>(int x) {
+  no_cnstr instance{};
+  instance.a = x;
+  instance.b = x;
+  instance.c = x;
```
keryell, line 122:
```suggestion
  instance = x;
```

tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
...
+  return instance;
+}
+
+template <>
+inline constexpr auto get_init_value_helper<def_cnstr>(int x) {
+  def_cnstr instance;
+  instance.assign(x);
```
keryell, line 129:
```suggestion
  instance= x;
```

tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
   }
 };
 
+namespace user_def_types {
+// A user-defined struct with several scalar member variables, no constructor,
+//  destructor or member functions.
```
keryell, line 38:
```suggestion
//  destructor or non-special member functions.
```

tests/common/type_list.h
```diff
@@ -33,37 +33,137 @@ struct user_struct {
...
+  void operator=(const int &v) {
+    no_def_cnstr temp(v);
+    this->a = temp.a;
+    this->b = temp.b;
+    this->c = temp.c;
+  }
+};
+
+template <typename T>
+inline constexpr auto get_init_value_helper(int x) {
```
keryell, line 108: Comment, including the specializations.

keryell (changes requested):
Always good to refactor!
I have the feeling it can still be simplified.
And you can always to take the opportunity to add more comments, specially at the end.

AndreyXRomanov:
@keryell @ProGTX Do you have any additional comments?

tests/common/type_list.h
```diff
@@ -33,37 +33,131 @@ struct user_struct {
...
-    "sycl::cl_short",
-    "sycl::cl_ushort",
-    "sycl::cl_int",
-    "sycl::cl_uint",
-    "sycl::cl_long",
-    "sycl::cl_ulong"};
-} // namespace get_cts_type
-
-#endif // __SYCLCTS_TESTS_COMMON_TYPE_LIST_H
+static const auto vector_types =
```
ProGTX, line 130: Shouldn't these global variables be `inline`?

ProGTX (approved):

tests/common/type_list.h
AndreyXRomanov: I think it will be better if we use a function that returns `vector_types` variable.
I applied this change in 963504b390bbb87f3c01063d8f1c389c30597a59

bader (approved):

AndreyXRomanov:
@keryell Do you have extra comments on this PR?

fraggamuffin:
Need approval from RK

fraggamuffin:
needs back merge

keryell (approved):
Thanks!

AndreyXRomanov:
> needs back merge

Done

PR #238, opened 2021-12-17
Run accessor_api_image_core test with int coords for nearest only

psamolysov-intel:
The OpenCL specification rules that when int is used as the coordinates, the
filter mode must be NEAREST (see https://man.opencl.org/read_imagef1d.html):

 > The read_imagef calls that take integer coordinates must use a sampler with
 > filter mode set to CLK_FILTER_NEAREST, normalized coordinates set to
 > CLK_NORMALIZED_COORDS_FALSE and addressing mode set to CLK_ADDRESS_CLAMP_TO_EDGE,
 > CLK_ADDRESS_CLAMP or CLK_ADDRESS_NONE; otherwise the values returned are
 > undefined.

### DIFF
--- a/tests/accessor/accessor_api_image_common.h
+++ b/tests/accessor/accessor_api_image_common.h
@@ -1104,16 +1104,26 @@ class image_accessor_api_sampled_r {
     const bool useNormalized =
         m_sampler.coordinate_normalization_mode ==
             sycl::coordinate_normalization_mode::normalized;
+    const bool useNearest =
+        m_sampler.filtering_mode == sycl::filtering_mode::nearest;
+
     if (useNormalized) {
       const bool worksForLower =
         check_read<acc_coord_tag::use_normalized_lower>(idx);
       if (worksForLower)
         check_read<acc_coord_tag::use_normalized_upper>(idx);
     } else {
-      const bool worksForInteger =
-        check_read<acc_coord_tag::use_int>(idx);
-      if (worksForInteger)
+      const bool worksForFloat =
         check_read<acc_coord_tag::use_float>(idx);
+      if (worksForFloat && useNearest) {
+        /*
+          OpenCL spec rules that when int is used as the coordinates, the
+          filter mode must be NEAREST: "The read_imagef calls that take integer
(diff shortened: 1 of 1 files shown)


keryell (approved):

ProGTX (approved):

psamolysov-intel:
@bader @illuhad Could you review this PR, please?

tests/accessor/accessor_api_image_common.h
```diff
@@ -1104,16 +1104,26 @@ class image_accessor_api_sampled_r {
...
         check_read<acc_coord_tag::use_normalized_upper>(idx);
     } else {
-      const bool worksForInteger =
-        check_read<acc_coord_tag::use_int>(idx);
-      if (worksForInteger)
+      const bool worksForFloat =
         check_read<acc_coord_tag::use_float>(idx);
+      if (worksForFloat && useNearest) {
+        /*
+          OpenCL spec rules that when int is used as the coordinates, the
```
bader, line 1120: We should reference SYCL spec, not OpenCL.

bader (approved):

tests/accessor/accessor_api_image_common.h
keryell: Good point.

keryell (changes requested):
Almost there!
There is still a comment to address from Alexey.

psamolysov-intel:
@keyerll @bader As I know, we requested a changes to the SYCL 2020 specification about this moment. Unfortunately, I have no idea what should I say here from the SYCL point of view.

psamolysov-intel:
@bader @keryell As @yuriykoch (thank you Yuriy) informed me, the test were written for the SYCL 1.2.1 and still are not updated for SYCL 2020. SYCL 1.2.1 mentions that SYCL inherits the OpenCL specification's capabilities and restrictions, so I've added this mention into the comment, now it contains references to both specs. Does it look good?

bader:
The comment looks okay for SYCL-1.2.1 test.
@keryell, what do you think?

keryell (approved):
Thanks.

PR #239, opened 2021-12-17
Add tests for sycl::ext::oneapi::property_value

kopylovanat:
sycl::ext::oneapi::property_value is part of extension [SYCL_EXT_ONEAPI_PROPERTY_LIST](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/PropertyList/SYCL_EXT_ONEAPI_property_list.asciidoc)
Extension [SYCL_EXT_ONEAPI_DEVICE_GLOBAL](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/DeviceGlobal/SYCL_INTEL_device_global.asciidoc) is also used to check  how compile-time-constant properties work.


keryell (commented):
So what is the current decision about where we put tests from an extension?
In a `oneapi` subdirectory somewhere mimicking the normal test directory so we can look quickly at all the tests from an extension easily?

bader:
> So what is the current decision about where we put tests from an extension? 

This question was raised in https://github.com/KhronosGroup/SYCL-CTS/pull/200 and I see that @vasilytric created a separate directory in that PR. I suggest we agree on the location there and then apply it to other PRs. @keryell, does it sound okay to you?

kopylovanat:
> So what is the current decision about where we put tests from an extension?
> In a `oneapi` subdirectory somewhere mimicking the normal test directory so we can look quickly at all the tests from an extension easily?

Moved tests to subfolder `extension` similarly to how it done in  https://github.com/KhronosGroup/SYCL-CTS/pull/200

docker/computecpp/configure.sh
```diff
@@ -7,4 +7,5 @@ cmake . -G Ninja -B build \
     -DCMAKE_PREFIX_PATH=/sycl \
     -DCMAKE_BUILD_TYPE=Release \
     -DSYCL_CTS_ENABLE_FULL_CONFORMANCE=0 \
+    -DSYCL_ENABLE_EXT_ONEAPI_PROPERTY_LIST_TESTS=0 \
```
ProGTX, line 10: All extension tests should be off by default, so we don't have to specify the variables for the implementations that don't support them.

docker/dpcpp/configure.sh
```diff
@@ -8,4 +8,5 @@ cmake . -G Ninja -B build \
     -DCMAKE_CXX_COMPILER=/sycl/bin/clang++ \
     -DCMAKE_BUILD_TYPE=Release \
     -DSYCL_CTS_ENABLE_FULL_CONFORMANCE=0 \
+    -DSYCL_ENABLE_EXT_ONEAPI_PROPERTY_LIST_TESTS=0 \
```
ProGTX, line 11: Is this intentional? Presumably DPC++ supports this extension?

ProGTX (approved):

docker/computecpp/configure.sh
kopylovanat: Removed.

docker/dpcpp/configure.sh
kopylovanat: These tests are off by default for all compilers. so removed this too.

keryell (approved):
Good!

fraggamuffin:
LGTM

keryell:
There are some merge conflicts to fix.

PR #234, opened 2021-12-09
Bump DPC++ and hipSYCL versions used for CI

psalz:
This bumps the version of DPC++ to 6699a5d (December 8) and hipSYCL to b836149 (December 9). I've also updated the CI filter for DPC++, which is now able to compile 12 additional test categories, failing for 2 new ones, for a total of 5 categories that do not compile yet.

I've also adjusted the CI workflow to encode an implementation's version into the Docker image tag (i.e. there is now a `dpcpp-6699a5d` tag, as opposed to just `dpcpp`) so that submitting this PR won't affect current CI runs (until it is merged).

Disable DPC++ range rounding feature to speed up compilation.

PR #240, opened 2021-12-20
[DPCPP] Upgrade DPCPP version for CI

psamolysov-intel:
I wish to remove the usm tests from the exclude list for
DPCPP and this requires using DPCPP with support for
the sycl::atomic_ref class template. So, the version of
the DPCPP compiler should be upgraded.

### DIFF
--- a/.github/workflows/build_ci_containers.yml
+++ b/.github/workflows/build_ci_containers.yml
@@ -41,7 +41,7 @@ jobs:
           # - sycl-impl: computecpp
           # version: 2.6.0
           - sycl-impl: dpcpp
-            version: f5c838cd
+            version: 1f3f9b9b
           - sycl-impl: hipsycl
             version: 72a29fb9
     steps:


bader (approved):
LGTM.
FYI, it interferes with https://github.com/KhronosGroup/SYCL-CTS/pull/234, which bumps DPC++ version among other things.

psamolysov-intel:
@bader Thank you for the information. @psalz could you have a look too?

psalz:
The current process for updating CI container image versions is somewhat flawed in that the PR itself does not get checked with the new version. If we had merged this PR directly, future CI runs would have failed since with newer versions of DPC++ two additional test categories (`exceptions` and `opencl_interop`) fail to compile (while several additional categories *do* now compile). This is something I tried to address in #234 by introducing versioned image tags, but we ran into some trouble with long compilation times and ComputeCpp.

Anyway, I've opened a new PR #241 that introduces versioned image tags while now also bumping DPC++ to the version you are suggesting here (1f3f9b9), and adding the failing categories to the filter. I'll therefore go ahead and close this PR!

PR #241, opened 2021-12-20
Bump DPC++ and hipSYCL versions used for CI

psalz:
This PR supersedes both #234 and #240.

Changes compared to #234:
- Bump DPC++ to 1f3f9b9 (December 19) to accommodate #240 
- Work around ComputeCpp container images not using versioned tags yet
- Don't enable a ton of new test categories for DPC++ due to compile time concerns (which we should address in a separate PR, see also the discussion on #234).

(Note that CI will fail on this initially since the new container for DPC++ is [still building](https://github.com/KhronosGroup/SYCL-CTS/runs/4586121973?check_suite_focus=true), need to re-run checks after that).

### DIFF
--- a/.github/workflows/build_ci_containers.yml
+++ b/.github/workflows/build_ci_containers.yml
@@ -41,9 +41,9 @@ jobs:
           # - sycl-impl: computecpp
           # version: 2.6.0
           - sycl-impl: dpcpp
-            version: f5c838cd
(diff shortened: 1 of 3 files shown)


bader (approved):

psalz:
Not sure why the status checks aren't updating but since CI has passed for all three implementations I will just go ahead an merge this.

PR #244, opened 2022-01-05
Bump ComputeCpp CI container version to 2.8.0

psalz:
ComputeCpp can now compile 8 additional test categories!

PR #245, opened 2022-01-06
Use ccache to speed up CI runs

psalz:
Following some discussion in #243, this introduces `ccache` as a mechanism to significantly speed up our CI builds. Since most PRs only touch a few files pertaining to some test category, we should be able to achieve very high cache hit rates.

Additional changes:
- Unify the CI workflows for building the testing containers and compiling the CTS into a single file. I originally separated them  mostly as a precaution, to avoid allowing external contributors to build containers that are then pushed to Khronos' DockerHub. However, this turned out to be quite inconvenient, as it required PRs that modify the CI setup (e.g. to bump a SYCL implementation's version) to always originate from the main repo, as otherwise a run of the container building workflow could not be manually triggered. Since GitHub now supports a setting (which I enabled) that requires manual approval for CI runs from external contributors, I think we can safely merge the two workflows.
- Remove the build time and date information stored within each (legacy) test case. This information wasn't being used, and the `__TIME__` and `__DATE__` macros used to generate the data would otherwise invalidate most of the ccache on every new build.

This PR is based on #244, so that one should be merged first.


psalz:
It looks like even though I've manually run the workflow on this branch before opening the PR, the CI run for this PR won't benefit from `ccache` yet because of GitHub's [restrictions for accessing a cache](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) -- although I don't fully understand why. [Here](https://github.com/KhronosGroup/SYCL-CTS/runs/4728929454?check_suite_focus=true) you can see a manual re-run of my previous run. While DPC++ required 47 minutes in the original run, the re-run completed in less than 2 minutes.

ProGTX (approved):
Nice work, thank you.

Another suggestion for improving build times: use a different linker. The default BFD linker is incredicly slow. This becomes even more important with ccache and should be simple to add. `gold` is very easy to set up, maybe even go with LLD.

bader (approved):
@alexbatashev, FYI.

alexbatashev (approved):
2 minutes sounds incredible!

psalz:
> Another suggestion for improving build times: use a different linker. The default BFD linker is incredicly slow. This becomes even more important with ccache and should be simple to add. `gold` is very easy to set up, maybe even go with LLD.

Yes, you're right. We could even go for the new kid on the block, [mold](https://github.com/rui314/mold/) (although I'm not sure how much we'll profit on GitHub's 2 vCores). I'll look into it when I have time.

PR #250, opened 2022-01-12
Tests for SYCL 2020 multi_ptr members

AndreyXRomanov:
Provides tests for multi_ptr members according to SYCL 2020 spec.

Depends on
- #231 
- #237 

Requires SYCL 2020 multi_ptr class implementation for compilation.


bader:
@AndreyXRomanov, please, use correct spelling when referring to the specification version - "SYCL 2020". SYCL-2020 can be used if you refer to the branch name in SYCL-CTS repository and I don't know the context where `sycl2020` can be used.

tests/common/get_cts_string.h
```diff
@@ -0,0 +1,161 @@
...
+namespace sycl_cts {
+namespace get_cts_string {
+
+enum class verbosity { brief = 0, detail };
+
+/** @brief Stringify bool value
+ *  @tparam bool flag Value to convert
+ *  @retval String with interpretation of bool
+ */
+inline std::string for_bool(bool flag) { return flag ? "true" : "false"; }
```
ProGTX, line 23: It seems to me like most (if not all) of these functions could be `constexpr` and return `std::string_view`

tests/common/get_cts_string.h
```diff
@@ -0,0 +1,161 @@
...
+/** @brief Return string's description depending on the type of bundle_state and
+ * verbosity level
+ *  @tparam State bundle state
+ *  @tparam Level verbosity level
+ *  @retval string's description of bundle state
+ */
+template <sycl::bundle_state State, verbosity Level = verbosity::brief>
+inline std::string for_bundle_state() {
+  std::string result;
+  if constexpr (State == sycl::bundle_state::input) {
```
ProGTX, line 34: I think that if you make the whole function `constexpr` then you can get away with a simple switch statement - you're not guaranteed that optimal code will be generated, but it becomes much more likely. The same for other functions.

tests/multi_ptr/multi_ptr_members.h
```diff
@@ -0,0 +1,128 @@
...
+         "sycl::multi_ptr::iterator_category doesent equal to "
+         "std::random_access_iterator_tag" +
+             log_suffix);
+  }
+
+  if constexpr (Decorated == sycl::access::decorated::no) {
+    if (!std::is_same_v<decltype(multi_ptr)::pointer,
+                        std::add_pointer_t<VariableT>>) {
+      FAIL(log,
+           "sycl::multi_ptr::pointer doesent equal to "
```
ProGTX, line 112:
```suggestion
           "sycl::multi_ptr::pointer doesn't equal to "
```

tests/multi_ptr/multi_ptr_members.h
```diff
@@ -0,0 +1,128 @@
...
+           "std::add_pointer_t<value_type>" +
+               log_suffix);
+    }
+    if (!std::is_same_v<decltype(multi_ptr)::reference,
+                        std::add_lvalue_reference_t<VariableT>>) {
+      FAIL(log,
+           "sycl::multi_ptr::reference doesent equal to "
```
ProGTX, line 119:
```suggestion
           "sycl::multi_ptr::reference doesn't equal to "
```

tests/multi_ptr/multi_ptr_members_fp64.cpp
```diff
@@ -0,0 +1,45 @@
...
+   */
+  void get_info(test_base::info &out) const override {
+    set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
+  }
+
+  /** execute this test
+   */
+  void run(util::logger &log) override {
+    auto queue = util::get_cts_object::queue();
+    if (!queue.get_device().has(sycl::aspect::fp16)) {
```
ProGTX, line 32:
```suggestion
    if (!queue.get_device().has(sycl::aspect::fp64)) {
```

ProGTX (approved):

tests/multi_ptr/multi_ptr_members.h
AndreyXRomanov: Fixed misprints

tests/multi_ptr/multi_ptr_members_fp64.cpp
ProGTX: I must have missed that, why is that better than checking for the aspect?

tests/common/get_cts_string.h
ProGTX: I see now that the function can't be `constexpr` because of the verbosity level, which is unfortunate. The overhead wouldn't be much IMO, especially compared to the rest of the test. The compiler is also likely to generate a jump table from a switch statement instead of branches anyway. In any case, I don't think this is that important.

tests/multi_ptr/multi_ptr_members_fp64.cpp
AndreyXRomanov: I've forgotten to write it, sorry. 
Basically, `util::extensions::availability` does the same thing, but it will protect us from a lot of changes in the future if method `queue.get_device().has(..)` will change. We will just modify the `availability` function instead of changing every test, that requires the `fp16` or `fp64` operations.

tests/common/get_cts_string.h
AndreyXRomanov: Thanks for the answer.

bader (approved):

keryell (approved):

PR #252, opened 2022-01-12
Tests for SYCL 2020 multi_ptr accessor constructor

AndreyXRomanov:
Provides tests for multi_ptr accessor constructor according to SYCL 2020 spec.

Depends on:
- #231
- #237
- #250 
- #251 

Requires SYCL 2020 multi_ptr class implementation for compilation.


tests/multi_ptr/multi_ptr_accessor_constructor.h
```diff
@@ -0,0 +1,147 @@
...
+template <typename T>
+class check_multi_ptr_accessor_constructor_for_type {
+ public:
+  void operator()(sycl_cts::util::logger& log, const std::string& type_name) {
+    using namespace sycl::access;
+    // Accroding to spec of multi_ptr, constructor can take only
+    // global_space and generic_space accessors
+    run_tests_for_space<T, address_space::global_space>(log, type_name);
+    run_tests_for_space<T, address_space::generic_space>(log, type_name);
+  }
```
tomdeakin, line 142: The spec says these constructors only exist for certain address spaces so can we have a test which we expect to fail if a different address space is used?

Also are we missing the deprecated constructors?

tomdeakin (changes requested):

tests/multi_ptr/multi_ptr_accessor_constructor.h
AndreyXRomanov: > test which we expect to fail if a different address space is used

Looks like a negative case to me. We usually do not cover negative cases in tests.

> Also are we missing the deprecated constructors?

We have different tests for legacy constructors [here](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/multi_ptr/multi_ptr_legacy_constructors_common.h). Is this what you meant?

tomdeakin: > We usually do not cover negative cases in tests.

OK, thanks for confirming.

> We have different tests for legacy constructors here. Is this what you meant?

Yes, they look OK. Sorry I missed them.

tomdeakin (approved):

fraggamuffin:
LGTM

keryell (approved):

PR #254, opened 2022-01-17
Functional test for extension oneapi::device_global. Several kernels on one device.

AndreyXRomanov:
This PR provides a functional test for `device_global`:
The test modifies the `device_global` instance in one kernel and expects to read the same value from another kernel on the same device.

Requires device_global class implementation and enum class `sycl::access::decorated` for build.

[Link](https://github.com/KhronosGroup/SYCL-CTS/pull/254/files/bacf91dabcdb7b8bcab10c270ec1f9377c220a72..HEAD) to changes related to this PR.

### DIFF
--- a/tests/extension/oneapi_device_global/device_global_functional_several_kernels_one_device.cpp
+++ b/tests/extension/oneapi_device_global/device_global_functional_several_kernels_one_device.cpp
@@ -14,7 +14,7 @@
 
 #include "../../common/common.h"
 #include "../../common/type_coverage.h"
-#include "../../util/array.h"
+#include "../../../util/array.h"
 #include "device_global_common.h"
 #include "type_pack.h"
 
@@ -31,7 +31,7 @@ namespace several_kernel_in_one_device {
 template <typename T>
 oneapi::device_global<T> dev_global;
 template <typename T>
-oneapi::device_global<T> const_dev_global;
+const oneapi::device_global<T> const_dev_global;
 
 template <typename T>
 struct first_kernel;
@@ -68,7 +68,7 @@ void run_test(util::logger& log, const std::string& type_name) {
   // If T is not array then element_type == T
   // As we fill the array with the same numbers with type of the array element,
   // we can store only one number to compare whole array with this number
-  using element_type = typename std::remove_extent_t<T>;
+  using element_type = std::remove_extent_t<T>;
 
   // Creating variables for store result of comparing
   constexpr size_t checks_size = integral(indx::size);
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

tests/extension/oneapi_device_global/device_global_functional_several_kernels_one_device.cpp
```diff
@@ -0,0 +1,154 @@
...
+      });
+    });
+    queue.wait_and_throw();
+  }
+
+  // Using remove_extent for get type of element when T is array
+  // If T is not array then element_type == T
+  // As we fill the array with the same numbers with type of the array element,
+  // we can store only one number to compare whole array with this number
+  using element_type = typename std::remove_extent_t<T>;
```
keryell, line 71: Do we need the `typename` here?

tests/extension/oneapi_device_global/device_global_functional_several_kernels_one_device.cpp
```diff
@@ -0,0 +1,154 @@
...
+using namespace device_global_common_functions;
+#if defined(SYCL_EXT_ONEAPI_PROPERTY_LIST) && \
+    defined(SYCL_EXT_ONEAPI_DEVICE_GLOBAL)
+namespace oneapi = sycl::ext::oneapi;
+
+namespace several_kernel_in_one_device {
+template <typename T>
+oneapi::device_global<T> dev_global;
+template <typename T>
+oneapi::device_global<T> const_dev_global;
```
keryell, line 34: Should we have
```suggestion
oneapi::device_global<std::add_const_t<T>> const_dev_global;
```
?
I do not understand well what you are trying to inthis test.

tests/extension/oneapi_device_global/device_global_functional_several_kernels_one_device.cpp
AndreyXRomanov: Good catch. We actually don't need it. I removed the `typename` here 3f512947af398d2ca0aa730c7c3b11276e411931

tests/extension/oneapi_device_global/device_global_functional_several_kernels_one_device.cpp
AndreyXRomanov: I forgot to put the `const` modifier to the variable. It should be  `const oneapi::device_global<T> const_dev_global` as in other tests. I fixed it in 3f512947af398d2ca0aa730c7c3b11276e411931.

In this test, I'm testing a case, when a user tries to access device_global instance from different kernels on one device. In the first kernel, the instance is being modified and in the second kernel, we expect, that the instance will contain the changed value. With `const_dev_global` we expect to get the default value from the second kernel.

keryell (approved):
Thanks

PR #259, opened 2022-01-27
CI: Pull hipSYCL from sycl-cts branch instead of fixed commit

illuhad:
Pull hipSYCL used in CI from `sycl-cts` branch instead of using fixed commit of hipSYCL. This allows us to fix CTS issues that come up in CTS CI due to hipSYCL bugs without having to change CTS CI everytime.

(I hope it works the way I did it by just specifying branch instead of commit hash)


illuhad:
Reopened in #260

PR #260, opened 2022-01-27
CI: Pull hipSYCL from sycl-cts branch instead of fixed commit v2

illuhad:
Pull hipSYCL used in CI from sycl-cts branch instead of using fixed commit of hipSYCL. This allows us to fix CTS issues that come up in CTS CI due to hipSYCL bugs without having to change CTS CI everytime.

(I hope it works the way I did it by just specifying branch instead of commit hash - Resubmitted PR to allow recreating docker container)

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -57,7 +57,7 @@ jobs:
           - sycl-impl: dpcpp
             version: ec97c57
           - sycl-impl: hipsycl
-            version: b836149
+            version: sycl-cts
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -96,7 +96,7 @@ jobs:
           - sycl-impl: dpcpp
             version: ec97c57
           - sycl-impl: hipsycl
-            version: b836149
+            version: sycl-cts
     env:
       container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
       parallel-build-jobs: 2


illuhad:
It seems that the repository is configured to require hipSYCL checks with a hardcoded commit to succeed. If we want to update the commit used in CI as in this PR, the required status check it looks for no longer exists and the PR cannot be merged. Does anybody know how we can fix this?

psalz:
> It seems that the repository is configured to require hipSYCL checks with a hardcoded commit to succeed. If we want to update the commit used in CI as in this PR, the required status check it looks for no longer exists and the PR cannot be merged. Does anybody know how we can fix this?

This is a known issue and there is nothing we can really do about (except for removing the required check), as long as GitHub doesn't allow to specify required checks with some kind of wildcard syntax. It's only a cosmetic problem though, we can still "force-merge" these.

bader (approved):

psalz:
After thinking about this some more, I have some concerns with this approach:
- Docker image builds are cached, so I don't think they will pick up changes to the branch if nothing changes within the Dockerfile.
- While the intent here is of course to improve hipSYCL's conformance and to be able to quickly address issues, chances are that regressions / incompatibilities will also be introduced from time to time. If we were to work around the image caching, this would mean that CI runs for unrelated PRs could suddenly break for seemingly no reason, simply because hipSYCL got updated.
- As hipSYCL's conformance improves over time, we have to update the respective [CI filters](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/ci/hipsycl.filter) to build more and more tests. This will still have to happen through dedicated PRs (i.e., we could also just bump the git revision there).

illuhad:
> Docker image builds are cached, so I don't think they will pick up changes to the branch if nothing changes within the Dockerfile.

Okay, this is a good technical point that we will need to look at. My intended workflow was to trigger the workflow manually if I want the image to be rebuilt, so maybe there can also be a manual mechanism to purge the cache?

> While the intent here is of course to improve hipSYCL's conformance and to be able to quickly address issues, chances are that regressions / incompatibilities will also be introduced from time to time. 

We don't test against the latest develop branch, but only against the sycl-cts branch of hipSYCL. The idea is that this would be a stable branch that we are confident will work in CTS. We could also add CTS CI in hipSYCL if it makes people happier with this - then the branch would be pre-tested against CTS before it gets used by CTS CI.

> If we were to work around the image caching, this would mean that CI runs for unrelated PRs could suddenly break for seemingly no reason, simply because hipSYCL got updated.

I think this is unlikely and can be further minimized as explained above. As it stands now, CI for unrelated PRs also break because hipSYCL did *not* get updated. The same could happen in the future with other hipSYCL issues or other issues for any other implementation.

With the current approach whenever we find some minor issue with the CI of any implementation, we will have to create a PR to update commits, wait for necessary approvals and then perform a force-merge because regular merging doesn't work. This is cumbersome and slows down the process for everyone.

> As hipSYCL's conformance improves over time, we have to update the respective CI filters to build more and more tests. This will still have to happen through dedicated PRs (i.e., we could also just bump the git revision there).

Sure, but how well this will solve issues such as the current one is, I think, unclear as it would depend on the frequency with which such PRs would be submitted versus the frequency of things breaking.

docker/hipsycl/Dockerfile
```diff
@@ -15,7 +15,6 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     rm -rf /var/lib/apt/lists*
 
 RUN git clone https://github.com/illuhad/hipSYCL.git \
-      --branch=develop --single-branch --shallow-since=2021-08-01 \
```
keryell, line 18: I am curious about this `--shallow-since=2021-08-01`.
If you want to minimize transfers you could still use some shallowing with `--depth 1 --shallow-submodules --branch $IMPL_VERSION`

docker/hipsycl/Dockerfile
```diff
@@ -15,7 +15,6 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
       --recurse-submodules /tmp/hipSYCL && \
     cd /tmp/hipSYCL && \
     git checkout $IMPL_VERSION && \
```
keryell, line 20: If the shallow git clone line picking the branch is used, this is redundant.

ProGTX (approved):

keryell (approved):

psalz:
> > Docker image builds are cached, so I don't think they will pick up changes to the branch if nothing changes within the Dockerfile.
> Okay, this is a good technical point that we will need to look at. My intended workflow was to trigger the workflow manually if I want the image to be rebuilt, so maybe there can also be a manual mechanism to purge the cache?

Yes we could probably conditionally disable caching based on the build trigger.

> > While the intent here is of course to improve hipSYCL's conformance and to be able to quickly address issues, chances are that regressions / incompatibilities will also be introduced from time to time.
> 
> We don't test against the latest develop branch, but only against the sycl-cts branch of hipSYCL. The idea is that this would be a stable branch that we are confident will work in CTS. We could also add CTS CI in hipSYCL if it makes people happier with this - then the branch would be pre-tested against CTS before it gets used by CTS CI.

I think it would be a good idea in general to test hipSYCL against the CTS during CI, but there is probably a non-negligible maintenance overhead involved.

> > If we were to work around the image caching, this would mean that CI runs for unrelated PRs could suddenly break for seemingly no reason, simply because hipSYCL got updated.
> 
> I think this is unlikely and can be further minimized as explained above. As it stands now, CI for unrelated PRs also break because hipSYCL did _not_ get updated. The same could happen in the future with other hipSYCL issues or other issues for any other implementation.

What I mean is that it's not transparent from a PR workflow's perspective. You could file a PR that passes CI just fine, then address some minor reviewer comment and suddenly CI breaks because hipSYCL was updated under the hood. With the current approach this can only happen when a PR does a back-merge of the target branch, and then it is at least apparent that hipSYCL was updated in the workflow file.

> With the current approach whenever we find some minor issue with the CI of any implementation, we will have to create a PR to update commits, wait for necessary approvals and then perform a force-merge because regular merging doesn't work. This is cumbersome and slows down the process for everyone.

Since you brought it up, I timed the last version bump I did (#269) and it took me 5 minutes, out of which 3.5 minutes where spent on [regenerating the filter](https://github.com/KhronosGroup/SYCL-CTS/blob/f0a22192510e1126763bb0559c0c3bd4d93073e9/ci/generate_exclude_filter.py). Before we move further with this, maybe you could give the current workflow a try yourself? IMO it's really not that bad. I think it would also be reasonable to adjust our policy around PRs for version bumps to not require any approvals, so they could be merged as soon as CI is green. At least if done by the respective implementors.

fraggamuffin:
close for more thinking

PR #267, opened 2022-01-31
Tests for exceptions thrown by `sycl::handler` and `sycl::queue` copy member functions with `device_global`

AndreyXRomanov:
This PR provides tests for the exceptions that are thrown by sycl::queue and sycl::handler copy methods overloads with `deivce_global` as one of the parameters.


tests/extension/oneapi_device_global/device_global_exceptions.cpp
```diff
@@ -0,0 +1,289 @@
...
+      is_exception_thrown &= false;
+    } catch (sycl::exception const& e) {
+      is_exception_correct &= is_errc_invalid(e);
+    }
+  });
+  queue.wait_and_throw();
+
+  if (!is_exception_thrown) {
+    FAIL(log, get_case_description(
+                  "deivice_global: sycl::handler .copy() method exception",
```
keryell, line 77: Change everywhere
```suggestion
                  "device_global: sycl::handler .copy() member function exception",
```

tests/extension/oneapi_device_global/device_global_exceptions.cpp
```diff
@@ -0,0 +1,289 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for the exceptions that are thrown by sycl::queue and
+//  sycl::handler copy methods overloads
```
keryell, line 6: In C++ we say `member functions` instead of `methods`.

keryell (changes requested):

tests/extension/oneapi_device_global/device_global_exceptions.cpp
AndreyXRomanov: Fixed

tests/extension/oneapi_device_global/device_global_exceptions.cpp
AndreyXRomanov: Applied change to every comment with "method"

keryell: You need also to fix the typo `deivice_global` → `device_global` everywhere.

AndreyXRomanov: Thank you, I missed the other ones. Fixed in this PR.

bader (approved):

tests/extension/oneapi_device_global/device_global_exceptions.cpp
```diff
@@ -0,0 +1,289 @@
...
+  bool is_exception_thrown{true};
+  bool is_exception_correct{true};
+  auto queue = util::get_cts_object::queue();
+
+  queue.submit([&](sycl::handler& cgh) {
+    // Try to write data beyond the end of the destination variable
+    // Have to throw an exception
+    try {
+      cgh.template copy<T>(src.get(), dev_global<T>, N, N / 2);
+      is_exception_thrown &= false;
```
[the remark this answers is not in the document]
AndreyXRomanov: Removed bitwise AND assignment in some places.
Also changed initial value for result variables to false, to remove possibility forget to set it and get false-positive test result.

commit https://github.com/KhronosGroup/SYCL-CTS/pull/267/commits/1b8b22c18e29eaf901aaff68722936bf3cad55f4

tests/extension/oneapi_device_global/device_global_exceptions.cpp
```diff
@@ -49,24 +49,26 @@ void run_test(util::logger& log, const std::string& type_name) {
...
+      is_exception_thrown = true;
+      is_exception_correct = is_errc_invalid(e);
     }
 
     try {
       cgh.template copy<T>(dev_global<T>, dest.get(), N, N / 2);
-      is_exception_thrown &= false;
+      is_exception_thrown = false;
     } catch (sycl::exception const& e) {
+      is_exception_thrown &= true;
```
bader, line 71: This code is not correct if the first check fails. I don't think that changing default value to `false` is a good idea.

AndreyXRomanov: I don't think that code is not correct. If the first check fails and the second is not, then `is_exception_thrown` will contain false and the test will fail as it is supposed to.

But I agree that `false` as the default value is not a good idea, because it makes code more complicated. If the initial value is `false`, you have to change `is_exception_thrown` in every case, not only in failure cases.

I have changed the initial value for the result variables to `true`.

tests/extension/oneapi_device_global/device_global_exceptions.cpp
```diff
@@ -0,0 +1,293 @@
...
+    } catch (sycl::exception const& e) {
+      is_exception_correct = is_errc_invalid(e);
+    }
+
+    try {
+      cgh.copy<T>(dev_global<T>, dest.get(), N, N / 2);
+      is_exception_thrown = false;
+    } catch (sycl::exception const& e) {
+      is_exception_correct &= is_errc_invalid(e);
+    }
```
bader, line 71: Please, move copy operations to different command group scopes.

From https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:command.group.scope

> A [command group scope](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#command-group-scope), as defined in [Section 3.7.1](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:executionmodel), may execute a single [command](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#command) such as invoking a kernel, copying memory, or executing a host task. It is legal for a [command group scope](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#command-group-scope) to statically contain more than one call to a [command](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#command) function, but **_any single execution of the [command group function object](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#command-group-function-object) may execute no more than one [command](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#command)_**.

bader (changes requested):

tests/extension/oneapi_device_global/device_global_exceptions.cpp
bader: > I don't think that code is not correct. If the first check fails and the second is not, then `is_exception_thrown` will contain false and the test will fail as it is supposed to.

Maybe I misunderstand the intention of this line. What is line of code supposed to be doing then?

AndreyXRomanov: This line is supposed to set `is_exception_thrown` to `true` if an exception was thrown, but only if the first invocation of member function caused exception too. If the first invocation did not cause an exception, then `is_exception_thrown` has to be `false` until the end of the test. To implement such behavior I used operator `&=` to avoid overwriting `false` to `true`.

Anyway, I changed the initial value to `true` so it doesn't matter anymore.

tests/extension/oneapi_device_global/device_global_exceptions.cpp
AndreyXRomanov: Do we need to apply this change to [this place](https://github.com/KhronosGroup/SYCL-CTS/blob/f8d877560d82e2e4a7eacefe9d252178a7d602c1/tests/extension/oneapi_device_global/device_global_handler_copy_memcpy_overloads.cpp#L72-L85) too?

bader: Yes. The spec rule applies to any SYCL code.

AndreyXRomanov: Okay. I'll make PR for it.

bader (approved):
Thanks.

tests/extension/oneapi_device_global/device_global_exceptions.cpp
psalz: cf https://github.com/KhronosGroup/SYCL-Docs/issues/223

ProGTX (approved):

keryell (approved):
Thanks!

PR #271, opened 2022-02-03
Fix test for queue::copy with device_global and depEvents parameters

kopylovanat:


ProGTX (approved):

tests/extension/oneapi_device_global/device_global_queue_copy.cpp
```diff
@@ -87,7 +87,7 @@ void run_test_copy_to_device_global(util::logger& log,
   auto event3 = queue.copy(src_data, dev_global3<T>,
                                 sizeof(T) / sizeof(element_type), 0, depEvents);
   event3.wait();
-  for (size_t i = numEvents - 1; i >= 0; --i) {
+  for (size_t i = numEvents - 1; i > 1; --i) {
```
keryell, line 90: While you are here, since there were bugs, this means that these raw loops are not so trivial, so add a comment on both of them with the intent.

keryell (approved):

tests/extension/oneapi_device_global/device_global_queue_copy.cpp
kopylovanat: Added comments.

PR #274, opened 2022-02-03
Add exception details to logs

yuriykoch:
To have std::error_code::value() available and to avoid application crash in case of nullptr messages


tests/common/macros.h
```diff
@@ -56,18 +56,34 @@ inline void set_test_info(sycl_cts::util::test_base::info& out,
...
+  const auto& code = e.code();
+  const auto& category = e.category();
+
+  auto append_str = [&message](const char* description, std::string&& value) {
+    message += description;
+    message += " - '" + value + "'\n";
+  };
+  auto append_cstr = [&append_str](const char* description, const char* value) {
+    if (!value) {
+      value = "nullptr";
```
bader, line 74:
```suggestion
      value = "nullptr\n";
```

bader (commented):
Can you give an example of how the output will look like with and w/o "value", please?

psalz:
I'm wondering whether this could somehow be abstracted to also work with Catch2's automatic printing of exceptions, i.e. by specializing [`Catch::StringMaker`](https://github.com/catchorg/Catch2/blob/devel/docs/tostring.md). Because as is, new-style test cases would not benefit from this change.

tests/common/macros.h
```diff
@@ -56,18 +56,34 @@ inline void set_test_info(sycl_cts::util::test_base::info& out,
...
-  std::string what = e.what();
-  if (!what.empty()) {
-    log.note("what - " + what);
-  }
+  std::string message = "SYCL exception\n";
+  // Using references to avoid object slicing
+  const auto& code = e.code();
+  const auto& category = e.category();
+
+  auto append_str = [&message](const char* description, std::string&& value) {
```
keryell, line 68:
```suggestion
  auto append_str = [&message](const char* description, const std::string& value) {
```

tests/common/macros.h
```diff
@@ -56,18 +56,34 @@ inline void set_test_info(sycl_cts::util::test_base::info& out,
...
+    message += description;
+    message += " - '" + value + "'\n";
+  };
+  auto append_cstr = [&append_str](const char* description, const char* value) {
+    if (!value) {
+      value = "nullptr";
+    }
+    append_str(description, std::string(value));
```
keryell, line 76:
```suggestion
    append_str(description, { value });
```

ProGTX (approved):

bader:
@yuriykoch, please, address code review comments.

yuriykoch:
Sorry for a delay, I'm planning to address all the comments alongside with the new-style test cases support request.

tests/common/macros.h
yuriykoch: Done, thanks

tests/common/macros.h
yuriykoch: Wouldn't it trigger unnecessary copy constructor in this case?

tests/common/macros.h
```diff
@@ -56,18 +56,34 @@ inline void set_test_info(sycl_cts::util::test_base::info& out,
...
+  };
+
+  // Collect exception details
+  append_cstr("category name", category.name());
+  append_str("code value", std::to_string(code.value()));
+  append_str("code message", code.message());
+  append_cstr("what", e.what());
+
+  // Print message in a single Catch2 warning
+  log.note(message);
```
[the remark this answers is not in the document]
yuriykoch: Sure. Looks like the `WARN` macro fits great for deprecated `log_exception()` call.

bader (changes requested):
@yuriykoch, please, fix pre-commit checks.

tests/common/exceptions.h
```diff
@@ -0,0 +1,302 @@
...
+}
+
+/**
+ * Stringification of SYCL exceptions for Catch2 tests
+ * Provides more details than simple what() message
+ *
+ * Note that neither this macro, not StringMaker specialization would provide
+ * support for expressions like `INFO("Exception: " << exception)`
+ */
+CATCH_TRANSLATE_EXCEPTION(const sycl::exception& e) {
```
psalz, line 284: I was confused at first why this works for a translation unit that does not include `common/exceptions.h` (e.g. `id.cpp`). However it turns out this only needs to be called once from *any* translation unit, which in this case probably was `test_base.cpp`. If you still plan to add `common/exceptions.cpp` (as you mentioned in a TODO note above), I suggest to move this call there.

tests/common/exceptions.h
```diff
@@ -0,0 +1,302 @@
...
+// Safely ensures free function availability w/o compile-time failures
+template <typename = void>
+struct sycl_make_error_code : std::false_type {};
+template <>
+struct sycl_make_error_code<
+    std::void_t<decltype(sycl::make_error_code(std::declval<sycl::errc>()))>>
+    : std::true_type {};
+
+}  // namespace has_function
+}  // namespace detail
```
psalz, line 64: Correct me if I'm wrong, but these SFINAE helpers test for *non-optional* features in implementations that haven't quite caught up yet, right? If so, I'd suggest to either add a note that these should be removed at some point, or preferably in view, simply do something like

```c++
#if !defined(__COMPUTECPP__) && !defined(__HIPSYCL__)
constexpr bool has_member_code = true;
constexpr bool has_member_category = true;
#else
constexpr bool has_member_code = false;
constexpr bool has_member_category = false;
#endif
```

tests/common/exceptions.h
```diff
@@ -0,0 +1,302 @@
...
+    const auto& category = e.category();
+    append_cstr("category name", category.name());
+  }
+
+  if constexpr (!detail::has_member::code<sycl::exception>::value) {
+    append_cstr("code", "not supported by implementation");
+  } else {
+    // Using reference to avoid object slicing
+    const auto& code = e.code();
+    append_str("code value", std::to_string(code.value()));
```
psalz, line 103: I think it would be very helpful to also print the corresponding `errc` enum value here!

tests/common/exceptions.h
```diff
@@ -0,0 +1,302 @@
...
+      return false;
+    } else {
+      // Semantic match
+      return other.code() == m_code_value;
+    }
+  }
+
+  std::string describe() const override {
+    std::ostringstream result;
+    result << "has code semantically matching to ";
```
psalz, line 173: I don't understand what you mean by "semantic match" here (and above), why not just "match"?

tests/common/exceptions.h
```diff
@@ -0,0 +1,302 @@
...
+    std::ostringstream result;
+    result << "has code value " << to_integral(m_code_value);
+    result << " for backend-specific category '" << expected_category().name();
+    result << "'";
+    return result.str();
+  }
+
+ private:
+  const std::error_category& expected_category() const {
+    return sycl::error_category_for<Backend>();
```
psalz, line 222: Seems like this is not supported by any (even DPC++, for me?) implementation yet.

psalz (changes requested):
Thanks, this is a lot more comprehensive than what I would have expected!

psalz:
Come to think of it: We can actually get rid of `test_base.cpp` altogether now, there is no need to manually wrap `test_base::run` in a `try`/`catch`-block, Catch2 will take care of failing the test and printing exception details by itself.

tests/common/exceptions.h
yuriykoch: The point here is that such header would be used within all SYCL CTS tests, so it's not intended to be removed.
I assume there would be two possible use-cases here.
 - Any new or existing SYCL implementation doesn't support non-optional API for SYCL exceptions
 - Any existing implementation encounter compilation error for already supported non-optional API

In any of these cases:
 - we could still gain from "what()" messages at least
 - it would be better to have specific SYCL CTS tests failed rather then all of them

For example:
 - there should be no failure if the test just includes `exceptions.h` or `common.h`
 - test should not fail if it just prints out exception information
 - test should fail on specific CATCH_THROW_MATCHES in case specific check is unavailable

Are there any concerns regarding suggested solution?

tests/common/exceptions.h
yuriykoch: My bad. I've got [p0824r1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0824r1.html) in mind and forgot that now SYCL simply has separate error code enums for each category, with no generic errors due to `error_condition` removal. Will fix and rename the matcher.

tests/common/exceptions.h
psalz: To be clear, I'm not questioning the intent of this code. I agree that it is important that fundamental CTS infrastructure such as this ensures compatibility with all implementations. However I don't think this should be a permament mechanism, and I'm not convinced that the complexity of introducing SFINAE helpers here is warranted either.

> so it's not intended to be removed.

All three SYCL implementations currently supported by the CTS are striving to become conformant at some point, which is when we could remove these helpers.

> Any new [...] SYCL implementation doesn't support non-optional API for SYCL exceptions

I don't think we should optimize for this hypothetical case. If we wanted any new SYCL implementation to immediately be able to compile the CTS, we would have to have these kinds of SFINAE helpers for *any and all* SYCL API interactions in CTS infrastructure code, which clearly we don't (and realistically, can't) have.

---

So all I'm saying is this: Ensuring compatibility with CCPP, DPC++ and hipSYCL is good (and required for CI), so having something like this (or my proposed `constexpr bool`s) is required at the moment. However I believe we should get rid of it as soon as those three implementations no longer require the workaround, and so we don't forget, I'm suggesting you add a TODO note :).

Or am I missing something?

tests/common/macros.h
keryell: Where ?

keryell: Passing by `const` reference is cheaper here than the move construction. Also, using an r-value reference prevents the l-value usage like
```
std::string a { "hello" };
append_str("yo", a);
```

keryell (approved):
Thanks.

tests/common/macros.h
yuriykoch: Missed `std::move` here. The `std::basic_string::operator+` has [different overloads](https://en.cppreference.com/w/cpp/string/basic_string/operator%2B) for l-value and r-value, so r-value restriction was intended. 
Still looks like any possible performance gain here doesn't worth it in the terms of further support, so it was aligned with the codebase to use `const std::string&`. Not to mention that `operator+` is not used anymore )

tests/common/exceptions.h
```diff
@@ -0,0 +1,302 @@
...
+
+#include "../../util/logger.h"
+#include "../common/conversion.h"
+#include "../common/macros.h"
+
+#include <sstream>
+#include <string>
+#include <utility>
+
+CATCH_REGISTER_ENUM(sycl::errc, sycl::errc::success, sycl::errc::runtime,
```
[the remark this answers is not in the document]
yuriykoch: I like this idea. The `CATCH_REGISTER_ENUM` was moved into the `common/enums.h`.

tests/common/exceptions.h
yuriykoch: I've added a `TODO` notice here as well. BTW, would it be OK to have any `.cpp` files within `tests/common` except `main.cpp`? Maybe it's better to do it within a separate PR in case of any possible architecture concerns due to `CMakeLists.txt` changes.

tests/common/exceptions.h
yuriykoch: Agree) Updated the code with feature flags and TODO notes regarding feature flags removal.

Preprocessor definitions used instead of constexpr variables because:
- looks like hipSYCL doesn't support `sycl::errc` enum yet
- there are too many cases when discarded constexpr substatement could still result in compilation failure

tests/common/exceptions.h
yuriykoch: Great catch, thanks. Done by 94dfef0afb0b95d78bc093b163b197322f408de7

tests/common/exceptions.h
yuriykoch: As result, there are 3 matchers now:

- `has_exception_category` - compares category only
- `equals_exception` - for generic errors, with `sycl::errrc` error codes and `sycl::sycl_category()` category
- `equals_exception_for` - for backend-specific errors

Though the `has_exception_category` has a limited functionality now, it was left to have all possible use-cases covered

tests/common/exceptions.h
psalz: I guess it should be fine, except that the name of the library (`main_funtion_object`) wouldn't be quite fitting anymore. Come to think of it, it seems to me that `stringify_sycl_exception` (as well as the call to `CATCH_TRANSLATE_EXCEPTION`) could also live in `CTS::utils`, which already consists of multiple translation units, right? At some point we need to reevaluate what goes where, but since exception printing is more of a "framework-level" responsibility, I would argue it fits there quite well.

tests/common/exceptions.h
yuriykoch: Also under feature flag now )

tests/common/exceptions.h
```diff
@@ -42,7 +42,7 @@ namespace sycl_cts {
 #define SYCL_CTS_SUPPORT_HAS_ERROR_CATEGORY_FOR 0
 #define SYCL_CTS_SUPPORT_HAS_MAKE_ERROR_CODE 0
 
-#elif defined(__DPCPP__)
+#elif defined(SYCL_IMPLEMENTATION_INTEL)
```
psalz, line 45: Good to know this exists now!

tests/common/exceptions.h
yuriykoch: What do you think about moving all three headers: `conversion.h`, `enums.h` and `exceptions.h` into the `CTS::utils`?

psalz: Sure, why not. I was just thinking, maybe we should get rid of `tests/common` altogether at some point.

keryell (approved):
Thanks!

bader:
@yuriykoch, please, fix the build with dpcpp.

../tests/math_builtin_api/math_builtin.h:141:5: error: use of undeclared identifier 'log_exception'; did you mean 'sycl_cts::log_exception'?

tests/common/exceptions.h
yuriykoch: @bader Thanks, updated with `__SYCL_COMPILER_VERSION`

tests/common/exceptions.h
yuriykoch: Done with renaming in a separate commit e6597c92d0561bda63d9b0172bbff9ff8b2b76e6 to avoid git history loss:

- `common/enums.h` -> `util/sycl_enums.h`
- `common/exceptions.h` -> `util/sycl_exceptions.h`
- `common/conversion.h` -> `util/conversion.h`

Also added `sycl_exceptions.cpp` file

bader (approved):

tests/common/macros.h
keryell: I am unsure about what you mean with "Missed `std::move` here." If this is about a missing `std::move` in my snippet, this is on purpose, since `a` could be used also later on.
But I think you are right, we do not care about these micro-optimziations in the CTS. :-)

yuriykoch: Oops, just to clarify - I meant that I had missed the `std::move` in original code. Sorry.

PR #275, opened 2022-02-08
Fix input parameter for make_event

kopylovanat:
Fixed input parameter for make_event in accordance with the SYCL 2020 specification.


kopylovanat:
> Can we remove `opencl_interop` from ci/dpcpp.filter or there are other blocking issues?

There is still a blocking issue with make_event.

ProGTX (approved):

keryell (approved):

PR #276, opened 2022-02-09
Move copy operations to different command group scopes in `sycl::handler` copy member functions overloads tests 

AndreyXRomanov:
This PR provides a patch to `sycl::handler` copy member functions overloads tests. The patch separates all operations with `single_task`,  `copy` and `memcpy` into different command group scopes.


bader (approved):

ProGTX (approved):

keryell (approved):
Good catch!

PR #277, opened 2022-02-10
Update `sycl::event` tests for SYCL 2020

psalz:
This is a complete rewrite of the tests for `sycl::event`, using the new Catch2 style.

The test cases pertaining to `event::wait_and_throw` rely heavily on the fact that exceptions thrown inside `host_task`s will be forwarded to the async handler, as was recently clarified in [KhronosGroup/SYCL-Docs#214](https://github.com/KhronosGroup/SYCL-Docs/issues/214).

I've added one test case (`"event::wait_and_throw reports asynchronous errors from related events on other queues"`) which I've marked as `[!mayfail]`, as DPC++ currently does not behave this way. I found the description in the spec rather ambiguous, so further clarification is needed here (I've also added the `[todo-spec]` tag, which is something I'd like to establish).

Additionally, I've snuck in some minor refactorings for various helper functions (introducing new overloads that don't require a deprecated `logger`), as well as a new `TODO_TEST_CASE` macro which can be used to designate test cases that should be implemented in the future.

Conformance status:
- ComputeCpp: Compiles but does not pass all tests (exceptions thrown in `host_task` are not forwarded)
- DPC++: Compiles and passes all tests except the one mentioned above
- hipSYCL: Does not compile due to `host_task` not being implemented, among other things


psalz:
(Turns out its a good idea to try and compile something after doing a refactoring... will fix shortly).

tests/event/event.cpp
```diff
@@ -0,0 +1,458 @@
...
+    CHECK(c != d);
+  }
+
+  SECTION("std::hash") {
+    const auto a = make_device_event();
+    const auto b = make_device_event();
+    const sycl::event c = a;
+    std::hash<sycl::event> hasher;
+    CHECK(hasher(a) != hasher(b));
+    CHECK(hasher(a) == hasher(c));
```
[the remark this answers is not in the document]
psalz: I do, which is why I've added that `TODO` note. However this will take some work, I will look at unifying this across tests in a future PR.

tests/event/event.cpp
```diff
@@ -0,0 +1,458 @@
...
+  SECTION("event::wait_and_throw") { e2.wait_and_throw(); }
+  SECTION("event::wait_and_throw(std::vector<event>)") {
+    sycl::event::wait_and_throw(std::vector{e2});
+  }
+
+  // This should be a no-op as the error already has been consumed
+  e1.wait_and_throw();
+
+  CHECK(teh.count() == 2);
+  CHECK(teh.has("foo"));
```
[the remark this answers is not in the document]
psalz: Okay sure, I've replaced it `"some-error"` and `"another-error"`.

psalz:
I've fixed the bug introduced in 8e4fdb0 as well as several race conditions surrounding `resolvable_host_event` (concurrency is hard).

tests/event/event.cpp
```diff
@@ -0,0 +1,466 @@
...
+    e2.wait();
+    // Since e2's CGF depends on e1, the latter must have completed by now
+    const auto status =
+        e1.get_info<sycl::info::event::command_execution_status>();
+    CHECK(status == sycl::info::event_command_status::complete);
+  }
+}
+
+// TODO: Figure out if/how we want to test this.
+// => Must throw exception w/ ercc:backend_mismatch if querying a paramter for
```
bader, line 415:
```suggestion
// => Must throw exception w/ errc::backend_mismatch if querying a parameter for
```

bader (approved):

tests/event/event.cpp
```diff
@@ -0,0 +1,466 @@
...
+
+  // A default constructed event is immediately completed.
+  const auto status =
+      event.get_info<sycl::info::event::command_execution_status>();
+  CHECK(status == sycl::info::event_command_status::complete);
+
+  // A default constructed event acts as though it was created from a
+  // default-constructed queue.
+  sycl::device default_device{};
+  CHECK(event.get_backend() == default_device.get_backend());
```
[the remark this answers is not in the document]
psalz: Right, I was implicitly following the second part of the default constructor's description:
> The event is constructed as though it was created from a default-constructed `queue`. Therefore, its backend is the same as the backend from the default device.

Interestingly this test now fails for DPC++, as (at least with CUDA enabled) a default constructed `queue` seems to have a different backend (`ext_oneapi_cuda`) than a default-constructed device (`host`).

tests/event/event.cpp
psalz: Fixed

bader: You seem to miss fixes for 2 typos out of 3.

tests/event/event.cpp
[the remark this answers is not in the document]
bader: Right. DPC++ currently implements SYCL-1.2.1 spec behavior, which says that queue default constructor should use the device returned by the default_selector and device default constructor must return a host device.

tests/event/event.cpp
psalz: Woops, you're right. I blame my spellchecker who seems to think that "paramter" is a word...

tests/event/event.cpp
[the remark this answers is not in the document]
psalz: I see!

ProGTX (approved):

bader:
@keryell, does it look good to you?

tests/common/common.h
```diff
@@ -74,16 +84,22 @@ bool check_return_type_bool(returnT returnVal) {
...
-void check_equal_type(sycl_cts::util::logger& log, actualT actualVal,
-                      std::string logMsg) {
+void check_equal_type(actualT actualVal, std::string logMsg) {
   if (typeid(expectedT) != typeid(actualT)) {
-    FAIL(log, logMsg + "\nGot type -> " +
-                  std::string(typeid(actualT).name()) +
-                  "\nExpected type -> " +
-                  std::string(typeid(expectedT).name()));
+    FAIL(logMsg + "\nGot type -> " + std::string(typeid(actualT).name()) +
+         "\nExpected type -> " + std::string(typeid(expectedT).name()));
```
keryell, line 90: I suspect this could be
```suggestion
    FAIL(logMsg + "\nGot type -> " + typeid(actualT).name() +
         "\nExpected type -> " + typeid(expectedT).name());
```

tests/event/event.cpp
```diff
@@ -0,0 +1,474 @@
...
+   * @param dependencies An optional list of events to depend on.
+   */
+  resolvable_host_event(const std::vector<sycl::event>& dependencies = {}) {
+    event = util::get_cts_object::queue().submit(
+        [this, &dependencies](sycl::handler& cgh) {
+          for (auto& dep : dependencies) {
+            cgh.depends_on(dep);
+          }
+          cgh.host_task([this] {
+            std::unique_lock<std::mutex> lk(mut);
```
keryell, line 138: C++17 CTAD
```suggestion
            std::unique_lock lk{mut};
```

tests/event/event.cpp
```diff
@@ -0,0 +1,474 @@
...
+            cv.wait(lk, [this] { return should_resolve; });
+          });
+        });
+  }
+
+  sycl::event& get_sycl_event() { return event; }
+
+  void resolve() {
+    std::lock_guard<std::mutex> lk(mut);
```
keryell, line 147: C++17 CTAD
```suggestion
    std::lock_guard lk{mut};
```

keryell: For this kind of code I use typically instead of this lock/cv/value a `std::future<void>` with the producer setting the promise (resolve) and the consumer just doing a `.get()`.

tests/event/event.cpp
```diff
@@ -0,0 +1,474 @@
...
+
+/**
+ * A resolvable_host_event that automatically resolves itself after a given
+ * delay.
+ */
+class delayed_host_event : public resolvable_host_event {
+ public:
+  delayed_host_event(std::chrono::milliseconds delay)
+      : resolvable_host_event() {
+    future = std::async(std::launch::async, [this, delay]() {
```
keryell, line 194:
```suggestion
    future = std::async(std::launch::async, [this, delay] {
```

keryell (commented):
Interesting refactoring!

tests/common/common.h
psalz: Yes, or even better, `operator<<`. Although as far as I can tell, this function is currently not being used anywhere, so it should probably just be removed (in favor of `CHECK(std::is_same_v<T, U>)`).

tests/event/event.cpp
psalz: You're right, that is much cleaner. Changed!

fraggamuffin:
ready to merge

keryell (approved):
Thanks!

PR #282, opened 2022-02-11
Add test plan for multi_ptr

kopylovanat:


test_plans/multi_ptr_test_plan.asciidoc
```diff
@@ -0,0 +1,431 @@
...
+`false` for m_ptr constracted from accessor
+
+* `bool operator!=(const multi_ptr& lhs, std::nullptr_t)`
+
+`false` for m_ptr constracted from nullptr_t
+`true` for m_ptr constracted from accessor
+
+* `bool operator<(const multi_ptr& lhs, std::nullptr_t)`
+
+std::less<multi_ptr<value_type, Space, access::decorated::yes>::pointer>()(multi_ptr.get(), nullptr)
```
keryell, line 393: Interesting... Do you think we should check all the these comparisons?
Looking at for example https://stackoverflow.com/questions/65185742/can-you-compare-nullptr-to-other-pointers-for-order-is-it-always-smaller
it is unclear.
Do we say anything in the spec about this?

test_plans/multi_ptr_test_plan.asciidoc
```diff
@@ -0,0 +1,431 @@
...
+For the following values as DecorateAddress parameter:
+
+* `access::decorated::yes`
+* `access::decorated::no`
+
+Constract multi_ptr with accessor to 5th element of array of 10 consecutive values starting with 1 of testing type
+with same template parameters and check result of operators:
+
+* `multi_ptr& operator++(multi_ptr& mp)` refers to `6`
+* `multi_ptr operator++(multi_ptr& mp)` refers to `6`
```
keryell, line 333:
```suggestion
* `multi_ptr operator++(multi_ptr& mp, int)` refers to `5`
```
https://en.cppreference.com/w/cpp/language/operator_incdec
But you need also to check in all these tests taking non-const references that the new value of `mp` and the value returned by the operation.

test_plans/multi_ptr_test_plan.asciidoc
```diff
@@ -0,0 +1,431 @@
...
+* `multi_ptr& operator--(multi_ptr& mp)` refers to `4`
+* `multi_ptr operator--(multi_ptr& mp, int)` refers to `4`
```
keryell, line 335:
```suggestion
* `multi_ptr operator--(multi_ptr& mp, int)` refers to `5`
```

keryell (changes requested):
Great contribution!
There are some points that should be clarified.

test_plans/multi_ptr_test_plan.asciidoc
kopylovanat: SYCL 2020 spec only says about these operators 'Comparison operator OP for multi_ptr class with a std::nullptr_t'
And since we need to check the result somehow I thought we can use std::less etc, similarly how it's described for `unique_ptr` and `std::nullptr_t` https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_cmp

keryell (approved):
Great!

PR #285, opened 2022-02-16
Test for `[[sycl::reqd_work_group_size]]` exception when mismatched `nd_range` was given

andreyromanof:
This PR provides test for optional kernel features. 
Test decorates kernel with attribute `[[sycl::reqd_work_group_size(N)]]` where `N` is supported size and pass `nd_range` instance with unsupported size to `parallel_for` invocation and expect exception `sycl::errc::nd_range`.

### DIFF
--- a/tests/optional_kernel_features/CMakeLists.txt
+++ b/tests/optional_kernel_features/CMakeLists.txt
@@ -1,0 +1,3 @@
+file(GLOB test_cases_list *.cpp)
+
+add_cts_test(${test_cases_list})
--- a/tests/optional_kernel_features/kernel_features_missmatched_nd_range_exception.cpp
+++ b/tests/optional_kernel_features/kernel_features_missmatched_nd_range_exception.cpp
@@ -1,0 +1,100 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for the exception that is thrown by
+//  [[sycl::reqd_work_group_size]] attribute when use nd_range of wrong size in
+//  kernel.
+//
+*******************************************************************************/
+
+#include "../common/common.h"
+#include "catch2/catch_template_test_macros.hpp"
+
+namespace kernel_features_missmathced_nd_range_exception {
+using namespace sycl_cts;
+
+// Define required size of work group for attribute
+constexpr int testing_wg_size = 16;
+
+class Functor {
+ public:
+  [[sycl::reqd_work_group_size(testing_wg_size)]] void operator()(
+      sycl::item<1>) const {}
+};
+
+TEST_CASE(
+    "Kernel features. Exceptions throwns by [[reqd_work_group_size()]] with "
+    "missmatched nd_range",
+    "[kernel_features]") {
+  auto queue = util::get_cts_object::queue();
+  const auto max_wg_size =
+      queue.get_device().get_info<sycl::info::device::max_work_group_size>();
+  if (max_wg_size < testing_wg_size) {
+    WARN("Device supported work group size too small. Skipping..");
+    return;
+  }
+
+  const bool is_exception_expected = true;
+  sycl::errc errc_expected = sycl::errc::nd_range;
+
+  const auto separate_lambda = [=](sycl::item<1>)
+      [[sycl::reqd_work_group_size(testing_wg_size)]]{};
+
+  // Create nd_range that have to cause an exception
+  sycl::nd_range<1> missmathced_nd_rage(sycl::range{max_wg_size + 1},
+                                        sycl::range{max_wg_size + 1});
+
+  {
+    INFO("Task as separate lambda");
+    bool is_exception_thrown = false;
+    try {
+      queue
+          .submit([&](sycl::handler& cgh) {
+            cgh.parallel_for(missmathced_nd_rage, separate_lambda);
+          })
+          .wait_and_throw();
+    } catch (const sycl::exception& e) {
+      is_exception_thrown = true;
+      INFO("Error code check");
+      CHECK(e.code() == errc_expected);
+    }
+    CHECK(is_exception_expected == is_exception_thrown);
+  }
+
+  {
+    INFO("Task as functor");
+    bool is_exception_thrown = false;
+    try {
+      queue
+          .submit([&](sycl::handler& cgh) {
+            cgh.parallel_for(missmathced_nd_rage, Functor{});
+          })
+          .wait_and_throw();
+    } catch (const sycl::exception& e) {
+      is_exception_thrown = true;
+      INFO("Error code check");
+      CHECK(e.code() == errc_expected);
+    }
+    CHECK(is_exception_expected == is_exception_thrown);
+  }
+
+  {
+    INFO("Task as submission call");
+    bool is_exception_thrown = false;
+    try {
+      queue
+          .submit([&](sycl::handler& cgh) {
+            cgh.parallel_for(missmathced_nd_rage, [
+            ](sycl::item<1>) [[sycl::reqd_work_group_size(testing_wg_size)]]{});
+          })
+          .wait_and_throw();
+    } catch (const sycl::exception& e) {
+      is_exception_thrown = true;
+      INFO("Error code check");
+      CHECK(e.code() == errc_expected);
+    }
+    CHECK(is_exception_expected == is_exception_thrown);
+  }
+}
+}  // namespace kernel_features_missmathced_nd_range_exception


ci/computecpp.filter
```diff
@@ -28,3 +28,4 @@ usm
 vector_api
 vector_constructors
 vector_swizzles
+optional_kernel_features
```
ProGTX, line 31: Do you maybe have output from a run of ComputeCpp or is this just to play it safe and not risk breaking CI?

ci/hipsycl.filter
```diff
@@ -37,3 +37,4 @@ vector_load_store
 vector_operators
 vector_swizzle_assignment
 vector_swizzles
+optional_kernel_features
```
ProGTX, line 40: I think it's better to maintain an alphabetical ordering, the same for other filters. Also, should we maybe name it `kernel_optional_features`?

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
```diff
@@ -0,0 +1,100 @@
...
+  auto queue = util::get_cts_object::queue();
+  const auto max_wg_size =
+      queue.get_device().get_info<sycl::info::device::max_work_group_size>();
+  if (max_wg_size < testing_wg_size) {
+    WARN("Device supported work group size too small. Skipping..");
+    return;
+  }
+
+  const bool is_exception_expected = true;
+  sycl::errc errc_expected = sycl::errc::nd_range;
```
ProGTX, line 39: From the spec:
> In order to guarantee source code portability of SYCL applications that use optional kernel features, all SYCL implementations must be able to compile device code that uses these optional features regardless of whether the implementation supports the features on any of its devices.

You've marked two implementations in the CI as not passing (adding the test to the filter), but I believe they should be able to pass the test even if they don't implement the features, otherwise they wouldn't be able to pass conformance in the current CTS design.

ProGTX (changes requested):
Some concerns about disabling the test for implementations that don't support these features.

ci/computecpp.filter
andreyromanof: [Here](https://github.com/KhronosGroup/SYCL-CTS/runs/5209986322?check_suite_focus=true) output from a run of ComputeCpp before I apply filter. On line 2791 warning messages about unknown attributes, which is why I decided to put test in the filter.

But right now I also noticed other errors like `no known conversion from 'cl::sycl::nd_item<1>' to 'sycl::item<1>'` on line 2814. Am I wrong with the lambda argument? Do I have to use `nd_item<1>` instead of `item<1>` on line 41 in `kernel_features_mismatched_nd_range_exception.cpp`?

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
andreyromanof: As I understand CI doesn't run tests, it's just built it. [Here](https://github.com/KhronosGroup/SYCL-CTS/runs/5209986361?check_suite_focus=true) you can find output from the building of tests on CI. ComputeCpp and hipsycl failed building due to some errors.

From hipsycl logs, line 1970: `error: ‘errc’ is not a member of ‘sycl’`. Looks like enum with error codes is not implemented yet and attributes too (line 1966).

Similar situation for computecpp.

But maybe I was wrong when I put these two implementations in the filter because both of them give the error message: `no known conversion from 'cl::sycl::nd_item<1>' to 'sycl::item<1>'` error, so it can be bug of tests itself.

ci/hipsycl.filter
andreyromanof: Sorted lines in this commit: acc4949b236617eb437398836aecf32b1c4f102a.

We can rename it but in other PR when all tests will be implemented, because some of the tests in other PR, some of the tests in development and it will be unhandy to rename these tests now.

IMHO, I think we can keep the current name.

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
andreyromanof: It was my mistake. I changed to `nd_item` but now getting another error. Right now investigating it.

andreyromanof:
I tried to fix the code for the computecpp CI but looks like this implementation has a few specific issues:
* It's doesn't support implementation-defined kernel names according to paragraph [4.9.4.2](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:invokingkernels). [Link to the error](https://github.com/KhronosGroup/SYCL-CTS/runs/5243793167?check_suite_focus=true#step:8:2800)
* I changed `const auto max_wg_size` to `const size_t max_wg_size`, because otherwise compilation was failed. This variable is used in the constructor of `nd_range` [Link to error](https://github.com/KhronosGroup/SYCL-CTS/runs/5243923322?check_suite_focus=true#step:8:2821)
* [Problem with linkage](https://github.com/KhronosGroup/SYCL-CTS/runs/5244251686?check_suite_focus=true#step:8:12447) I was unable to solve, but my guess that its kind of stringify function for the `sycl::errc` that cause an error.

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+namespace kernel_features_mismatched_nd_range_exception {
+using namespace sycl_cts;
+
+struct kernel_separate_lambda;
+struct kernel_functor;
+struct kernel_submission_call;
+
+// Define required size of work group for attribute
+constexpr int testing_wg_size = 16;
```
psalz, line 22: Why not test with a work group size of 1? This should be supported by every device!

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+class Functor {
+ public:
+  [[sycl::reqd_work_group_size(testing_wg_size)]] void operator()(
+      sycl::nd_item<1>) const {}
+};
+
+TEST_CASE(
+    "Kernel features. Exceptions throwns by [[reqd_work_group_size()]] with "
+    "mismatched nd_range",
```
psalz, line 32: Suggest rewording to something like `Submitting a kernel with an nd-range that does not match [[reqd_work_group_size()]] throws an exception`.

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+  const auto separate_lambda = [=](sycl::nd_item<1>)
+      [[sycl::reqd_work_group_size(testing_wg_size)]]{};
+
+  // Create nd_range that have to cause an exception
+  sycl::nd_range<1> mismatched_nd_rage(sycl::range(max_wg_size + 1),
+                                       sycl::range(max_wg_size + 1));
+
+  {
+    INFO("Task as separate lambda");
```
psalz, line 53: Why not use `SECTION`s here instead?

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
```diff
@@ -0,0 +1,107 @@
...
+      queue
+          .submit([&](sycl::handler& cgh) {
+            cgh.parallel_for<kernel_separate_lambda>(mismatched_nd_rage,
+                                                     separate_lambda);
+          })
+          .wait_and_throw();
+    } catch (const sycl::exception& e) {
+      is_exception_thrown = true;
+      INFO("Error code check");
+      CHECK(e.code() == errc_expected);
```
psalz, line 65: These checks could be replaced with `CHECK_THROWS_MATCHES` introduced by @yuriykoch in #274!

psalz (changes requested):

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
andreyromanof: I was thinking that it is not necessary overhead with reconstructing context like `queue`, `max_wg_size`, `mismatched_nd_rage` and etc. but maybe you're right and overhead is too small to think about it.

Changed to `SECTION`

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
andreyromanof: Looks like #274 is still not merged. 
I planned to patch these tests after `SKIP` will be introduced in `catch2` https://github.com/catchorg/Catch2/pull/2360. I think Yuriy's patch will be merged by that time so I will replace this part of code with `CHECK_THROWS_MATCHES`.

psalz: Okay sure. `SKIP` will still take a while though, I'm afraid.

psalz (approved):

tests/optional_kernel_features/kernel_features_mismatched_nd_range_exception.cpp
andreyromanof: Because tag doesn't appear in output, when test fails. By adding this prefix I wanted to make test output more clear for enduser.

psalz: I see. IMO it's still a bit redundant because the file name is listed directly below the failing test case description:
image

If clarity is a major concern we could also investigate adding a custom console reporter that groups test cases more visibly into their respective categories.

andreyromanof: Okay. I thought we talked about tag `[kernel_features]` from the second argument. Looks like I misunderstood you. 
I agree, it looks redundant. I will remove the prefix from the test name.

keryell (approved):
Thanks.

andreyromanof:
Build error on `ComputeCpp` points to `SECTION` invocation and `use of undeclared identifier INTERNAL_CTS_ENABLED_TEST_CASE_BODY`. Don't have any more ideas on how to fix it and seems like it is not a compiler bug, so we can't just add the test to filter.

psalz:
> Build error on `ComputeCpp` points to `SECTION` invocation and `use of undeclared identifier INTERNAL_CTS_ENABLED_TEST_CASE_BODY`. Don't have any more ideas on how to fix it and seems like it is not a compiler bug, so we can't just add the test to filter.

Actually the root cause is the [error before that](https://github.com/KhronosGroup/SYCL-CTS/runs/5355235297?check_suite_focus=true#step:8:2807): `error: _Pragma takes a parenthesized string literal`. I've actually seen this before and already have a fix for it. I'll open a PR shortly. This will however not really solve the problem at hand, as ComputeCpp then still fails during linking with `undefined reference to `cl::sycl::detail::errc_to_str[abi:cxx11](cl::sycl::detail::errc)'` (cc @ProGTX).

psalz:
The aforementioned bug should be fixed by #296.

keryell (approved):

bader:
@andreyromanof, please, fix the build with computecpp.

bader:
@ProGTX, please, check if your concerns are addressed.

ProGTX (approved):

PR #286, opened 2022-02-16
Fix tests for sub_group_mask

kopylovanat:


ProGTX (approved):

tests/extension/oneapi_sub_group_mask/sub_group_mask_flip_api.cpp
```diff
@@ -18,11 +18,16 @@ using namespace sycl_cts;
 struct check_result_flip {
   bool operator()(sycl::ext::oneapi::sub_group_mask &sub_group_mask,
                   const sycl::sub_group &sub_group) {
-    unsigned long before_flip, after_flip;
-    sub_group_mask.extract_bits(before_flip);
+    // since all bits need to be checked and size of sub_group_mask is not
+    // known beforehand we can't use fixed type to extract bits and check them
+    // after the flip
```
bader, line 23: I disagree with the statement and suggest removing or adjusting this comment.
NOTE: there is no such a comment in tests/extension/oneapi_sub_group_mask/sub_group_mask_set_api.cpp

I think something like this should work:
```c++
    unsigned long before_flip, after_flip;
    sub_group_mask.extract_bits(before_flip);
    sub_group_mask.flip();
    sub_group_mask.extract_bits(after_flip);
    // mask off irrelevant bits
    unsigned long mask = ULONG_MAX << sub_group_mask.size();
    after_flip = after_flip & mask;
    before_flip = before_flip | ~mask;
    return after_flip == ~before_flip;
```
`unsigned long` size is not "fixed", but the same can be done with any fixed type as long as it can hold all `sub_group_mask` bits.

I guess the performance of this part is not critical, so I'm okay to have a loop as it's easier to read.

tests/extension/oneapi_sub_group_mask/sub_group_mask_insert_bits.cpp
```diff
@@ -34,7 +34,7 @@ void get_bits(sycl::marray<T, nElements> &out) {
 
 template <typename T>
 struct check_result_insert_bits {
-  bool operator()(const sycl::ext::oneapi::sub_group_mask &sub_group_mask,
```
bader, line 37: I think we should keep `const` here or pass by value. This operator is supposed to **check** mask properties and it might lead to unexpected bugs if it changes it.
Please, apply to all checkers.

kopylovanat: Changed to pass by value.

tests/extension/oneapi_sub_group_mask/sub_group_mask_flip_api.cpp
```diff
@@ -16,18 +16,21 @@ using namespace sycl_cts;
...
                   const sycl::sub_group &sub_group) {
     unsigned long before_flip, after_flip;
     sub_group_mask.extract_bits(before_flip);
     sub_group_mask.flip();
     sub_group_mask.extract_bits(after_flip);
-    return after_flip == ~before_flip;
+    // mask off irrelevant bits
+    unsigned long mask =
+        ULONG_MAX >> (CHAR_BIT * sizeof(unsigned long) - sub_group_mask.size());
+    return ((after_flip & before_flip) & mask) == 0;
```
bader, line 28: I'm fine this code as well, but it can be improved.

`CHAR_BIT * sizeof(unsigned long) - sub_group_mask.size()` <- I think we need to make sure that there are no UB in this expression.

> if the value of the right operand is negative or is greater or equal to the number of bits in the promoted left operand, the behavior is undefined

If I get it right, it means `0 < sub_group_mask.size() <= CHAR_BIT * sizeof(unsigned long)` should hold.

Side question: why do we use `unsigned long` here? The spec says:
> an integral type or a SYCL marray of integral types

How do we know that `sub_group_mask` fits into `unsigned long` type?

kopylovanat: > How do we know that `sub_group_mask` fits into `unsigned long` type?

`globalSize` is 32 in `sub_group_mask_common.h` so sub_group_mask's size is not expected to be bigger than that and `unsigned long`'s size is at least 32. Yes, we can use `exctract_bits()` with bigger types but it seems to be excessive here. 
That's also why it's expected `sub_group_mask.size() <= CHAR_BIT * sizeof(unsigned long)` to be true. Do you think that the test should check if it is true and, for example, return `false` if it's not?

bader: If `sub_group_mask.size() > CHAR_BIT * sizeof(unsigned long)` the test has UB (shift operand is negative) and we can't rely on the results.

kopylovanat: Added check to rule out UB.

bader: Where do we check `sub_group_mask.size() != 0`?

kopylovanat: Added check for `sub_group_mask.size() != 0`

tests/extension/oneapi_sub_group_mask/sub_group_mask_set_api.cpp
```diff
@@ -16,18 +16,25 @@ using namespace sycl_cts;
 #ifdef SYCL_EXT_ONEAPI_SUB_GROUP_MASK
 
 struct check_result_set {
-  bool operator()(sycl::ext::oneapi::sub_group_mask &sub_group_mask,
+  bool operator()(sycl::ext::oneapi::sub_group_mask sub_group_mask,
                   const sycl::sub_group &sub_group) {
+    int shift = CHAR_BIT * sizeof(unsigned long) - sub_group_mask.size();
+    // sub_group_mask's size is expected to be greater than 0 and not to be
+    // greater than 32 in these tests but check is added to make sure of it
+    if (shift < 0 || sub_group_mask.size() == 0) return false;
```
bader, line 24: To be pedantic till the end, then the comment doesn't correspond the code. `shift < 0` doesn't match "sub_group_mask's size is expected not to be greater than 32" if sizeof(unsigned long) > 4. There are platforms where sizeof(unsigled long) == 8.

It might be easier to make the check straightforward and do exactly what comment says before computing the `shift` value. In addition to that, I would consider rephrasing the comment to something like "sub_group_mask's size must be in the range between 0 (excluded) and 32 (included)".

kopylovanat: Changed to straightforward check and changed comment.

tests/extension/oneapi_sub_group_mask/sub_group_mask_flip_api.cpp
```diff
@@ -16,18 +16,25 @@ using namespace sycl_cts;
 #ifdef SYCL_EXT_ONEAPI_SUB_GROUP_MASK
 
 struct check_result_flip {
-  bool operator()(sycl::ext::oneapi::sub_group_mask &sub_group_mask,
+  bool operator()(sycl::ext::oneapi::sub_group_mask sub_group_mask,
                   const sycl::sub_group &sub_group) {
+    int shift = CHAR_BIT * sizeof(unsigned long) - sub_group_mask.size();
```
bader, line 21:
```suggestion
```
Dead code?

kopylovanat: Yes, sorry, removed.

bader (approved):

PR #287, opened 2022-02-17
Tests for `[[device_has]]` attribute from optional kernel features 

andreyromanof:
This PR provides tests for optional kernel features.
Test decorates kernel or function invoked by the kernel with the attribute `[[sycl::device_has(aspect)]]` and expected exception if a device doesn't support aspect, otherwise expect no throw. 

Depends on:
- https://github.com/KhronosGroup/SYCL-CTS/pull/296


tests/optional_kernel_features/kernel_features_separate_unit.cpp
```diff
@@ -0,0 +1,43 @@
...
+  float temp = 42.25f;
+  double feature(temp);
+  feature += 42;
+}
+
+using AtomicRefT =
+    sycl::atomic_ref<unsigned long long, sycl::memory_order::relaxed,
+                     sycl::memory_scope::device>;
+template <>
+SYCL_EXTERNAL void
```
ProGTX, line 37: Should an implementation define this macro even if it doesn't support the functionality? If not (and the spec suggests so https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_linker), this test will not pass on that implementation, so it should be made optional somehow. Related to the discussion in https://github.com/KhronosGroup/SYCL-CTS/pull/285#pullrequestreview-885634602 .

ProGTX (approved):

tests/optional_kernel_features/kernel_features_separate_unit.cpp
bader: @andreyromanof, can we remove optional_kernel_features test from all the filters?

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  SECTION(
+      "Kernel with tested feature but with attribute [[sycl::device_has()]] "
+      "for another feature.") {
+    constexpr sycl::aspect AnotherFeatureAspect =
+        get_another_aspect<FeatureAspectT>();
+    bool other_feature_exception_expect = true;
+    if (queue.device_has(AnotherFeatureAspect) &&
+        queue.device_has(FeatureAspectT)) {
```
bader, line 242: @andreyromanof, is this a DPC++ extension?

bader (commented):
@andreyromanof, could you extend the description of the PR with the information about what prevent the compilation of this test with computcpp/dpcpp/hipsycl?

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
andreyromanof: No. As far as I know, it is not an extension.

andreyromanof:
> @andreyromanof, could you extend the description of the PR with the information about what prevent the compilation of this test with computcpp/dpcpp/hipsycl?

Done

tests/optional_kernel_features/kernel_features_separate_unit.cpp
andreyromanof: I think right now we cant remove these tests, due to issues that I mentioned in the description of the PR.

keryell (commented):
You could remove some useless `()`

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
bader: I don't see such member function in the specification, so it seems to be illegal. Rigth?

andreyromanof: You're right. There is only `has(aspect)` member function of `device` class. Looks like I have to change all `queue.device_has(..)` to `queue.get_device().has(..)`

bader (changes requested):
> * Build fails on computecpp due to [linkage error](https://github.com/KhronosGroup/SYCL-CTS/runs/5244251686?check_suite_focus=true#step:8:12447), which I was unable to solve.

@andreyromanof, the link points to the log for a different PR. According to @psalz, there are at least two issues with compiling this PR with computecpp:
- tests use unnamed lambda
- tests use "device_has"

I expected you to list both in the description. 

I suggest we remove optional_kernel_features from all the filters and disable specific test cases with the macro introduced by https://github.com/KhronosGroup/SYCL-CTS/pull/293.

tests/optional_kernel_features/kernel_features_separate_unit.cpp
```diff
@@ -0,0 +1,45 @@
...
+  float temp = 42.25f;
+  sycl::half feature(temp);
+  feature += 42;
+}
+
+template <>
+SYCL_EXTERNAL void
+use_feature_function_external_decorated<double, sycl::aspect::fp64>() {
+  float temp = 42.25f;
+  double feature(temp);
```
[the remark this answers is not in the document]
andreyromanof: I don't think I understood you. Suggested change doesn't change anything

andreyromanof:
> I suggest we remove optional_kernel_features from all the filters and disable specific test cases with the macro introduced by https://github.com/KhronosGroup/SYCL-CTS/pull/293.

That makes sense. I'm going to patch this test and tests from other PRs.

andreyromanof:
But even if I apply macro from #293, computecpp and hipsycl will still break CI

tests/optional_kernel_features/kernel_features_separate_unit.cpp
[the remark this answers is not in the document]
andreyromanof: I don't think I understood you. Suggested change doesn't change anything

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  // Set expected error code
+  constexpr sycl::errc expected_errc = sycl::errc::kernel_not_supported;
+
+  SECTION(
+      "Kernel that uses the tested feature but does not have any attribute "
+      "[[sycl::device_has()]]") {
+    {
+      const auto lambda_no_arg = []() { USE_FEATURE(FeatureTypeT); };
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+                          NO_ATTRIBUTE, USE_FEATURE(FeatureTypeT));
+    }
+  }
+
+  SECTION(
+      "Kernel that calls a function that uses the tested feature. Neither "
+      "the kernel nor the function have an attribute "
+      "[[sycl::device_has()]].") {
+    {
+      const auto lambda_no_arg = []() {
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  #ifdef SYCL_EXTERNAL
+  SECTION(
+      "Kernel does not have the attribute [[sycl::device_has()]] but it "
+      "calls a SYCL_EXTERNAL function which uses the tested feature. The "
+      "SYCL_EXTERNAL function is defined in another translation unit. The "
+      "SYCL_EXTERNAL function is declared with the corresponding attribute "
+      "[[sycl::device_has()]].") {
+    {
+      const auto lambda_no_arg = []() {
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+                                                  FeatureAspectT>());
+    }
+  }
+  #endif
+
+  SECTION(
+      "Kernel does not use the tested feature but is decorated with the "
+      "corresponding attribute [[sycl::device_has()]].") {
+    {
+      const auto lambda_no_arg = []() [[sycl::device_has(FeatureAspectT)]] {
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+                          dummy_function_non_decorated());
+    }
+  }
+
+  SECTION(
+      "Kernel that calls a function which is decorated with the feature's "
+      "corresponding attribute [[sycl::device_has()]]. Neither the kernel "
+      "nor the function use the feature.") {
+    {
+      const auto lambda_no_arg = []() {
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+                          dummy_function_decorated<FeatureAspectT>());
+    }
+  }
+
+  SECTION(
+      "Kernel that calls a function which is decorated with the feature's "
+      "corresponding attribute [[sycl::device_has()]]. Function uses the "
+      "feature and kernel doesn't.") {
+    {
+      const auto lambda_no_arg = []() {
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
+        get_another_aspect<FeatureAspectT>();
+    bool other_feature_exception_expect = true;
+    if (queue.device_has(AnotherFeatureAspect) &&
+        queue.device_has(FeatureAspectT)) {
+      other_feature_exception_expect = false;
+    }
+
+    {
+      const auto lambda_no_arg =
+          []() [[sycl::device_has(AnotherFeatureAspect)]] {
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,329 @@
...
```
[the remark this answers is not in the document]
andreyromanof: Removed

tests/optional_kernel_features/kernel_features_separate_unit.cpp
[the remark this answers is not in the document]
keryell: Sure. This is completely useless but as you did with `sycl::half` you have a similar code qith some computation with the `+=`, not only a construction  of the type.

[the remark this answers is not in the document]
andreyromanof: Looks like there is misunderstanding again) 
By "doesn't change anything" I meant, that suggested code already in the source file. [Here](https://github.com/KhronosGroup/SYCL-CTS/blob/b90a11e64f1688617ee632cb4285b16a908c8c63/tests/optional_kernel_features/kernel_features_separate_unit.cpp#L31) is the link to the line where operator `+=` is used with `double`

andreyromanof:
I added compile time disable of the test and removed this test from the filter.
I also noticed, that for some reason I should pass implementation names in strict order. For example
- `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, DPCPP, hipSYCL)` build fails due strange errors
- `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, hipSYCL, DPCPP)` build finishes with no errors

@psalz could you please verify is this a bug or not.

andreyromanof:
Build still fails hipSYCL because there is no `sycl::errc` definition.

psalz:
> I added compile time disable of the test and removed this test from the filter. I also noticed, that for some reason I should pass implementation names in strict order. For example
> 
>     * `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, DPCPP, hipSYCL)` build fails due strange errors
> 
>     * `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, hipSYCL, DPCPP)` build finishes with no errors
> 
> @psalz could you please verify is this a bug or not.

Yes same bug as in #289, fixed in #295.

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    {
+      RUN_SUBMISSION_CALL(is_exception_expected, expected_errc, queue,
+                          [[sycl::device_has(FeatureAspectT)]],
+                          dummy_function_non_decorated());
+    }
+  }
+
+  SECTION(
```
bader, line 171: Does it make sense to move "sections" to separate "test cases"? Will it enable compilation of some test cases with any of implementations?

ci/hipsycl.filter
```diff
@@ -21,6 +21,7 @@ math_builtin_api
 multi_ptr
 nd_item
 nd_range
+optional_kernel_features
```
bader, line 24:
```suggestion

```

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
```diff
@@ -0,0 +1,330 @@
...
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for the exception that are thrown by [[sycl::device_has]]
+//  attribute.
+//
+*******************************************************************************/
+
+#include "../common/disabled_for_test_case.h"
+#include "catch2/catch_template_test_macros.hpp"
+#include "kernel_features_common.h"
```
bader, line 12:
```suggestion
// FIXME: re-enable compilation with hipSYCL when `sycl::errc` is supported
#ifndef __HIPSYCL__
#include "kernel_features_common.h"
#endif
```

bader (commented):
Will this work?

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
andreyromanof: I think it makes sense. 
This will allow us to enable most of the tests from this PR for DPCPP, but not for other implementations. 
ComputeCpp still has an issue with linkage (Philip mentioned that [here](https://github.com/KhronosGroup/SYCL-CTS/pull/285#issuecomment-1054035325)) and it does not support unnamed kernels yet.
hipSYCL does not have a definition of `sycl::errc`.
So it will not enable compilation of some tests with any implementation.

Anyway I spilt sections into different test cases in this commit:  https://github.com/KhronosGroup/SYCL-CTS/pull/287/commits/e297e1e06527e7bcf1a477eeb46e2deb5edaa383

tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
andreyromanof: It will work. Applied this in https://github.com/KhronosGroup/SYCL-CTS/pull/287/commits/85c27137801e86d89a7e9f3eec2fd1d7524716da

bader:
@andreyromanof, does merging https://github.com/KhronosGroup/SYCL-CTS/pull/299 and https://github.com/KhronosGroup/SYCL-CTS/pull/296 fix pre-commit?

andreyromanof:
Can't say for sure. 
#296 should fix build for computecpp which I can't test locally. Also, Philip [mentioned a bug](https://github.com/KhronosGroup/SYCL-CTS/pull/285#issuecomment-1054035325) with linkage even after with fix. So probably we will need to update computecpp version too.
#299 Should fix ci build for dpcpp.

tests/optional_kernel_features/kernel_features_separate_unit.cpp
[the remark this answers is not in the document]
keryell: Ah, good. :-) Thanks for the clarification.

tests/optional_kernel_features/kernel_features_common.h
```diff
@@ -0,0 +1,519 @@
...
+ */
+void dummy_function_non_decorated() {
+  int var = 0;
+  var += 42;
+}
+
+/**
+ * @brief Macro for generating code that will use TYPE
+ */
+#define USE_FEATURE(TYPE)       \
```
keryell, line 80: Not clear why you are using a macro instead of a generic function or lambda.

tests/optional_kernel_features/kernel_features_common.h
```diff
@@ -0,0 +1,519 @@
...
+      } catch (const sycl::exception &e) {                                     \
+        is_exception_thrown = true;                                            \
+        if (IS_EXCEPTION_EXPECTED) {                                           \
+          INFO("Error code check");                                            \
+          CHECK(e.code() == ERRC);                                             \
+        }                                                                      \
+      }                                                                        \
+      CHECK(is_exception_thrown == IS_EXCEPTION_EXPECTED);                     \
+      check_async_exception(QUEUE, false);                                     \
+    }                                                                          \
```
keryell, line 514: Why such a huge macro instead of a function/class/lambda? Is this to pass the attribute with the preprocessor?
Also a lot of copy-paste inside the macro itself.

tests/optional_kernel_features/kernel_features_common.h
```diff
@@ -0,0 +1,519 @@
...
+      is_exception_thrown = false;
+    } catch (const sycl::exception &e) {
+      is_exception_thrown = true;
+      if (is_exception_expected) {
+        INFO("Error code check");
+        CHECK(e.code() == errc_expected);
+      }
+    }
+    CHECK(is_exception_thrown == is_exception_expected);
+    check_async_exception(queue, false);
```
keryell, line 346: I have the feeling to mostly see 3 times the same code.
Factorize in a function/lambda called 3 times and taking a lambda for executing the different code?

tests/optional_kernel_features/kernel_features_common.h
```diff
@@ -0,0 +1,519 @@
...
```
keryell, line 430: Idem: mostly 3 copy-pastes.

tests/optional_kernel_features/kernel_features_common.h
andreyromanof: You're right, it looks like I can optimize these places. I tried to refactor these functions in this commit: 2fb2d5e4004329d31498a25c27355e0fc3c59dd7

tests/optional_kernel_features/kernel_features_common.h
andreyromanof: I used macro because in some cases kernel should be sent as a submission call (without storing it in separate lambda or class functor for example). To achieve that I wrote this huge macro, that should paste code with submission call.

This macro also refactored in this commit: 2fb2d5e4004329d31498a25c27355e0fc3c59dd7

andreyromanof:
@bader Right now computecpp build fails, because there are using unnamed kernels in the code.  [Spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:invokingkernels) says that it's allowed. Should we use a workaround with the named kernel to avoid this build issue, or these tests can just wait in the filter until feature with automatic kernel names will be implemented in computecpp? Please note, that unnamed kernels are also used in these PRs: #292, #294, #289. 

Also, computecpp will not compile even with named kernels due to a linkage error that looks like an implementation bug. (This error occurs in [other PR](https://github.com/KhronosGroup/SYCL-CTS/pull/285) where named kernels are used. [Link to linkage error in other PR](https://github.com/KhronosGroup/SYCL-CTS/runs/5430550063?check_suite_focus=true#step:8:12613))

andreyromanof:
@psalz I don't understand why code in `kernel_features_device_has_exceptions.cpp` still compiling on computecpp, despite the fact that I disable implementation in every test with `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, hipSYCL)`. Can you suggest me what I doing wrong? :)

psalz:
> @psalz I don't understand why code in `kernel_features_device_has_exceptions.cpp` still compiling on computecpp, despite the fact that I disable implementation in every test with `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, hipSYCL)`. Can you, please, suggest to me what I doing wrong? :)

Fixed in #303.

andreyromanof:
> Fixed in https://github.com/KhronosGroup/SYCL-CTS/pull/303.

Thanks.

bader (approved):

PR #291, opened 2022-02-21
Fix and update tests for extension oneapi_properties

kopylovanat:
Since FAIL can't be used in constexpr function.


bader (approved):

ProGTX (approved):

keryell (approved):
Thanks.

tests/extension/oneapi_property_list/property_list_prop_eq_op.cpp
```diff
@@ -15,8 +15,7 @@ namespace TEST_NAMESPACE {
 using namespace sycl_cts;
 
 template <typename T, typename U>
-constexpr void check_equal(util::logger &log, T prop1, U prop2,
-                           bool expected_equal) {
+void check_equal(util::logger &log, T prop1, U prop2, bool expected_equal) {
```
psalz, line 18: Given that we have Catch2 now, wouldn't it be more readable to instead simply write something like

```c++
CHECK(prop_value_device_image_scope == prop_value_device_image_scope);
CHECK_FALSE(prop_value_device_image_scope != prop_value_device_image_scope);
```

?

kopylovanat: Done in https://github.com/KhronosGroup/SYCL-CTS/pull/291/commits/3c74460d5577cce2bb4e3f74e36de58e6e9b5c2d

kopylovanat:
Update tests with extension spec changes

bader:
I think we should disregard all approves as new version of the patch is completely different from the approved one.

keryell (approved):
Thanks!

PR #293, opened 2022-02-24
Add support for compile-time disabled test cases

psalz:
This introduces a new type of test case macro that can be used to disable test cases during compile-time. This is useful for when a test case contains code that  currently does not compile for a given implementation, while other test cases in the same translation unit would otherwise compile.

A disabled test case will fail automatically at runtime.

While all Catch2 test macros could likely be supported in principle, this initial implementation is limited to these two:

 - `DISABLED_FOR_TEST_CASE` (corresponding to `TEST_CASE`)
 - `DISABLED_FOR_TEMPLATE_TEST_CASE_SIG` (corresponding to `TEMPLATE_TEST_CASE_SIG`)

Usage example from `id` tests:
```c++
DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(hipSYCL)
("id supports unary +/- operators", "[id]", ((int D), D), 1, 2, 3)({
  // ...
});
```

which then produces the following console output:

image

### DIFF
--- a/tests/common/disabled_for_test_case.h
+++ b/tests/common/disabled_for_test_case.h
@@ -1,0 +1,138 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2022 The Khronos Group Inc.
+//
+*******************************************************************************/
+
+#ifndef __SYCLCTS_TESTS_COMMON_DISABLED_FOR_TEST_CASE_H
+#define __SYCLCTS_TESTS_COMMON_DISABLED_FOR_TEST_CASE_H
+
+#include "macro_utils.h"
+
+// TODO: Add other Catch2 test case variants, as needed
+
+/**
+ * Registers a test case that is compile-time disabled for one or more SYCL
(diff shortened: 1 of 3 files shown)


tests/common/disabled_for_test_case.h
```diff
@@ -0,0 +1,138 @@
...
+#define INTERNAL_CTS_IS_IMPL(impl) \
+  INTERNAL_CTS_CHECK(INTERNAL_CTS_IMPL_PROBE(impl))
+
+// ------------------------------------------------------------------------------------
+
+// Helper macros for expanding recursive calls.
+// This allows us to check for up to 16 + 1 SYCL implementations, which should
+// be plenty. (Otherwise this could always be extended by another factor of
+// two).
+#define INTERNAL_CTS_EVAL(...) INTERNAL_CTS_EVAL16(__VA_ARGS__)
```
keryell, line 72: Perhaps we could use Boost.Preprocessor instead?

keryell (approved):
Thanks.

bader (approved):

PR #298, opened 2022-03-01
Bump dpcpp version in ci

andreyromanof:
This will allow us to enable most of the tests from https://github.com/KhronosGroup/SYCL-CTS/pull/287, as the current version of the compiler doesn't support using `[[sycl::device_has()]]` with lambdas

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -55,7 +55,7 @@ jobs:
           # - sycl-impl: computecpp
           # version: 2.8.0
           - sycl-impl: dpcpp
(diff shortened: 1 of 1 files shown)


andreyromanof:
@bader Am I allowed to change the version of the dpcpp compiler?

psalz:
The PR needs to originate from this repo unfortunately, otherwise the necessary build secrets for accessing DockerHub aren't available.

bader:
@andreyromanof, I've moved your commits to the local branch in https://github.com/KhronosGroup/SYCL-CTS/pull/299.

PR #301, opened 2022-03-02
Fix unsigned wrap in `id` tests

yuriykoch:
The `size_t` doesn't have guarantee in size compared to `unsigned long` or `unsigned long long`.
Therefore we should either explicitly cast to `size_t` or use implicit cast.

Implicit cast fits better, because it uses the type of `get_cts_object::id<>::get()` parameters, avoiding possible support efforts in future.


keryell (approved):
Thanks.

PR #303, opened 2022-03-07
Include SYCL header for DISABLED_FOR_* implementation detection

psalz:
This may not be needed for all implementations, but at least ComputeCpp defines `__COMPUTECPP__` inside a header, instead of on the command line.


bader:
Maybe we can set some macro ourselves in CMake instead and not rely on implementation specific ones?
We detect implementations in cmake/Find*.cmake files and we can set the macro for each compiler.

psalz:
> Maybe we can set some macro ourselves in CMake instead and not rely on implementation specific ones? We detect implementations in cmake/Find*.cmake files and we can set the macro for each compiler.

Yes, I'm still planning on doing that, but I'd like to then also update all locations that currently check for `__HIPSYCL__` et al. So that I'll do that as a separate change.

bader (approved):

PR #304, opened 2022-03-10
Add first draft of developer documentation

psalz:
As discussed in the WG, we hope to offer a more streamlined experience for new developers wanting to contribute to the CTS. I've started to draft some stuff, but it is still very bare-bones. Feedback is welcome!


docs/writing_tests.md
```diff
@@ -0,0 +1,156 @@
...
+
+**TODO**
+
+### Generators
+
+**TODO**
+
+## Special Macros & Custom Matchers
+
+**TODO**
```
andreyromanof, line 94: I don't understand why do we need to describe `cathc2` macros in our doc. Can't we just leave a link to the `catch2` documentation? 

Also, I think it is more important to describe macros that were implemented only in our repo. For example `DISABLE_FOR_TEST_CASE` and `DISABLE_FOR_TEMPLATE_TEST_CASE_SIG`

andreyromanof (approved):

docs/writing_tests.md
```diff
@@ -0,0 +1,156 @@
...
+  Use `CHECK(condition)` instead.
+- Keep test cases small and focused to a single concept / behavior.
+  Even a single function could be tested with several test cases.
+- Use natural language descriptions for test cases.
+    - Avoid: `"host_accessor range mismatch exception"`.
+    - Prefer: `"host_accessors throws if accessed range exceeds buffer dimensions"`.
+- Tag test cases according to the feature being tested:
+    - Use `[some_type]` for types that exist in the SYCL specification (example: `[host_accessor]`).
+    - Use `[some-concept]` for concepts without a clearly associated type (example: `[backend-interop]`).
+- Try to write test cases in the same order as the associated API specification (if possible).
```
gmlueck, line 112: I think it would be good to provide some guidance on how many test cases to include in a single source file.  For example, is it appropriate to have a single source file that has dozens of test cases for a single SYCL API (e.g. there will probably be a lot of test cases for `accessor`)?  Or, is it more appropriate to create a test category for `accessor` and have many source files?  If this is the case, should each source file have just one test case, or should related ones be grouped into the same source file?

docs/writing_tests.md
```diff
@@ -0,0 +1,156 @@
...
+      }
+  }
+};
+
+util::test_proxy<TEST_NAME> proxy;
+}
+```
+
+While legacy test cases are still mapped to Catch2 under the hood, they require a lot of boilerplate code and therefore testing logic for distinct aspects of a feature are often grouped into a single test case, making them harder to comprehend and debug.
+Although technically not required, usually only one class extending `test_base` is defined per file.
```
gmlueck, line 154: I'm not sure what your plans are for documenting the legacy style.  I would suggest that you do not spend much time documenting it, though.  We want people to use the Catch2 style, not the legacy one.

gmlueck (commented):
I think this is a great start, and I think it will be very helpful for people to come up to speed on the CTS.  I certainly would have benefited from documentation like this when I first started looking at the legacy test suite!  I have a few comments below.

docs/writing_tests.md
psalz: I completely agree, it was not my intent to go to great lengths explaining Catch2 here. Instead I just wanted to list the most important macros and how they tend to be used in the CTS (for example, using `TEMPLATE_TEST_CASE_SIG` for testing APIs with 1/2/3 dimensions).

> Also, I think it is more important to describe macros that were implemented only in our repo. For example DISABLE_FOR_TEST_CASE and DISABLE_FOR_TEMPLATE_TEST_CASE_SIG

Yes, this is what the "Special Macros & Custom Matchers" was intended for!

docs/writing_tests.md
psalz: This already is about the extent to which I wanted to explain legacy test cases. The only reason I'm including this at all is so that people aren't confused when they browse existing tests!

docs/writing_tests.md
psalz: Agreed, although I don't have a clear answer here yet. I'm leaning towards fewer files with multiple test cases each, though. The only reasons for having multiple files at all (that I can think of right now) would be to group them conceptually, and to keep incremental build times low.

keryell:
Good to have!

psalz:
I haven't had much time lately to work on this unfortunately. I just added some more sections and I think we're in a decent enough state for an initial release!

keryell (approved):
Thanks!

PR #305, opened 2022-03-15
Fix opencl constructors tests

kopylovanat:
Add setup for tests.

Remove opencl tests from dpcpp filter

Remove clRelease* functions for instances from get_native since according to [C.6. Interoperability with the OpenCL API](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:interfacing-with-opencl) 

> the destructor for the [SYCL runtime](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sycl-runtime) classes which encapsulate a reference counted OpenCL opaque type must release that instance to decrease the reference count of the OpenCL resource

Instances of SYCL classes runtime classes that encapsulate OpenCL opaque types are destructed during tests and get_native retrieve instances of that opaque types, not creating new ones. Instances of OpenCL types are released in test_base_opencl::cleanup()

### DIFF
--- a/ci/dpcpp.filter
+++ b/ci/dpcpp.filter
@@ -1,5 +1,4 @@
 buffer
 exceptions
 multi_ptr
-opencl_interop
 reduction
--- a/tests/opencl_interop/opencl_interop_constructors.cpp
+++ b/tests/opencl_interop/opencl_interop_constructors.cpp
@@ -85,9 +85,6 @@ class TEST_NAME :
         if (interopDeviceID != m_cl_device_id) {
           FAIL(log, "device was not constructed correctly");
         }
-        if (!CHECK_CL_SUCCESS(log, clReleaseDevice(interopDeviceID))) {
-          FAIL(log, "failed to release OpenCL device");
-        }
       }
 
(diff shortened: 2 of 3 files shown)


ProGTX (approved):
From my understanding this section is more relevant:
```
The lifetime of a [SYCL runtime](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sycl-runtime) class that encapsulates an OpenCL opaque type and the instance of that opaque type retrieved via the get_native() free function are not tied in either direction given correct usage of OpenCL reference counting. For example if a user were to retrieve a cl_command_queue instance from a SYCL queue instance and then immediately destroy the SYCL queue instance, the cl_command_queue instance is still valid. Or if a user were to construct a SYCL queue instance from a cl_command_queue instance and then immediately release the cl_command_queue instance, the SYCL queue instance is still valid.
```

That's because in SYCL 1.2.1 calling `get` would retain the object as well, but that's not the case anymore in SYCL 2020 with `get_native`.

That doesn't affect the actual PR, though.

PR #306, opened 2022-03-16
Improve generic code in `type_coverage.h`

andreyromanof:
This PR provides a new function `for_all_combinations` that can combinate types, values, etc.
Refactor old functions like `for_all_type` for removing unused variable and dummy values in fold expression.
Also made constructor of `named_type_pack` private. Instead of a constructor, there is the factory function `generate_named`. This change will allow us to freely change attributes and the way to construct the actual type pack inside the class without changing existing tests.

The function `for_all_combinations` will be used in new accessor tests.

Depends on:
- #308

### DIFF
--- a/tests/common/type_coverage.h
+++ b/tests/common/type_coverage.h
@@ -14,7 +14,7 @@
 
 #include <sycl/sycl.hpp>
 
-#include "type_traits.h"
+#include "../../util/type_traits.h"
 
 /**
  * @brief Retrieve type name; by default just forward the given one
--- a/tests/common/type_coverage.h
+++ b/tests/common/type_coverage.h
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-// Provide common functions for type coverage
+//  Provide common functions for type coverage
 //
 *******************************************************************************/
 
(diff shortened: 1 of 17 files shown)


tests/common/type_coverage.h
```diff
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-// Provide common functions for type coverage
+//  Provide common functions for type coverage
```
bader, line 5: Minor: Is it a change made by clang-format? If so, I think it tries to align with the line 3, so I would remove one space there before SYCL instead.

tests/common/type_coverage.h
```diff
@@ -14,14 +14,14 @@
...
+
 /**
  * @brief Retrieve type name; by default just forward the given one
  */
 template <typename T>
 struct type_name_string {
-    static std::string get(std::string dataType) {
-        return dataType;
-    }
+  static std::string get(std::string dataType) { return dataType; }
```
bader, line 24: @andreyromanof, please, refrain from adding code formatting changes for the parts, which logic you don't modify. It makes code review difficult. Please, move such changes to a separate PR. It would much easier to review this patch if there is a clear separation between code formatting (non-functional) changes and changes "improving" `type_coverage.h`.

andreyromanof: I moved format changes to separate PR (#308), to simplify the review process for this PR.

bader: Do you mean we should hold reviews until #308 is merged?

andreyromanof: I think we should because of the large changes in format. 
I converted PR to draft until #308 is merged.

andreyromanof:
I can't edit the reviewers so I duplicate:
>I think we should hold reviews because of the large changes in format.
>I converted PR to draft until https://github.com/KhronosGroup/SYCL-CTS/pull/308 is merged.

ProGTX (approved):

PR #308, opened 2022-03-16
Apply format to `type_coverage` and `type_traits`

andreyromanof:
This PR doesn't contain functional or logical changes, only format applying to `type_coverage` and `type_traits`. 
Mostly to simplify the review process for #306, as after merging this PR, #306 will contain only functional changes.

### DIFF
--- a/tests/common/type_coverage.h
+++ b/tests/common/type_coverage.h
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-// Provide common functions for type coverage
+//  Provide common functions for type coverage
 //
 *******************************************************************************/
 
@@ -19,9 +19,7 @@
  */
 template <typename T>
 struct type_name_string {
-    static std::string get(std::string dataType) {
-        return dataType;
-    }
+  static std::string get(std::string dataType) { return dataType; }
 };
 
 /**
@@ -31,9 +29,9 @@ struct type_name_string {
  */
 template <typename T, size_t nElements>
 struct type_name_string<sycl::vec<T, nElements>> {
-    static std::string get(const std::string& dataType) {
-      return "sycl::vec<" + dataType + "," + std::to_string(nElements) + ">";
-    }
+  static std::string get(const std::string& dataType) {
+    return "sycl::vec<" + dataType + "," + std::to_string(nElements) + ">";
+  }
 };
 
 /**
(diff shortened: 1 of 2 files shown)


tests/common/type_coverage.h
```diff
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-// Provide common functions for type coverage
+//  Provide common functions for type coverage
```
bader, line 5:
```suggestion
// Provide common functions for type coverage
```

tests/common/type_coverage.h
```diff
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
```
bader, line 3:
```suggestion
// SYCL 2020 Conformance Test Suite
```

util/type_traits.h
```diff
@@ -61,6 +56,6 @@ using is_cl_float_type =
                        std::is_same<sycl::cl_double, T>::value ||
                        std::is_same<sycl::cl_half, T>::value>;
 
-} // namespace
+}  // namespace
```
bader, line 59: Probably we can discuss why do we have this configuration: https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/.clang-format#L41 and whether we want to change it. I'm fine with current existing trailing comment style (i.e. 1 space).

bader (approved):

tests/common/type_coverage.h
andreyromanof: I think we should leave two spaces for consistency with the rest codebase.
Few examples of sources where two spaces used in the file header:
- [First](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/image/image_api.cpp)
- [Second](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/common/macros.h)
- [Third](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/group/group_api.cpp)

util/type_traits.h
andreyromanof: I'm ok with changing format, but even right now not all files corresponding to the current format. In case of changing, we probably should run `clangformat.py` to format all files at once.

ProGTX (approved):

PR #309, opened 2022-03-17
Fix macro for enabling extension properties tests

kopylovanat:


kopylovanat:
@bader 

>What problem does this change suppose to fix?

It fixes the problem that macro for enabling tests is erroneously named the same as macro provided by the extension.

> Shouldn't the option start with SYCL_CTS_ENABLE_ prefix and end with _TESTS suffix?

Renamed to SYCL_CTS_ENABLE_EXT_ONEAPI_PROPERTIES_TESTS. Should I rename macros for other extensions to have SYCL_CTS prefix?

bader:
> Should I rename macros for other extensions to have SYCL_CTS prefix?

I would really appreciate this.

kopylovanat:
> > Should I rename macros for other extensions to have SYCL_CTS prefix?
> 
> I would really appreciate this.

Done.

bader:
Probably, we should also update `SYCL_CTS_TEST_DEPRECATED_FEATURES` to `SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS`. What do you think?

keryell:
This is probably something we should add in the spec.
If `SYCL_` macros are used instead of `__SYCL_` macros we should somehow reserve the name space and why not state that `SYCL_CTS_` is reserved for the SYCL CTS.

kopylovanat:
> Probably, we should also update `SYCL_CTS_TEST_DEPRECATED_FEATURES` to `SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS`. What do you think?

Yes, good idea, updated.

bader:
I'm not sure if we need to reserve a name space for SYCL-CTS macro in the spec. The issue @kopylovanat mentioned this [this comment](https://github.com/KhronosGroup/SYCL-CTS/pull/309#issuecomment-1072124120) looks like a bug in the cmake scripts. Anyone can make a similar mistake with standard C++ macro by misusing it in some other project.

bader (approved):
Thanks!

gmlueck:
> This is probably something we should add in the spec.
> If `SYCL_` macros are used instead of `__SYCL_` macros we should somehow reserve the name space and why not state that `SYCL_CTS_` is reserved for the SYCL CTS.

The SYCL 2020 spec already reserves all macros that start with `SYCL_`.  Macros that start with `SYCL_EXT` and `SYCL_IMPLEMENTATION` are reserved for vendor extensions.  These rules already prevent vendors from defining macros that start with `SYCL_CTS`.

keryell:
> The SYCL 2020 spec already reserves all macros that start with `SYCL_`. Macros that start with `SYCL_EXT` and `SYCL_IMPLEMENTATION` are reserved for vendor extensions. These rules already prevent vendors from defining macros that start with `SYCL_CTS`.

I understood the spec as the `SYCL_` macro namespace as nobody *outside* of the implementation can define a macro there but an implementation can (so potentially conflicting with some `SYCL_CTS_`), for example some SYCL implementation use internally some RS-232 protocol with Clear To Send to launch the kernel on some accelerator board. :-).

Also in your interpretation I cannot see why some external `SYCL_CTS` would be allowed, except by a side-project agreement with the CTS. So instead it should be mentioned in the spec.

Anyway, I have not been able to find in the current spec the restriction about `SYCL_` macro namespace. Do you have the place?

gmlueck:
> Anyway, I have not been able to find in the current spec the restriction about SYCL_ macro namespace. Do you have the place?

Actually, the wording I was thinking about was in the chapter on extensions.  This says that macros starting with `SYCL_EXT` and `SYCL_IMPLEMENTATION` are reserved for vendor extensions, but it does not forbid an implementation from defining other macros that start with `SYCL`.

I also do not see any language forbidding a vendor from defining other symbols in the `::sycl` namespace.  We say that `::sycl::detail` is reserved for a vendor's implementation details, but we do not forbid a vendor from defining symbols elsewhere.

I'm not sure it makes sense for the spec to forbid an implementation from defining other macros that start with `SYCL` or from defining other symbols in `::sycl`.  I think this is a bad idea, but it seems like overreach to have the spec forbid this.

The way I see it, the CTS is created by the Khronos SYCL group, so we are free to use macros that start with `SYCL_CTS`.  We know that we will never define any macros in future versions of the spec that have this prefix.  If some other entity outside of Khronos wants to define their own test suite, I don't see why they would use macros that start with `SYCL` at all.  Wouldn't they just choose some other prefix for their macros?

run_conformance_tests.py
```diff
@@ -121,7 +121,7 @@ def generate_cmake_call(cmake_exe, build_system_name, full_conformance,
         '..',
         '-G' + build_system_name,
         '-DSYCL_CTS_ENABLE_FULL_CONFORMANCE=' + full_conformance,
-        '-DSYCL_CTS_TEST_DEPRECATED_FEATURES=' + test_deprecated_features,
+        '-DSSYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS=' + test_deprecated_features,
```
ProGTX, line 124:
```suggestion
        '-DSYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS=' + test_deprecated_features,
```

ProGTX (approved):

PR #311, opened 2022-03-30
Add tests for `sycl::exception` constructors

andreyromanof:
This PR provides tests for sycl::exception constructors according to [paragraph 2.4](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/exceptions.asciidoc#24-syclexception) of the test plan

### DIFF
--- a/tests/exceptions/exceptions_constructors.cpp
+++ b/tests/exceptions/exceptions_constructors.cpp
@@ -14,12 +14,12 @@ namespace exception_constructors {
 
 /**
  * @brief The function helps to verify that exception was constructed correctly.
- * This overload will be expect that .has_context() return false and will try to
- * catch exception after calling get_context().
+ * This overload expects that \c .has_context() returns false and the right
+ * exception is thrown after calling \c get_context().
  *
  * @param e The exception instance that needs to be verified
- * @param errcode Expected error code from .code() member function
- * @param errcat Expected error category from .category() member function
+ * @param errcode Expected error code from \c .code() member function
+ * @param errcat Expected error category from \c .category() member function
  */
 inline void check_exception(sycl::exception e, const std::error_code errcode,
                             const std::error_category& errcat) {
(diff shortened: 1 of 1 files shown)


ProGTX (approved):

tests/exceptions/exceptions_constructors.cpp
```diff
@@ -0,0 +1,302 @@
...
+    }
+  }
+}
+
+#ifdef SYCL_BACKEND_OPENCL
+TEST_CASE("Constructors for sycl::exception with OpenCL error code",
+          "[exception]") {
+  const std::string what_arg_str = "test";
+  const sycl::context ctx;
+  if (ctx.get_backend() == sycl::backend::opencl) {
```
ProGTX, line 194: I think it would be better if this was an early return for the opposite case:

```suggestion
  if (ctx.get_backend() != sycl::backend::opencl) {
    WARN("OpenCL backend is not supported on this device");
    return;
  }
  // Rest of code without extra indentation
```

tests/exceptions/exceptions_constructors.cpp
```diff
@@ -0,0 +1,302 @@
...
+      check_exception(copy, errcode, sycl::sycl_category());
+    }
+  }
+}
+
+#ifdef SYCL_BACKEND_OPENCL
+TEST_CASE("Constructors for sycl::exception with OpenCL error code",
+          "[exception]") {
+  const std::string what_arg_str = "test";
+  const sycl::context ctx;
```
ProGTX, line 193: I guess it's up to the implementation to ensure this constructs a context that uses the OpenCL backend?

ProGTX (approved):

keryell (changes requested):
That looks good.

tests/exceptions/exceptions_constructors.cpp
andreyromanof: I added lambda that should prefer a device with OpenCL backend. I didn't add try/catch, because catch block will be too far down which will be quite unhandy in my opinion

tests/exceptions/exceptions_constructors.cpp
andreyromanof: Applied

keryell (approved):
Thanks!

PR #312, opened 2022-03-30
Add tests for sycl::exception_list

kopylovanat:
Tests for [2.5. sycl::exception_list](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/exceptions.asciidoc#25-syclexception_list)

Fix test plan because for empty exception_list begin() and end() should be equal.

### DIFF
--- a/tests/exceptions/exception_list.cpp
+++ b/tests/exceptions/exception_list.cpp
@@ -1,0 +1,40 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for sycl::exception_list
+//
+*******************************************************************************/
+
+#include <catch2/catch_template_test_macros.hpp>
+
+#include "../common/common.h"
+
+using namespace sycl_cts;
+
+TEST_CASE("Check sycl::exception_list member types", "[exception]") {
+  using sycl::exception_list;
+  CHECK(std::is_same_v<exception_list::value_type, std::exception_ptr>);
+  CHECK(std::is_same_v<exception_list::reference, exception_list::value_type&>);
+  CHECK(std::is_same_v<exception_list::const_reference,
+                       const exception_list::value_type&>);
+  CHECK(std::is_same_v<exception_list::size_type, std::size_t>);
+}
+
+TEST_CASE(
+    "Check that sycl::exception_list iterators satisfy LegacyForwardIterator "
+    "requirements",
+    "[exception]") {
+  using It = sycl::exception_list::iterator;
+  CHECK(std::is_base_of_v<std::forward_iterator_tag,
+                          std::iterator_traits<It>::iterator_category>);
+  using ConstIt = sycl::exception_list::iterator;
+  CHECK(std::is_base_of_v<std::forward_iterator_tag,
+                          std::iterator_traits<ConstIt>::iterator_category>);
+}
+
+TEST_CASE("Check default-constructed sycl::exception_list", "[exception]") {
+  sycl::exception_list list;
+  CHECK(list.size() == 0);
+  CHECK(list.begin() == list.end());
+}
--- a/test_plans/exceptions.asciidoc
+++ b/test_plans/exceptions.asciidoc
@@ -219,7 +219,7 @@ Verify that:
 For default-constructed `exception_list` check that:
 
 * `exception_list::size()` is zero
-* `++exception_list::begin()` is equal to `exception_list::end()`
+* `exception_list::begin()` is equal to `exception_list::end()`
 
 === sycl::async_handler
 


bader (approved):

tests/exceptions/exception_list.cpp
```diff
@@ -0,0 +1,40 @@
...
+    "Check that sycl::exception_list iterators satisfy LegacyForwardIterator "
+    "requirements",
+    "[exception]") {
+  using It = sycl::exception_list::iterator;
+  CHECK(std::is_base_of_v<std::forward_iterator_tag,
+                          std::iterator_traits<It>::iterator_category>);
+  using ConstIt = sycl::exception_list::iterator;
+  CHECK(std::is_base_of_v<std::forward_iterator_tag,
+                          std::iterator_traits<ConstIt>::iterator_category>);
+}
```
ProGTX, line 34: This test doesn't seem to cover all the requirements listed in https://en.cppreference.com/w/cpp/named_req/ForwardIterator . I'm happy if that is done as a separate PR later - maybe we should have some generic checker for some C++ named requirements?

ProGTX (approved):

tests/exceptions/exception_list.cpp
keryell: We need to switch to C++20 to have this check for free. :-)

keryell: > This test doesn't seem to cover all the requirements listed in https://en.cppreference.com/w/cpp/named_req/ForwardIterator . I'm happy if that is done as a separate PR later - maybe we should have some generic checker for some C++ named requirements?

I would prefer to update this PR accordingly.

keryell: > C++20 aside there are also checks for C++ named requirements in boost (https://www.boost.org/doc/libs/1_78_0/libs/concept_check/using_concept_check.htm). Maybe we could consider to use boost in SYCL CTS as already existing and tested solution for such checks?

I do not know.
For such a simple test it might be simpler to just SFINAE on a small function which exercises the iterator features.

psalz: I would advise against introducing Boost into the CTS. I've had numerous hard to debug issues in the past when trying to compile various Boost libraries through SYCL compilers (mostly because Boost has a ton of hacks/workarounds for various compilers but is not yet (AFAIK) SYCL aware).

keryell: OK, so it looks like we are more in the mood of having a SFINAE function/class which exercises all the iterator operations.

vasilytric: In the [test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/exceptions.asciidoc#25-syclexception_list) for `sycl::exception_list` says that `sycl::exception_lis::iterator` should satisfies LegacyForwardIterator. But in the [SYCL-2020 rev.5 #4.13.2. Exception class interface](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:exception.class) there aren't any requirements for `sycl::exception_list::iterator`. The SYCL-2020 state the following statement to `sycl::exception_list::iterator`: 
- `using iterator = /*unspecified*/` 

According to statements above I have two questions @keryell @ProGTX  :
- Should we provide any verifications for `sycl::exception_list::iterator`?
- Should we create an issue in [SYCL-Docs](https://github.com/KhronosGroup/SYCL-Docs) that `sycl::exception_list::iterator` has no requirements?

gmlueck: There is already any open issue against the spec saying that `sycl::exception_list` needs to be clarified: https://github.com/KhronosGroup/SYCL-Docs/issues/167

keryell: At the end, even if the iterator are unspecified, they remain iterators, so it makes sense to test for their basic features.

vasilytric: > There is already any open issue against the spec saying that `sycl::exception_list` needs to be clarified: [KhronosGroup/SYCL-Docs#167](https://github.com/KhronosGroup/SYCL-Docs/issues/167)

@gmlueck Thanks for the information

vasilytric: > At the end, even if the iterator are unspecified, they remain iterators, so it makes sense to test for their basic features.

@keryell Thank you, I'll take note :)

vasilytric: @bader we will have a lot of changes, so I think that we are going to create a new PR

bader: Okay. I suggest we merge this pull request and address the comments from this conversation separately. @ProGTX, @keryell, does it sound okay to you?

keryell: This is good for me too.

keryell (approved):
Thanks.

PR #315, opened 2022-04-01
Fix input value type for float and half math tests

kopylovanat:
`float` and `half` tests shouldn't use double type values in the device code.


bader (approved):

keryell (approved):
Thanks.

PR #316, opened 2022-04-01
Move checks for handler::copy for double type to separate test

kopylovanat:
Separate checks for double type to run only if a device doesn't support fp64.

### DIFF
--- a/tests/handler/handler_copy_fp64.cpp
+++ b/tests/handler/handler_copy_fp64.cpp
@@ -1,0 +1,28 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests the API for sycl::handler::copy for double
+//
+*******************************************************************************/
+
+#include "handler_copy_common.h"
+
+#include "catch2/catch_test_macros.hpp"
+
+namespace handler_copy_fp64 {
+using namespace handler_copy_common;
+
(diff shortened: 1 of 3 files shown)


tests/handler/handler_copy_fp64.cpp
```diff
@@ -0,0 +1,28 @@
...
+TEST_CASE("Tests the API for sycl::handler::copy for double", "[handler]") {
+  auto queue = util::get_cts_object::queue();
+  if (queue.get_device().has(sycl::aspect::fp64)) {
+    log_helper lh;
+
+    test_all_variants<double>(lh, queue);
+    test_all_variants<sycl::double16>(lh, queue);
+  } else {
+    WARN("Device does not support double precision floating point operations");
+  }
```
bader, line 25:
```suggestion
  if (!queue.get_device().has(sycl::aspect::fp64)) {
    WARN("Device does not support double precision floating point operations. Skipping the test case.");
    return;
  }
  
  log_helper lh;
  test_all_variants<double>(lh, queue);
  test_all_variants<sycl::double16>(lh, queue);
```

kopylovanat: Done.

bader (approved):

keryell (approved):
Thank you for the cleaning.

PR #317, opened 2022-04-04
Move test for invoke template kernel for `double` type to separate test

kopylovanat:
Separate checks for double type to run only if a device doesn't support fp64.


bader (approved):

keryell (approved):
Thanks for the cleaning.

PR #318, opened 2022-04-05
Tests implementation for SYCL2020 `accessor` constructors

andreyromanof:
This PR provides tests for `sycl::accessor` constructors. Tests cover only non-deprecated access modes and targets.

Depends on:
- #321 
- #322 
- #324

### DIFF
--- a/ci/computecpp.filter
+++ b/ci/computecpp.filter
@@ -1,4 +1,3 @@
-accessor
 accessor_legacy
 atomic
 buffer
--- a/ci/computecpp.filter
+++ b/ci/computecpp.filter
@@ -1,3 +1,4 @@
+accessor
 accessor_legacy
 atomic
 buffer
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -12,7 +12,7 @@
 #include "../../util/sycl_exceptions.h"
 #include "../common/common.h"
 #include "../common/type_coverage.h"
-#include "../common/value_helper.h"
+#include "../common/value_operations.h"
 
 #include <catch2/matchers/catch_matchers.hpp>
 
@@ -264,11 +264,11 @@ void read_write_zero_dim_acc(AccT testing_acc, ResultAccT res_acc) {
 
   if constexpr (AccessMode != sycl::access_mode::write) {
     DataT acc_ref(testing_acc);
-    res_acc[0] = value_helper::are_equal(acc_ref, other_data);
+    res_acc[0] = value_operations::are_equal(acc_ref, other_data);
   }
   if constexpr (AccessMode != sycl::access_mode::read) {
     DataT acc_ref(testing_acc);
-    value_helper::change_val(acc_ref, changed_val);
+    value_operations::assign(acc_ref, changed_val);
   }
 }
 
@@ -316,7 +316,7 @@ void check_zero_dim_constructor(GetAccFunctorT get_accessor_functor) {
     CHECK(compare_res);
   }
   if constexpr (AccessMode != sycl::access_mode::read) {
-    CHECK(value_helper::are_equal(some_data, changed_val));
+    CHECK(value_operations::are_equal(some_data, changed_val));
   }
 }
 
@@ -339,10 +339,10 @@ void read_write_acc(AccT testing_acc, ResultAccT res_acc) {
   auto id = util::get_cts_object::id<Dimension>::get(0, 0, 0);
 
   if constexpr (AccessMode != sycl::access_mode::write) {
-    res_acc[0] = value_helper::are_equal(testing_acc[id], other_data);
+    res_acc[0] = value_operations::are_equal(testing_acc[id], other_data);
   }
   if constexpr (AccessMode != sycl::access_mode::read) {
-    value_helper::change_val(testing_acc[id], changed_val);
+    value_operations::assign(testing_acc[id], changed_val);
   }
 }
 
@@ -395,7 +395,7 @@ void check_common_constructor(GetAccFunctorT get_accessor_functor,
     CHECK(compare_res);
   }
   if constexpr (AccessMode != sycl::access_mode::read) {
-    CHECK(value_helper::are_equal(some_data, changed_val));
+    CHECK(value_operations::are_equal(some_data, changed_val));
   }
 }
 
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -1,0 +1,554 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Common functions for the accessor tests.
+//
+*******************************************************************************/
+
+#ifndef SYCL_CTS_ACCESSOR_COMMON_H
+#define SYCL_CTS_ACCESSOR_COMMON_H
+#include "../../util/sycl_exceptions.h"
+#include "../common/common.h"
+#include "../common/get_cts_string.h"
+#include "../common/type_coverage.h"
+
+#include "catch2/catch_test_macros.hpp"
+#include "catch2/matchers/catch_matchers.hpp"
+
+#include "accessor_stub.hpp"
+
+namespace accessor_tests_common {
+
+using namespace sycl_cts;
+
+constexpr int expected_val = 42;
+constexpr int changed_val = 1;
+
+/**
+ * @brief Function helps to get string section name that will contain template
+ * parameters and function arguments
+ *
+ * @tparam DimensionT Integer representing dimension
+ * @param type_name String with name of the testing type
+ * @param access_mode_name String with name of the testing access mode
+ * @param target_name String with name of the testing target
+ * @param section_description String with human-readable description of the test
+ * @return std::string String with name for section
+ */
+template <int DimensionT>
+inline std::string get_section_name(const std::string& type_name,
+                                    const std::string& access_mode_name,
+                                    const std::string& target_name,
+                                    const std::string& section_description) {
+  using namespace sycl_cts::get_cts_string;
+
+  std::string name = "Test ";
+  name += section_description;
+  name += " with parameters: <";
+  name += type_name;
+  name += "><";
+  name += access_mode_name;
+  name += "><";
+  name += target_name;
+  name += "><";
+  name += std::to_string(DimensionT) + ">";
+  return name;
+}
+
+/**
+ * @brief Factory function for getting type_pack with fp16 type
+ */
+inline auto get_fp16_type() {
+  static const auto types = named_type_pack<sycl::half>::generate("sycl::half");
+  return types;
+}
+
+/**
+ * @brief Factory function for getting type_pack with fp64 type
+ */
+inline auto get_fp64_type() {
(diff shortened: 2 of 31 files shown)


andreyromanof:
Have to add to filter hipsycl and computecpp as they doesn't support marray([hipsycl](https://github.com/KhronosGroup/SYCL-CTS/runs/5830730001?check_suite_focus=true#step:8:108) error message; [computecpp](https://github.com/KhronosGroup/SYCL-CTS/runs/5830729872?check_suite_focus=true#step:8:15) error message).

ci/computecpp.filter
```diff
@@ -1,3 +1,4 @@
+accessor
```
bader, line 1: Can we disable the test cases instead of the whole test?

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,578 @@
...
+ * @brief Returns result of comparing of left and right
+ * @return Returns true if left == right, false otherwise
+ */
+template <typename T, typename U = T, int N>
+inline bool compare_vals(const sycl::vec<T, N>& left, const U& right) {
+  bool are_equal = true;
+  for (size_t i = 0; i < N; ++i) {
+    are_equal &= left[i] == right;
+  }
+  return are_equal;
```
keryell, line 185: Use early exit
```suggestion
  for (size_t i = 0; i < N; ++i)
    if (left[i] != right)
      return false;
  return true;
```

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,578 @@
...
+ * @brief Returns result of comparing of left and right
+ * @return Returns true if left == right, false otherwise
+ */
+template <typename T, typename U = T, int N>
+inline bool compare_vals(const sycl::marray<T, N>& left, const U& right) {
+  bool are_equal = true;
+  for (size_t i = 0; i < N; ++i) {
+    are_equal &= left[i] == right;
+  }
+  return are_equal;
```
keryell, line 198: Idem

tests/accessor/generic_accessor_constructors.hpp
```diff
@@ -0,0 +1,299 @@
...
+template <typename DataT, int DimensionT, sycl::access_mode AccessModeT,
+          sycl::target TargetT>
+void test_default_constructor(const std::string& type_name,
+                              const std::string& access_mode_name,
+                              const std::string& target_name) {
+  const auto section_name = get_section_name<DimensionT>(
+      type_name, access_mode_name, target_name, "Default constructor");
+
+  SECTION(section_name) {
+    auto get_acc_functor = []() {
```
keryell, line 29:
```suggestion
    auto get_acc_functor = [] {
```

tests/accessor/accessor_common.h
keryell: More generally I have the feeling that all these double definitions for `sycl::vec` and `sycl::marray` could be factorized.
Perhaps by adding the equivalent of a C++20 concept true for `sycl::vec` or `sycl::marray`, but implemented by `std::enable_if_t`.

keryell (changes requested):
Great contribution.

ci/computecpp.filter
andreyromanof: I added `#ifdef` command to disable part of code and `DISABLE_FOR_TEST_CASE` to disable some tests . This should allow us to remove whole test from the filter.

tests/accessor/accessor_common.h
andreyromanof: Fixed in #324

tests/accessor/accessor_common.h
andreyromanof: As it's yet another one implementation of `value_helper` functions ([first impl](https://github.com/KhronosGroup/SYCL-CTS/blob/b00bde0e3580d0c8a3cd0c7f7c46f60c5cb4d596/tests/extension/oneapi_device_global/device_global_common.h#L84) in `device_global` tests), I moved these functions in separate file in PR #324

tests/accessor/generic_accessor_constructors.hpp
andreyromanof: Fixed

keryell (approved):
Thanks.

keryell:
@illuhad @ProGTX any second review on this big PR with a lot of other ones depending on it?

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,450 @@
...
+
+#ifndef SYCL_CTS_ACCESSOR_COMMON_H
+#define SYCL_CTS_ACCESSOR_COMMON_H
+
+#include "../../util/sycl_exceptions.h"
+#include "../common/common.h"
+#include "../common/type_coverage.h"
+#include "../common/value_helper.h"
+
+#include "catch2/catch_test_macros.hpp"
```
psalz, line 17: I think we usually use `<angle brackets>` for third-party includes!

psalz (changes requested):
Thanks! While the fact that none of the implementations can currently compile these tests is obviously not great, I think we agreed that such PRs can still be merged to support test-driven development. Do you think this situation could be at least somewhat remedied if we had a `DISABLED_FOR_SECTION`?

tests/accessor/accessor_common.h
yuriykoch: Actually, looks like it would be better to use the following:
```
#include "../common/common.h"
// instead of <catch2/catch_test_macros.hpp>
```
and
```
#include <catch2/catch_template_test_macros>
#include "../common/macros.h"
// instead of <catch2/catch_template_test_macros.hpp>
```
because of `#undef FAIL` within [../common/macros.h](https://github.com/KhronosGroup/SYCL-CTS/blob/0a5d5d689124edcc2280a0d7d36d2e16513904f5/tests/common/macros.h#L16)
@psalz, @andreyromanof   what do you think?

psalz: You're right, including `<catch2/catch_test_macros.hpp>` is superfluous in this case, just including "`../common/common.h"` should be sufficient (as that transitively includes `macros.h`). Also the custom `FAIL` macro is actually not really needed in this case, it only exists for backwards compatibility with the old `FAIL` macro that received a `logger`.

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,450 @@
...
+
+    queue
+        .submit([&](sycl::handler& cgh) {
+          sycl::accessor res_acc(res_buf);
+          auto acc = get_accessor_functor();
+          if constexpr (Target == sycl::target::host_task) {
+            cgh.host_task(
+                [=] { check_def_constructor_post_conditions(acc, res_acc); });
+          } else if constexpr (Target == sycl::target::device) {
+            cgh.parallel_for_work_group(r, [acc, res_acc](sycl::group<1>) {
```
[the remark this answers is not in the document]
andreyromanof: I'm reusing this common function for `local_accessor` tests. As I understand, we can't use `local_accessor` in `single_task` or `parallel_for` with `sycl::range` as argument, so I use `parallel_for_work_group` instead for compability.

>A local accessor must not be used in a [SYCL kernel function](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sycl-kernel-function) that is invoked via single_task or via the simple form of parallel_for that takes a range parameter. [4.7.6.11. Local accessor](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:accessor.local)

tests/accessor/accessor_common.h
andreyromanof: Removed extra includes and added angle brackets to catch header `catch2/matchers/catch_matchers.hpp` include

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,451 @@
...
+    sycl::buffer<DataT, Dimension> data_buf(&some_data, r);
+
+    queue
+        .submit([&](sycl::handler& cgh) {
+          sycl::accessor res_acc(res_buf);
+          auto acc = get_accessor_functor(data_buf, cgh);
+
+          if (acc.is_placeholder()) {
+            cgh.require(acc);
+          }
```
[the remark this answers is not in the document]
yuriykoch: Upd: found `AccType` dependency in #325 : https://github.com/KhronosGroup/SYCL-CTS/pull/325/files#diff-8067235ffc00be4a68e8da8448acbf789b606099c5dbc1e969df066646bd2f64R410.

tests/accessor/accessor_common.h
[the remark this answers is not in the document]
psalz: Ah yes, makes sense!

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,451 @@
...
+
+          if (Target == sycl::target::host_task) {
```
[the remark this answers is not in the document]
andreyromanof: Fixed

tests/accessor/accessor_common.h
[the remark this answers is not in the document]
yuriykoch: > For me, the link goes just to the Files changed tab not to a concrete place.

Yep, looks like github restriction. It doesn't show exact line in `accessor_common.h` until the diff is loaded by clicking on "load diff".

> I stood for waiting until all PRs are merged, not only this one.

Makes sense. Unfortunatelly looks like I'll be a bit blocked until then, but I'll try to find a workaround.

> There is a difference in API. Zero dimension accessors behave like a scalar variable without a subscript operator, unlike 1-2-3 dimension accessors([spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_command_accessors) lines 155 and 158). That's why read_write_zero_dim_acc functions exist, as they change accessor without invoking the subscript operator. No doubt we can constantly check if we working with a zero dimension accessor, to separate usage of 1-2-3 dimension accessor from 0 dimension accessor, but IMO it will be too many constexpr checks, that can be avoided by a different function.

Look, I totally agree that havin a separate function within `accessor_common.h` might be really nice.
But as I see it, there is no reason to present a separate API to the user of `accessor_common.h`, as it would require to use `if constexpr (Dimension == 0)` even if there is no difference at all from the point of the caller. Looks like even `sycl::range` usage might be resolved in a generic way.
Still, moving refactoring out of this PR means this question is out of scope of this PR as well.

keryell:
So, what do we do with this?
Do we merge this as is and go on improving with other PR?
@psalz you asked for some changes. Are you happy now?

andreyromanof:
> Do we merge this as is and go on improving with other PR?

I think it's a right decision to make a improving PR after merging current PRs with accessor tests. It will allow us to make changes in one common branch.

andreyromanof:
> Do you think this situation could be at least somewhat remedied if we had a DISABLED_FOR_SECTION?

Sorry for delay answer. In other cases it can help, but in accessors tests I think it won't make a big deal. For example DPCPP doesn't implement `sycl::target` that used in all sections, so if we theoretically could disable sections, we would disable all of them for DPCPP.

psalz (approved):

PR #319, opened 2022-04-05
Remove stub from `device_global` test

andreyromanof:
This provides changing of non-existing `sycl_stub` namespace to the `sycl` namespace that corresponds to the spec.


bader (approved):

keryell (approved):
Thanks.

keryell (approved):
Thanks.

PR #321, opened 2022-04-05
Fix marray namespace and typos in the comments

bader:


bader:
@andreyromanof, FYI.
I think some tests from https://github.com/KhronosGroup/SYCL-CTS/pull/318 fail compilation due to wrong usage of the namespace.

keryell (approved):
Thanks!

PR #322, opened 2022-04-05
Change cl types to generic c++ types

andreyromanof:


util/type_traits.h
```diff
@@ -53,9 +53,8 @@ template <typename T>
 using is_cl_float_type =
     std::bool_constant<std::is_floating_point<T>::value ||
                        std::is_same<sycl::half, T>::value ||
-                       std::is_same<sycl::cl_float, T>::value ||
-                       std::is_same<sycl::cl_double, T>::value ||
-                       std::is_same<sycl::cl_half, T>::value>;
+                       std::is_same<float, T>::value ||
+                       std::is_same<double, T>::value>::value>;
```
bader, line 57: Aren't these types already covered by `std::is_floating_point<T>::value` check?

util/type_traits.h
```diff
@@ -53,9 +53,8 @@ template <typename T>
 using is_cl_float_type =
```
bader, line 53:
```suggestion
using is_sycl_float_type =
```

andreyromanof: Changed.

util/type_traits.h
andreyromanof: They are. Removed double and float check.

util/type_traits.h
```diff
@@ -50,12 +50,9 @@ using has_atomic_support = contains<T, int, unsigned int, long, unsigned long,
  * @brief Checks whether T is a floating-point sycl type
  */
 template <typename T>
-using is_cl_float_type =
-    std::bool_constant<std::is_floating_point<T>::value ||
-                       std::is_same<sycl::half, T>::value ||
-                       std::is_same<sycl::cl_float, T>::value ||
-                       std::is_same<sycl::cl_double, T>::value ||
-                       std::is_same<sycl::cl_half, T>::value>;
+using is_sycl_floating_point_type =
```
bader, line 53: I just noticed that there is `is_sycl_floating_point`, I think this alias can removed completely.
Please, update the code using `is_cl_float_type`.

bader (changes requested):

util/type_traits.h
bader: BTW, what is the reason of this change? Can we use `is_sycl_floating_point` instead of `is_cl_float_type`? I see that `is_cl_float_type` is still used by legacy tests and I suppose might be required to validate the compatibility with SYCL 1.2.1 version.

andreyromanof: > BTW, what is the reason of this change?

I faced an issue with hipsycl compilation in #318. The error was saying that there are no cl_* types,  and I didn't found these types in [paragrarph 4.14.1](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_scalar_data_types) so I thought we can get rid of them

andreyromanof: Replaced old alias names

bader: Thanks. I hope we will not end-up with removing `cl_float` and `cl_double` types from the test. :)

bader (approved):

keryell (approved):
Thanks.

PR #324, opened 2022-04-12
Move modify/compare functions into separate file

andreyromanof:
As we already have `value_helper` in `device_global` and in #318, we can move functions for modifying and comparing values to separate file to reuse them in existing tests and future tests.

Depends on:
- #322

### DIFF
--- a/tests/common/value_helper.h
+++ b/tests/common/value_helper.h
@@ -38,8 +38,9 @@ inline void change_val(ArrayT<LeftArrT, LeftArrN>& left,
 }
 
 template <typename LeftArrT, typename RightNonArrT>
-inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT>> change_val(
-    LeftArrT& left, const RightNonArrT& right) {
+inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT> &&
+                                 !has_subscript_operator_v<RightNonArrT>>
+change_val(LeftArrT& left, const RightNonArrT& right) {
   for (size_t i = 0; i < left.size(); ++i) {
     left[i] = right;
   }
@@ -47,11 +48,11 @@ inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT>> change_val(
 
 template <typename LeftArrT, typename RightArrT>
 inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT> &&
-                                 has_subscript_operator_v<LeftArrT>>
+                                 has_subscript_operator_v<RightArrT>>
 change_val(LeftArrT& left, const RightArrT& right) {
   assert((left.size() == right.size()) && "Arrays have to be the same size");
   for (size_t i = 0; i < left.size(); ++i) {
-    left[i] = right;
+    left[i] = right[i];
   }
 }
 
@@ -84,7 +85,9 @@ inline bool are_equal(const ArrayT<LeftArrT, LeftArrN>& left,
 }
 
 template <typename LeftArrT, typename RightNonArrT>
-inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT>, bool>
+inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT> &&
+                                     !has_subscript_operator_v<RightNonArrT>,
+                                 bool>
 are_equal(const LeftArrT& left, const RightNonArrT& right) {
   for (size_t i = 0; i < left.size(); ++i) {
     if (left[i] != right) return false;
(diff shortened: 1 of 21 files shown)


tests/common/value_helper.h
```diff
@@ -0,0 +1,120 @@
...
+
+template <typename LeftArrT, typename RightArrT>
+inline typename std::enable_if_t<has_subscript_operator_v<LeftArrT> &&
+                                     has_subscript_operator_v<RightArrT>,
+                                 bool>
+are_equal(const LeftArrT& left, const RightArrT& right) {
+  assert((left.size() == right.size()) && "Arrays have to be the same size");
+  for (size_t i = 0; i < left.size(); ++i) {
+    if (left[i] != right) return false;
+  }
```
yuriykoch, line 106:
```suggestion
  assert((left.size() == right.size()) && "Arrays have to be the same size");
  for (size_t i = 0; i < left.size(); ++i) {
    if (left[i] != right[i]) return false;
  }
```

tests/common/value_helper.h
```diff
@@ -0,0 +1,120 @@
...
+//  This file contains helper functions for modifying and comparing values,
+//  arrays, and objects that implements operator[]
+//
+*******************************************************************************/
+
+#ifndef __SYCLCTS_TESTS_COMMON_VALUE_HELPER_H
+#define __SYCLCTS_TESTS_COMMON_VALUE_HELPER_H
+#include "../../util/type_traits.h"
+
+namespace value_helper {
```
yuriykoch, line 17:
```suggestion
#include "../../util/type_traits.h"

#include <cassert>

namespace value_helper {
```
BTW, just as an option - we might use `REQUIRE()` from Catch2 for preconditions check at runtime instead of `assert` to avoid dependency on `NDEBUG` macro and to have more clear output on which exact test case triggered a failure

keryell: Good idea!

andreyromanof: I have a few concerns about this solution:
- If `value_helper` functions will be used outside of a `TEST_CASE` scope, an exception will be thrown
- Do we really want to depend from catch2 in places where it can be easily avoided?

yuriykoch: Oops, sorry, looks like I need to clarify. I do not suggest to remove `assert` from entire codebase. I'm just assuming this specific code is designed to be used within `TEST_CASE` scope.

andreyromanof: I didn't intend to use this code only with catch2, so I think it's not a good idea to add a dependency from the `TEST_CASE` scope.

tests/common/value_helper.h
```diff
@@ -0,0 +1,122 @@
...
+//
+*******************************************************************************/
+
+#ifndef __SYCLCTS_TESTS_COMMON_VALUE_HELPER_H
+#define __SYCLCTS_TESTS_COMMON_VALUE_HELPER_H
+#include "../../util/type_traits.h"
+
+#include <cassert>
+
+namespace value_helper {
```
psalz, line 19: IMO `value_helper` is too nondescript. I realize that (some of) these helpers also work on scalars, but the only reason to use them would be in a generic context where values can be array-like, right? So maybe something like `array_value_helper` or `vector_value_helper` would be clearer?

tests/common/value_helper.h
```diff
@@ -0,0 +1,122 @@
...
+
+template <typename T, size_t N>
+using ArrayT = T[N];
+
+// Modify functions
+template <typename T, size_t N>
+inline void change_val(ArrayT<T, N>& left, const T& right) {
```
psalz, line 26: Maybe call these functions `assign` instead?

tests/common/value_helper.h
andreyromanof: I agree, that `value_halper` doesn't provide enough description, but if we call it `vector_value_helper` it will not be clear that functions can also work with scalar values too. As this is not a class, maybe we can call namespace `value_operations`? It will be more clearer and accurate because the namespace actually contains functions for assigning and comparing operations for scalar and vector types.

psalz: Yes `value_operations` sounds better I think. Another idea could be something like `element_wise_operations` (where a scalar could be considered a vector with a single element). But either is fine with me.

tests/common/value_helper.h
andreyromanof: I think it makes more sense. I changed the name to `assign`

tests/common/value_helper.h
andreyromanof: Changed namespace to `value_operations`

psalz (approved):

keryell:
Waiting for https://github.com/KhronosGroup/SYCL-CTS/pull/318 to land to simplify the diff here.

andreyromanof:
> Waiting for https://github.com/KhronosGroup/SYCL-CTS/pull/318 to land to simplify the diff here.

I guess we should merge this first, to simplify #318

keryell:
Ah I messed up the dependency direction...

keryell (approved):
Thanks!

PR #326, opened 2022-04-12
Tests implementation for SYCL2020 `local_accessor` constructors 

andreyromanof:
This provides tests for `local_accessor` constructors

Depends on:
- #318 

[Link](https://github.com/KhronosGroup/SYCL-CTS/pull/326/files/6637d4d264accb65dc30fa6ff053705c6d3d8f5e..HEAD) to related changes


tests/accessor/local_accessor_constructors.h
```diff
@@ -0,0 +1,118 @@
...
+      check_common_constructor<AccTypeT, DataT, DimensionT,
+                               sycl::access_mode::read_write,
+                               sycl::target::device>(get_acc_functor);
+    }
+  }
+}
+
+template <typename T, typename DimensionTypeT>
+class run_tests_constructors {
+  static constexpr int DimensionT = DimensionTypeT::value;
```
yuriykoch, line 88:
```suggestion
  static constexpr int Dimension = DimensionT::value;
```
Nit: it might be a bit confusing to use a `T` suffix for non-type template parameters, what do you think?

andreyromanof: I did it purpose. I wanted to make `DimensionT` looks like a template parameter, because `DimensionT` 100% depends on `typename DimensionTypeT` which is an actual template parameter. It's kind of workaround.

We can't get rid of `typename DimensionTypeT` completely and change it to `size_t DimensionT` because of limitation of `for_all_combination` (at least at this point I don't know how to do it).

yuriykoch: AFAIK the `T` suffix usually stands for type, not for template parameter - see `std::make_signed_t` for example with `snake_case` naming. My concern is that `DimensionT` looks like type, not like an integral value.

andreyromanof: Good point. 
I also checked paragraph 4.7.6.9.1. of the spec, to see naming of template parameters. It appears that accessor in spec doesn't have `T` suffix in `Dimensions` or `Target` template parameters. I will fix this issue in all accessor PRs.

andreyromanof: Done.

yuriykoch: Great! Haven't thought about SYCL 2020 spec, thanks for in-depth background check :)

keryell: Yes, the template arguments have been cleaned up relatively recently https://github.com/KhronosGroup/SYCL-Docs/pull/147

tests/accessor/local_accessor_constructors.h
```diff
@@ -0,0 +1,114 @@
...
+
+  SECTION(section_name) {
+    auto get_acc_functor = [r](sycl::buffer<DataT, Dimension>& data_buf,
+                               sycl::handler& cgh) {
+      return sycl::local_accessor<DataT, Dimension>(r, cgh);
+    };
+    if constexpr (std::is_const_v<DataT>) {
+      check_common_constructor<AccType, DataT, Dimension,
+                               sycl::access_mode::read, sycl::target::device>(
+          get_acc_functor);
```
yuriykoch, line 76: I'm a bit curious why don't we use `AccType` within `check_common_constructor` to differentiate between different use cases instead of `Target`. Or why don't we have partial specializations with `check_common_constructor` being a struct instead of function (to avoid issues with parameter overloading).
Still looks like we have a bunch of tests in open PRs already, so I suppose any updates might result in too many changes within interdependent PRs. Any thoughts?

tests/accessor/local_accessor_constructors.h
```diff
@@ -0,0 +1,114 @@
...
+  auto offset = sycl::id<Dimension>();
+
+  auto section_name =
+      get_section_name<Dimension>(type_name, "From sycl::range constructor");
+
+  SECTION(section_name) {
+    auto get_acc_functor = [r](sycl::buffer<DataT, Dimension>& data_buf,
+                               sycl::handler& cgh) {
+      return sycl::local_accessor<DataT, Dimension>(r, cgh);
+    };
```
yuriykoch, line 72: It will break in case we change the `range` used in `check_common_constructor` on line [393](https://github.com/KhronosGroup/SYCL-CTS/pull/326/files#diff-8067235ffc00be4a68e8da8448acbf789b606099c5dbc1e969df066646bd2f64R393)

andreyromanof: I added range to the function parameter to avoid this potential break.

tests/accessor/local_accessor_constructors.h
andreyromanof: These functions can be refactored, but it is too difficult to do, when there are at least 5 PRs, that all depend on one "common" file. IMO it will be easier to do when all PRs that depend on #318 will be merged, and we can provide refactoring with separate PR.

tests/accessor/local_accessor_constructors.h
```diff
@@ -0,0 +1,114 @@
...
+}
+
+template <typename DataT>
+void test_zero_dimension_buffer_constructor(const std::string& type_name) {
+  const auto section_name =
+      get_section_name<0>(type_name, "Zero dimension constructor");
+
+  SECTION(section_name) {
+    auto get_acc_functor = [](sycl::buffer<DataT, 1>& data_buf,
+                              sycl::handler& cgh) {
```
yuriykoch, line 47: Let's provide a single function `test_buffer_constructor` by using the following:
```
constexpr int BufferDimension = (Dimension == 0) ? 1 : Dimension;
```

tests/accessor/local_accessor_constructors.h
keryell: Yes.

keryell:
Waiting for https://github.com/KhronosGroup/SYCL-CTS/pull/318 to land to simplify the diff here.

keryell:
There are some merge conflicts to resolve first.

tests/accessor/local_accessor_constructors.h
andreyromanof: Same as https://github.com/KhronosGroup/SYCL-CTS/pull/326#discussion_r855750172

keryell (approved):
Thanks!

bader:
@andreyromanof, @yuriykoch, could you resolve merge conflicts, please?

andreyromanof:
Conflicts resolved.

PR #328, opened 2022-04-13
Add tests for SYCL 2020 host_accessor API

kopylovanat:
This PR depends on https://github.com/KhronosGroup/SYCL-CTS/pull/320 because of common code (merged).

PR #330, opened 2022-04-20
Add tests for SYCL 2020 local_accessor API

kopylovanat:
This PR depends on https://github.com/KhronosGroup/SYCL-CTS/pull/328 because of common code.
[Link](https://github.com/KhronosGroup/SYCL-CTS/pull/330/files/ac73ae737ff1b5b475704c2f53f7232561ce58ce..HEAD) to related changes.


tests/accessor/local_accessor_api_common.h
```diff
@@ -0,0 +1,134 @@
...
+
+              sycl::accessor res_acc(res_buf, cgh);
+              cgh.parallel_for(nd_range, [=](sycl::nd_item<dim> item) {
+                auto ref_1 = acc[0];
+
+                auto id = util::get_cts_object::id<dims>::get(1, 1, 1);
+                auto ref_2 = get_subscript_overload<T, AccT, dims>(acc, id);
+
+                res_acc[0] =
+                    std::is_same_v<decltype(ref_1), typename AccT::reference>;
```
yuriykoch, line 93: Don't we have multiple work-items here? As far as I recall there was a discussion regarding same element access from different work-items even in case we are writing the same value in it. The discussion was mentioning the article by [Hans-J. Boehm](https://hboehm.info/boehm-hotpar11.pdf.)
I mean, even the following code would be a data race:
```
                if (!std::is_same_v<decltype(ref_1), typename AccT::reference>) {
                  res_acc[0] = false;
                }
```

tests/accessor/local_accessor_api_common.h
```diff
@@ -0,0 +1,134 @@
...
+                  res_acc[0] &=
+                      value_helper::are_equal(acc_other[0], expected_val);
+                  res_acc[0] &=
+                      value_helper::are_equal(acc_other[1], changed_val);
+                }
+              });
+            })
+            .wait_and_throw();
+      }
+      CHECK(res);
```
yuriykoch, line 115: Minor: it's a pity that we are missing separate log messages for different types of failures. There would be no way to tell which device-side check resulted in failure based on `CHECK(res)` output.

tests/accessor/local_accessor_api_fp16.cpp
```diff
@@ -0,0 +1,33 @@
...
+#endif
+
+namespace local_accessor_api_fp16 {
+
+DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp, DPCPP)
+("sycl::local_accessor api. fp16 type", "[accessor]")({
+  using namespace local_accessor_api_common;
+
+  auto queue = sycl_cts::util::get_cts_object::queue();
+  if (queue.get_device().has(sycl::aspect::fp16)) {
```
vasilytric, line 25: Would you mind to add `else` branch that will say the reason of test skipping
with the following message, for example: "Device does not support double precision floating point operations"

tests/accessor/accessor_common.h
```diff
@@ -79,6 +79,28 @@ inline std::string get_section_name(const std::string& type_name,
...
+ *
+ * @tparam DimensionT Integer representing dimension
+ * @param type_name String with name of the testing type
+ * @param section_description String with human-readable description of the test
+ * @return std::string String with name for section
+ */
+template <int DimensionT>
+inline std::string get_section_name(const std::string& type_name,
+                                    const std::string& section_description) {
+  using namespace sycl_cts::get_cts_string;
```
andreyromanof, line 94: Don't need

tests/accessor/local_accessor_api_common.h
andreyromanof: Looks like we can use [`group_barier`](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_group_barrier) here.

```
myQueue.submit([&](handler & cgh) {
  cgh.parallel_for(
      nd_range<3>(range<3>(4, 4, 4), range<3>(2, 2, 2)), [=](nd_item<3> item) {
        //[kernel code]
        // Internal synchronization
        group_barrier(item.get_group());
        //[kernel code]
      });
});
```

kopylovanat: Changed `res` to bool array with result for each item.

tests/accessor/local_accessor_api_common.h
```diff
@@ -0,0 +1,208 @@
...
+                  acc, sizeof(T) * local_range.size() /* expected_byte_size*/,
+                  local_range.size() /*expected_size*/,
+                  local_range /*expected_range*/);
+
+              sycl::accessor res_acc(res_buf, cgh);
+              cgh.parallel_for(nd_range, [=](sycl::nd_item<dim> item) {
+                auto ref_1 = acc[0];
+
+                auto id = util::get_cts_object::id<dims>::get(1, 1, 1);
+                auto ref_2 = get_subscript_overload<T, AccT, dims>(acc, id);
```
yuriykoch, line 117:
```suggestion
                auto&& ref_2 = get_subscript_overload<T, AccT, dims>(acc, id);
```
Is it what you've meant? Note that `auto` would resolve to the non-const non-reference variable no matter of `get_subscript_overload` return type.

tests/accessor/local_accessor_api_common.h
```diff
@@ -0,0 +1,208 @@
...
+                size_t item_id = item.get_global_linear_id();
+                res_acc[0, item_id] =
+                    std::is_same_v<decltype(ref_1), typename AccT::reference>;
+                res_acc[1, item_id] =
+                    std::is_same_v<decltype(ref_2), typename AccT::reference>;
+                if constexpr (!std::is_const_v<T>) {
+                  value_helper::change_val(ref_1, expected_val);
```
yuriykoch, line 124: AFAIK given the statement above:
```
auto ref_1 = acc[0];
```
this check doesn't correspond to the `INFO("return type for operator[](id<Dimensions> index)");` below

tests/accessor/local_accessor_api_common.h
```diff
@@ -0,0 +1,208 @@
...
+          INFO("result for operator[]](size_t index)");
+          for (size_t i = 0; i < global_range.size(), i++) CHECK(res[3, i]);
+        }
+        {
+          INFO("return type for get_multi_ptr<sycl::access::decorated::no>()");
+          for (size_t i = 0; i < global_range.size(), i++) CHECK(res[4, i]);
+        }
+        {
+          INFO("result for get_multi_ptr<sycl::access::decorated::no>()");
+          for (size_t i = 0; i < global_range.size(), i++) CHECK(res[5, i]);
```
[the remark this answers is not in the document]
kopylovanat: Changed to enum class of checks https://github.com/KhronosGroup/SYCL-CTS/pull/330/commits/c82e663913e22dd80c5653d26ecdaff2dc55d356

tests/accessor/accessor_common.h
```diff
@@ -0,0 +1,561 @@
...
+  STATIC_CHECK(std::is_same_v<typename AccT::const_reference, const T&>);
+  STATIC_CHECK(std::is_same_v<typename AccT::size_type, size_t>);
+}
+
+template <typename T, typename AccT, int dims>
+T& get_subscript_overload(const AccT& accessor, size_t index) {
+  if constexpr (dims == 1) return accessor[index];
+  if constexpr (dims == 2) return accessor[index][index];
+  if constexpr (dims == 3) return accessor[index][index][index];
+}
```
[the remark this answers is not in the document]
kopylovanat: Changed to `decltype(auto)`

keryell:
Waiting for https://github.com/KhronosGroup/SYCL-CTS/pull/328 to land first. I was wondering why the test was about `host_accessor`... :-)

bader:
> Waiting for #328 to land first. I was wondering why the test was about `host_accessor`... :-)

Done. @kopylovanat, please, update the branch to resolve merge conflicts.

PR #331, opened 2022-04-21
Add unified macros for detecting SYCL implementations

psalz:
Defines a new macro of the form `SYCL_CTS_COMPILING_WITH_<IMPL>` when compiling with a given SYCL implementation. `<IMPL>` can currently be one of `COMPUTECPP`, `DPCPP` or `HIPSYCL`.

I've also removed `ComputeCpp` as the default value for the `SYCL_IMPLEMENTATION` CMake option; IMO we should not be favoring any particular implementation over another.


keryell (approved):

psalz:
I'm very confused... [It seems like](https://github.com/KhronosGroup/SYCL-CTS/runs/6571204721?check_suite_focus=true#step:8:14575) the detection macro is not correctly being set for ComputeCpp, however I can neither reproduce this locally on my machine nor when running inside the *actual CI Docker container*.

PR #333, opened 2022-04-27
Add tag factory for accessor tests

andreyromanof:
This provides tag factory functions for existing and future accessor tests

### DIFF
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -139,8 +139,11 @@ inline auto add_vectors_to_type_pack(StrNameType type_name) {
                                   "vec<" + type_name + ", 16>");
 }
 
-template <accessor_type AccT>
-struct tag_factory {};
+template <accessor_type AccType>
+struct tag_factory {
+  static_assert(AccType != AccType,
+                "There is no tag support for such accessor type")
+};
 
 /**
  * @brief Function helps to get TagT corresponding to AccessMode and Target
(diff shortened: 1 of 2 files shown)


tests/accessor/accessor_common.h
```diff
@@ -139,30 +139,70 @@ inline auto add_vectors_to_type_pack(StrNameType type_name) {
                                   "vec<" + type_name + ", 16>");
 }
 
+template <accessor_type AccT>
+struct tag_factory {};
```
yuriykoch, line 143:
```suggestion
template <accessor_type AccType>
struct tag_factory {
  static_assert(AccType != AccType, "There is no tag support for such accessor type");
};
```

1. We can avoid `tag_factory<accessor_type::local_accessor>` in such way at the cost of entire struct instantiation support
2. Let's use `AccType` for naming consistency through the header

andreyromanof: Good idea. Applied it.

tests/accessor/accessor_common.h
```diff
@@ -139,30 +139,64 @@ inline auto add_vectors_to_type_pack(StrNameType type_name) {
                                   "vec<" + type_name + ", 16>");
 }
 
+template <accessor_type AccType>
+struct tag_factory {
```
yuriykoch, line 143: Would you mind to move the `accessor_type` enum declaration before it's used? There would be compilation failure on linux otherwise.

tests/accessor/accessor_common.h
```diff
@@ -139,30 +139,64 @@ inline auto add_vectors_to_type_pack(StrNameType type_name) {
...
+  static_assert(AccType != AccType,
+                "There is no tag support for such accessor type")
```
yuriykoch, line 145:
```suggestion
  static_assert(AccType != AccType,
                "There is no tag support for such accessor type");
```
Here and below :)

tests/accessor/accessor_common.h
andreyromanof: Fixed

tests/accessor/accessor_common.h
andreyromanof: Fixed

keryell (approved):

keryell:
Thanks.
Sorry for the delay but all the SYCL committee is swamped by the preparation of SYCLcon happening next week.
In the meantime, you can also register to SYCLcon since it is free to attend remotely. :-)

PR #335, opened 2022-05-05
Test for SYCL-2020 accessor template default values tests

vasilytric:
The test verify that:
- Generic `accessor` has the following template parameters default values:
   - Dimensions
   - sycl::access_mode
   - sycl::target
   - sycl::access::placeholder
- `host_accessor` has the following template parameters default values:
   - Dimensions
   - sycl::access_mode
- `local_accessor` has the following template parameter default value:
   - Dimensions


tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+      "Verify default value for accessTarget template parameter for " +
+      type_name + " data type. [generic accessor]";
+  SECTION(section_name) {
+    REQUIRE(std::is_same_v<
+            sycl::accessor<T, expected_dims,
+                           std::is_const_v<T> ? sycl::access_mode::read
+                                              : sycl::access_mode::read_write>,
+            sycl::accessor<T, expected_dims,
+                           std::is_const_v<T> ? sycl::access_mode::read
+                                              : sycl::access_mode::read_write,
```
yuriykoch, line 112: What about using
```
constexpr auto expected_mode = std::is_const_v<T> ? sycl::access_mode::read : sycl::access_mode::read_write>;
```

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+  }
+  section_name =
+      "Verify default value for isPlaceholder template parameter for " +
+      type_name + " data type. [generic accessor]";
+  SECTION(section_name) {
+    REQUIRE(std::is_same_v<
+            sycl::accessor<T, expected_dims,
+                           std::is_const_v<T> ? sycl::access_mode::read
+                                              : sycl::access_mode::read_write,
+                           expected_target>,
```
yuriykoch, line 123: It will not cover the case of implementation having different default values for different specializations.
Are you sure we shouldn't cover all possible combinations of template parameters (dimensions and access modes)?

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+ * @tparam T Current data type
+ * @param type_name Current data type string representation
+ */
+template <typename T>
+class run_tests {
+ public:
+  void operator()(const std::string& type_name) {
+    test_for_generic_acc<T>(type_name);
+    test_for_host_acc<T>(type_name);
+    test_exception_for_local_acc<T>(type_name);
```
yuriykoch, line 143:
```suggestion
    const auto const_type_name = "const " + type_name;
    test_for_generic_acc<T>(type_name);
    test_for_generic_acc<const T>(const_type_name);
    test_for_host_acc<T>(type_name);
    test_for_host_acc<const T>(const_type_name);
    test_exception_for_local_acc<T>(type_name);
    test_exception_for_local_acc<const T>(const_type_name);
```
Did I get the intent right?

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+    sycl::access::placeholder::false_t;
+
+/**
+ * @brief Provides functor that lets verify that local_accessor's template
+ *        parameters have expected default values
+ * @tparam T Current data type
+ * @param type_name Current data type string representation
+ */
+template <typename T>
+void test_exception_for_local_acc(const std::string& type_name) {
```
yuriykoch, line 30:
```suggestion
template <typename T>
void test_for_local_acc(const std::string& type_name) {
```
Looks like it would be a bit more consistent this way

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+//  SYCL 2020 Conformance Test Suite
+//
+//  This file provides functions for tests on accessor default values.
+//
+*******************************************************************************/
+#ifndef SYCL_CTS_ACCESSOR_DEFAULT_VALUES_H
+#define SYCL_CTS_ACCESSOR_DEFAULT_VALUES_H
+
+#include "accessor_common.h"
+#include "catch2/catch_test_macros.hpp"
```
yuriykoch, line 12: We are currently using Catch2 as a major architecture dependency, so not sure if we actually need this one

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+
+#include <type_traits>
+
+namespace accessor_default_values_test {
+
+constexpr int expected_dims = 1;
+constexpr sycl::target expected_target = sycl::target::device;
+constexpr sycl::access::placeholder expected_placeholder =
+    sycl::access::placeholder::false_t;
```
yuriykoch, line 21: Just as a suggestion:
```suggestion
namespace detail {
constexpr int expected_dims = 1;
constexpr auto expected_target = sycl::target::device;
constexpr auto expected_placeholder = sycl::access::placeholder::false_t;
}  // namespace detail
```
By using `detail` namespace we are saying that we are not going to use these values outside of this header

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,149 @@
...
+  SECTION(section_name) {
+    REQUIRE(
+        std::is_same_v<sycl::accessor<T, expected_dims>,
+                       sycl::accessor<T, expected_dims,
+                                      std::is_const_v<T>
+                                          ? sycl::access_mode::read
+                                          : sycl::access_mode::read_write>>);
+  }
+  section_name = "Verify default value for accessMode template parameter for " +
+                 type_name + " data type. [generic accessor]";
```
yuriykoch, line 93: Doesn't this section duplicate the one above?

tests/accessor/accessor_default_values.h
andreyromanof: +1

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,168 @@
...
+  static constexpr int Dimension = DimensionT::value;
+  static constexpr sycl::target Target = TargetT::value;
+
+ public:
+  void operator()(const std::string& type_name,
+                  const std::string& access_mode_name,
+                  const std::string& target_name) {
+    auto section_name = get_section_name<Dimension>(
+        type_name, access_mode_name, target_name,
+        "Verify default value for dimensions template parameter.");
```
andreyromanof, line 67: Maybe we should add the tag `[host_accessor]` in these section names like you did with `local_accessor`? As an alternative, we can remove the tag from `local_accessor` for consistency of code.

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,168 @@
...
+ */
+template <typename T>
+class test_for_local_acc {
+ public:
+  void operator()(const std::string& type_name) {
+    auto section_name =
+        "Verify default value for dimensions template parameter for " +
+        type_name + " data type. [local_accessor]";
+    SECTION(section_name) {
+      REQUIRE(std::is_same_v<sycl::local_accessor<T>,
```
andreyromanof, line 41: I remember we used `CHECK` recently. Are there some advantages to using `REQUIRE`?

vasilytric: we don't have any advantages to use `REQUIRE`, but there is no advantages to use `CHECK` becase we havejust one verification in each section

tests/accessor/accessor_default_values.h
vasilytric: yes, removed [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9L92-L101)
and [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R107-R114) we have only one this section

tests/accessor/accessor_default_values.h
vasilytric: namespace `detail` has been added [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R17)

tests/accessor/accessor_default_values.h
vasilytric: this include has been removed [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9L12)

tests/accessor/accessor_default_values.h
vasilytric: `test_exception_for_local_acc` has been renamed [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R34)

tests/accessor/accessor_default_values.h
vasilytric: yes, coverage for const types has been added [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R157-R162)

tests/accessor/accessor_default_values.h
vasilytric: coverage for different `access_mode`, `dimension` and `target has been implemented [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/c040cc7560db878fb467d794a504c6e5e20a0d0e#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R146-R155)

tests/accessor/accessor_default_values.h
vasilytric: yes, it's a good tip!
accessor type has been added to section name in this commit: 8877b5d1b09868da3efc1224ef0df9df1511c63f

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,174 @@
...
+                                    "template parameter. [host_accessor]");
+    SECTION(section_name) {
+      REQUIRE(std::is_same_v<sycl::host_accessor<T>,
+                             sycl::host_accessor<T, detail::expected_dims>>);
+    }
+    section_name =
+        get_section_name<Dimension>(type_name, access_mode_name, target_name,
+                                    "Verify default value for accessMode "
+                                    "template parameter. [host_accessor]");
+    SECTION(section_name) {
```
yuriykoch, line 77: It will give a same section name for different `Dimension` values.
See [Catch2 limitations]( https://github.com/catchorg/Catch2/blob/devel/docs/limitations.md#sections-nested-in-loops) for details. Also please be aware that you might have section names prototypes using https://github.com/KhronosGroup/SYCL-CTS/pull/337

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,174 @@
...
+  void operator()(const std::string& type_name,
+                  const std::string& access_mode_name,
+                  const std::string& target_name) {
+    auto section_name =
+        get_section_name<Dimension>(type_name, access_mode_name, target_name,
+                                    "Verify default value for dimensions "
+                                    "template parameter. [host_accessor]");
+    SECTION(section_name) {
+      REQUIRE(std::is_same_v<sycl::host_accessor<T>,
+                             sycl::host_accessor<T, detail::expected_dims>>);
```
yuriykoch, line 71: We are not using `Dimension` here, but we would run this check for every `Dimension` value. Is it intended?
Maybe we might have something like the folllowing here?
```
constexpr bool should_check_dimension =
    (Dimension == detail::expected_dims);

if constexpr (should_check_dimension ) {
```

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,174 @@
...
+    for_all_combinations<test_for_host_acc, T>(access_modes, dimensions,
+                                               targets, type_name);
+    test_for_local_acc<T>(type_name);
+    // Run test with const data type
+    const auto const_type_name = "const " + type_name;
+    for_all_combinations<test_for_generic_acc, const T>(
+        access_modes, dimensions, targets, const_type_name);
+    for_all_combinations<test_for_host_acc, const T>(access_modes, dimensions,
+                                                     targets, const_type_name);
+    test_for_local_acc<const T>(const_type_name);
```
yuriykoch, line 168: Nit; what do you think about adding a blank line?

```suggestion
    // Run test with non const data type
    for_all_combinations<test_for_generic_acc, T>(access_modes, dimensions,
                                                  targets, type_name);
    for_all_combinations<test_for_host_acc, T>(access_modes, dimensions,
                                               targets, type_name);
    test_for_local_acc<T>(type_name);

    // Run test with const data type
    const auto const_type_name = "const " + type_name;
    for_all_combinations<test_for_generic_acc, const T>(
        access_modes, dimensions, targets, const_type_name);
    for_all_combinations<test_for_host_acc, const T>(access_modes, dimensions,
                                                     targets, const_type_name);
    test_for_local_acc<const T>(const_type_name);
```

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,174 @@
...
+    // will not compile
+    const auto access_modes = get_access_modes();
+    const auto dimensions = get_all_dimensions();
+    const auto targets = get_targets();
+
+    // Run test with non const data type
+    for_all_combinations<test_for_generic_acc, T>(access_modes, dimensions,
+                                                  targets, type_name);
+    for_all_combinations<test_for_host_acc, T>(access_modes, dimensions,
+                                               targets, type_name);
```
yuriykoch, line 160: Looks like we don't need to iterate over `access_modes` and `targets` to check default template parameters of `host_accessor`

tests/accessor/accessor_default_values.h
vasilytric: this is by design because we may not iterate over default values of `Dimension`, `target`, e.t.c. and in this case this verifications will be skipped

yuriykoch: You are right, it definitely makes things much harder.

Still I'm bit worried that we don't have any simple logic to switch off repeating checks in such case.
For example, if `sycl::accessor<sycl::half>` doesn't have proper default dimension value for some reason - we don't need to print out an assertion failure for each possible combination of the rest template parameters.
How do you think, might something like this be handy?

```
namespace detail {
template <template <typename...> class, typename ...>
struct run_once_per {
  /**
   * @brief Returns true only once per application run for each instantiation of
   *        `run_once_per` class
   * @details Uses Meyer's singleton to have lazy initialization with no static
   *          initialization order fiasco possible; not thread-safe
   */
  static bool should_run() {
    static bool triggered = false;
    const bool result = triggered;
    return result ^ (triggered = true);
  }
};
}

template <typename T, typename AccessModeT, typename DimensionT,
          typename TargetT>
class test_for_host_acc {
  template <typename ... Types>
  using run_once_per = detail::run_once_per<test_for_host_acc, T, Types...>;
  using run_once = run_once_per<>;
  ...
  if (run_once::should_run()) {
    ...
    REQUIRE(std::is_same_v<sycl::host_accessor<T>,
                             sycl::host_accessor<T, detail::expected_dims>>);
  }
  ...
};

template <typename T, typename AccessModeT, typename DimensionT,
          typename TargetT>
class test_for_generic_acc {
  template <typename ... Types>
  using run_once_per = detail::run_once_per<test_for_generic_acc, T, Types...>;
  using run_once = run_once_per<>;
  ...
  if (run_once::should_run()) {
    ...
    REQUIRE(std::is_same_v<sycl::host_accessor<T>,
                             sycl::host_accessor<T, detail::expected_dims>>);
  }
  if (run_once_per<DimensionT>::should_run()) {
    ...
    REQUIRE(std::is_same_v<
              sycl::accessor<T, Dimension>,
              sycl::accessor<T, Dimension, detail::expected_mode<T>>>);
  }
  if (run_once_per<DimensionT, AccessModeT>::should_run()) {
    ...
    REQUIRE(std::is_same_v<sycl::accessor<T, Dimension, AccessMode>,
                             sycl::accessor<T, Dimension, AccessMode,
                                            detail::expected_target>>);
  }
  ...
};
```

yuriykoch: As a follow-up for the previous suggestion - looks like it would be better to state dependency on `T` explicitly:
```
template <typename T, typename AccessModeT, typename DimensionT,
          typename TargetT>
class test_for_generic_acc {
  template <typename ... Types>
  using run_once_per = detail::run_once_per<test_for_generic_acc, Types...>;
  ...
  if (run_once_per<T>::should_run()) {
    ...
    REQUIRE(std::is_same_v<sycl::host_accessor<T>,
                             sycl::host_accessor<T, detail::expected_dims>>);
  }
  if (run_once_per<T, DimensionT>::should_run()) {
    ...
    REQUIRE(std::is_same_v<
              sycl::accessor<T, Dimension>,
              sycl::accessor<T, Dimension, detail::expected_mode<T>>>);
  }
```

vasilytric: > For example, if sycl::accessor<sycl::half> doesn't have proper default dimension value

This should be explicitly state into SYCL-2020 spec., until all specializations should have specified default values, so now we should werify it

> How do you think, might something like this be handy?

We may use such solution, but it looks some strange :)
We should add not a little code for skip some the same verifications, but this could be useful

tests/accessor/accessor_default_values.h
vasilytric: no, we will have the different section names with different `Dimension` values. The `Dimension` will be used [here](https://github.com/KhronosGroup/SYCL-CTS/pull/337/files#diff-8067235ffc00be4a68e8da8448acbf789b606099c5dbc1e969df066646bd2f64R154)

tests/accessor/accessor_default_values.h
vasilytric: this comment has been applied in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/fdc3712565b19f017f6c64f03c15c0edfeaf25e0

tests/accessor/accessor_default_values.h
vasilytric: blank line has been added [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/e2e50aa7cd5d3b4f749e1d1c3149f1c8a5d2c727#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R163)

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,174 @@
...
+
+#include <type_traits>
+
+namespace accessor_default_values_test {
+
+namespace detail {
+constexpr int expected_dims = 1;
+constexpr sycl::target expected_target = sycl::target::device;
+constexpr sycl::access::placeholder expected_placeholder =
+    sycl::access::placeholder::false_t;
```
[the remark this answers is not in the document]
vasilytric: type deducing with `auto` has been added [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/e2e50aa7cd5d3b4f749e1d1c3149f1c8a5d2c727#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R20-R21)

tests/accessor/accessor_default_values.h
vasilytric: @andreyromanof @kopylovanat What do you think about this suggesion?

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,170 @@
...
+};
+
+/**
+ * @brief Provides functor that lets verify that host_accessor's template
+ *        parameters have expected default values
+ * @tparam T Current data type
+ * @param type_name Current data type string representation
+ */
+
+template <typename T, typename DimensionTI>
```
andreyromanof, line 55:
```suggestion
template <typename T, typename DimensionT>
```

tests/accessor/accessor_default_values.h
andreyromanof: I guess we can easily skip extra checks by using `if constexpr`.
Something like this:
```
if constexpr (Dimension == /*something*/) { //preform one-time check}
```
but I'm also okay with the current solution.

tests/accessor/accessor_default_values.h
vasilytric: This comment has been applied in this commit: fdc3712565b19f017f6c64f03c15c0edfeaf25e0

tests/accessor/accessor_default_values_fp16.cpp
```diff
@@ -0,0 +1,37 @@
...
+namespace accessor_exceptions_test_fp16 {
+using namespace sycl_cts;
+
+DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp, DPCPP)
+("Accessors constructor default values test fp16 types.", "[accessor]")({
+#ifndef SYCL_CTS_ENABLE_FULL_CONFORMANCE
+  run_tests<sycl::half>{}("sycl::half");
+#else
+  for_type_vectors_marray<run_tests, sycl::half>("sycl::half");
+#endif  // SYCL_CTS_ENABLE_FULL_CONFORMANCE
```
bader, line 34: I think 
```c++
#ifdef A
  X
#else
  Y
#endif
```
is more readable than
```c++
#ifndef A
  Y
#else
  X
#endif
```
Please, prefer the version w/o negation in condition to reduce mental load of the code.

tests/accessor/accessor_default_values.h
```diff
@@ -0,0 +1,253 @@
...
+void verify_host_acc_dim_value(const std::string& type_name) {
+  auto section_name = get_section_name<detail::expected_dims>(
+      type_name,
+      "Verify default value for dimensions "
+      "template parameter. [local_accessor]");
+  SECTION(section_name) {
+    REQUIRE(std::is_same_v<sycl::local_accessor<T>,
+                           sycl::local_accessor<T, detail::expected_dims>>);
+  }
+}
```
bader, line 62: Shouldn't this function test host accessor? If so, there are a lot of copy-paste errors where `local` has to be replaced with `host`.

keryell (changes requested):

tests/accessor/accessor_default_values.h
vasilytric: > Shouldn't this function test host accessor? 

Yes, replaced with `sycl::host_accessor` [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/20bb630b11d3e25acee89b37b3aacde18ae3fda1#diff-9ed3eae8f2889bc04dfc0fad247a8fa4e381e1fff4144e0f1a23670d108f6fe9R60-R61)

> If so, there are a lot of copy-paste errors where local has to be replaced with host.

Where? I think that in all other cases all is good

tests/accessor/accessor_default_values_fp16.cpp
```diff
@@ -0,0 +1,37 @@
...
+using namespace accessor_default_values_test;
+using namespace accessor_tests_common;
+#endif
+
+namespace accessor_exceptions_test_fp16 {
+using namespace sycl_cts;
+
+DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp, DPCPP)
+("Accessors constructor default values test fp16 types.", "[accessor]")({
+#ifndef SYCL_CTS_ENABLE_FULL_CONFORMANCE
```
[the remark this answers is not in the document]
vasilytric: Yes, this verification has been added [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/20bb630b11d3e25acee89b37b3aacde18ae3fda1#diff-67b2986690dd6b47fa6c9bada44ae452cd8e8b737ef97ca87ee0a9bac185cc8bR30-R36) also for `fp64`, [here](https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/20bb630b11d3e25acee89b37b3aacde18ae3fda1#diff-5a71685afd434f6fdf9df841b26a7ba46dfeca510f969345464edf1e5c3aa5a8R30-R36)

tests/accessor/accessor_default_values.h
bader: > > If so, there are a lot of copy-paste errors where local has to be replaced with host.
> 
> Where? I think that in all other cases all is good

In the part I selected for this comment.

bader (approved):

keryell (approved):
Thanks.

bader:
@vasilytric, please, resolve merge conflicts.

vasilytric:
> @vasilytric, please, resolve merge conflicts.

tests/accessor/accessor_default_values_fp64.cpp
```diff
@@ -0,0 +1,45 @@
...
+("Accessors constructor default values test fp64 types.", "[accessor]")({
+  auto queue = sycl_cts::util::get_cts_object::queue();
+  if (!queue.get_device().has(sycl::aspect::fp64)) {
+    WARN(
+        "Device does not support double precision floating point operations. "
+        "Skipping the test case.");
+    return;
+  }
+
+#ifdef SYCL_CTS_ENABLE_FULL_CONFORMANCE
```
bader, line 38:
```suggestion
#if SYCL_CTS_ENABLE_FULL_CONFORMANCE
```

tests/accessor/accessor_default_values_fp64.cpp
```diff
@@ -0,0 +1,45 @@
...
+//  for generic accessor, host_accessor and local_accessor with double type.
+//
+*******************************************************************************/
+
+#include "../common/common.h"
+#include "../common/disabled_for_test_case.h"
+
+// FIXME: re-enable when sycl::accessor is implemented
+#if !defined(__HIPSYCL__) && !defined(__COMPUTECPP__) && \
+    !defined(__SYCL_COMPILER_VERSION)
```
bader, line 17:
```suggestion
#if !SYCL_CTS_COMPILING_WITH_HIPSYCL && !SYCL_CTS_COMPILING_WITH_COMPUTECPP && \
    !SYCL_CTS_COMPILING_WITH_DPCPP
```

bader (changes requested):
Please, apply to all new files.

tests/accessor/accessor_default_values_fp64.cpp
vasilytric: I've applied this comment in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/66c0562df41dca778c91018c6669d171f18b0807

tests/accessor/accessor_default_values_fp64.cpp
vasilytric: I've applied this comment in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/335/commits/66c0562df41dca778c91018c6669d171f18b0807

tests/accessor/host_accessor_constructors_fp64.cpp
```diff
@@ -24,10 +24,17 @@ DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp, DPCPP)
...
+    return;
+  }
+
+#if SYCL_CTS_ENABLE_FULL_CONFORMANCE
   for_type_vectors_marray<run_host_constructors_test, double>("double");
+#else
+  run_host_constructors_test<double>{}("double");
 #endif  // SYCL_CTS_ENABLE_FULL_CONFORMANCE
-})
+});
```
bader, line 39: To be honest, I'm not sure if `;` is needed after `DISABLED_FOR_TEST_CASE` macro.

bader (approved):

tests/accessor/host_accessor_constructors_fp64.cpp
vasilytric: We have an example of using that has been provided by Philip Salzmann and this example contained the `;`
This example can be found [here](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/common/disabled_for_test_case.h#L30)

bader: According to my understanding the macro is expanded into either a function definition or no-op. Does it sound right to you? Both of these bases shouldn't require `;` at the end.
It's minor issue as extra `;` doesn't break anything.

vasilytric: > Does it sound right to you?

Yes, I think that I am agreed with you

> It's minor issue as extra ; doesn't break anything.

So, let's don't delete extra `;`

PR #340, opened 2022-05-16
Bump dpcpp version in ci

againull:
This version of the DPCPP contains a fix for get_native_mem() return
type.

PR #338, opened 2022-05-13
Move definition of fract to resolve nextafter dependency

steffenlarsen:
The definition of fract in the math_reference utils rely on nextafter. nextafter is however not declared for half until later in the header, so these changes move the definitions of fract to the source file, making the declarations of nextafter visible.

This PR also enables math_builtin_api for DPCPP after being disabled in https://github.com/KhronosGroup/SYCL-CTS/pull/341.

PR #332, opened 2022-04-27
Update return type for get_native_mem() per SYCL 2020

againull:
Per "C.6. Interoperability with the OpenCL API" of the SYCL 2020
specification native type for sycl::buffer is std::vector<cl_mem>.

Update test accordingly.

Signed-off-by: Artur Gainullin <artur.gainullin@intel.com>

PR #341, opened 2022-05-17
Bump dpcpp version in ci

psalz:
Supersedes #340.

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -59,7 +59,7 @@ jobs:
           # - sycl-impl: computecpp
           # version: 2.8.0
           - sycl-impl: dpcpp
-            version: 977f22d
+            version: caa696f
           - sycl-impl: hipsycl
             version: 3d8b1cd
     steps:
@@ -98,7 +98,7 @@ jobs:
           - sycl-impl: computecpp
             version: 2.8.0
           - sycl-impl: dpcpp
-            version: 977f22d
+            version: caa696f
           - sycl-impl: hipsycl
             version: 3d8b1cd
     env:


bader (approved):

psalz:
[It seems like there is a regression](https://github.com/KhronosGroup/SYCL-CTS/runs/6469267928?check_suite_focus=true#step:8:25295) in the `backend_return_t<backend::opencl, accessor<..., target::device>`, as this now returns a `std::vector<cl_mem>` instead of a `__global T*`. Should we just disable this test category for now?

bader:
> Should we just disable this test category for now?

Sounds good to me.

@againull, could you file a bug report for DPC++ to fix this regression, please?

steffenlarsen:
@psalz - Is this the error at hand?

```
../tests/host_task/host_task_interop_api.cpp:106:20: error: no viable conversion from 'backend_return_t<(sycl::backend)'\x01', buffer<unsigned long, 1>>' (aka 'vector<_cl_mem *>') to 'cl_mem' (aka '_cl_mem *')
```

If so, it seems to be `backend_return_t` of `buffer` and not `accessor`, which according to the [spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:interfacing-with-opencl) should be `std::vector<cl_mem>`, though I believe that was a recent change. In DPC++ the behavior can be switched back to the old return types by unsetting `SYCL2020_CONFORMANT_APIS`. Keep in mind though that this also changes the `backend_return_t` of `event` back to a single `cl_event` too.

psalz:
> @psalz - Is this the error at hand?
> 
> ```
> ../tests/host_task/host_task_interop_api.cpp:106:20: error: no viable conversion from 'backend_return_t<(sycl::backend)'\x01', buffer<unsigned long, 1>>' (aka 'vector<_cl_mem *>') to 'cl_mem' (aka '_cl_mem *')
> ```
> 
> If so, it seems to be `backend_return_t` of `buffer` and not `accessor`, which according to the [spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:interfacing-with-opencl) should be `std::vector<cl_mem>`, though I believe that was a recent change. In DPC++ the behavior can be switched back to the old return types by unsetting `SYCL2020_CONFORMANT_APIS`. Keep in mind though that this also changes the `backend_return_t` of `event` back to a single `cl_event` too.

Hmm, looks like you're right. While the [line in question](https://github.com/KhronosGroup/SYCL-CTS/blob/224f0d54dc2da06ed57f1c013dfe1e5b25d69062/tests/host_task/host_task_interop_api.cpp#L106) passes an accessor (not a buffer) into `get_native_mem`, [apparently](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interfaces.hosttasks.overview) this returns a value of type `backend_return_t<Backend, buffer>`, not `backend_return_t<Backend, accessor>`. In any case, the test is currently broken and needs fixing, I suggest we leave the filter in place and do this in a separate change.

againull:
> > @psalz - Is this the error at hand?
> > ```
> > ../tests/host_task/host_task_interop_api.cpp:106:20: error: no viable conversion from 'backend_return_t<(sycl::backend)'\x01', buffer<unsigned long, 1>>' (aka 'vector<_cl_mem *>') to 'cl_mem' (aka '_cl_mem *')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If so, it seems to be `backend_return_t` of `buffer` and not `accessor`, which according to the [spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:opencl:interfacing-with-opencl) should be `std::vector<cl_mem>`, though I believe that was a recent change. In DPC++ the behavior can be switched back to the old return types by unsetting `SYCL2020_CONFORMANT_APIS`. Keep in mind though that this also changes the `backend_return_t` of `event` back to a single `cl_event` too.
> 
> Hmm, looks like you're right. While the [line in question](https://github.com/KhronosGroup/SYCL-CTS/blob/224f0d54dc2da06ed57f1c013dfe1e5b25d69062/tests/host_task/host_task_interop_api.cpp#L106) passes an accessor (not a buffer) into `get_native_mem`, [apparently](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interfaces.hosttasks.overview) this returns a value of type `backend_return_t<Backend, buffer>`, not `backend_return_t<Backend, accessor>`. In any case, the test is currently broken and needs fixing, I suggest we leave the filter in place and do this in a separate change.

I have a PR to fix tests/host_task/host_task_interop_api.cpp test:
https://github.com/KhronosGroup/SYCL-CTS/pull/332

psalz:
> I have a PR to fix tests/host_task/host_task_interop_api.cpp test: #332

Ah I see, so this is why we're updating DPC++ in the first place!

It seems like there's also an unrelated issue with the `math_builtin_api` tests, I've disabled those as well for now.

psalz:
> > It seems like there's also an unrelated issue with the `math_builtin_api` tests, I've disabled those as well for now.
> 
> This should be fixed with #338.

Perfect, then @steffenlarsen and @againull please back-merge or rebase and remove the two categories from the CI filter again!

againull:
@psalz Thanks a lot for the help!

PR #337, opened 2022-05-12
Add section name builder

yuriykoch:
Using a builder with fluent interface makes any combination of parameters
explicit and possible. It might be useful for accessor conversion coverage
for example.

The `get_section_name` factory functions are left for backwards compatibility

PR #342, opened 2022-05-18
Add tests for multi_ptr local_accessor constructor

kopylovanat:
Tests for case [3.2.4. local_space constructor](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/multi_ptr_test_plan.asciidoc#324-local_space-constructor).

Used section name builder from https://github.com/KhronosGroup/SYCL-CTS/pull/337


tests/multi_ptr/multi_ptr_local_accessor_constructor.h
```diff
@@ -0,0 +1,76 @@
...
+#include "multi_ptr_common.h"
+
+namespace multi_ptr_local_accessor_constructors {
+
+constexpr int expected_val = 42;
+
+template <typename T, typename Address_SpaceT, typename DimensionT>
+class run_local_accessor_cnstr_tests {
+  static constexpr sycl::access::address_space Space = Address_SpaceT::value;
+  static constexpr int dims = DimensionT::value;
```
vasilytric, line 22: `Space` starts from the capital letter, but `dims` starts from the lowercase letter.  should they follow the one variable naming style?

tests/multi_ptr/multi_ptr_local_accessor_constructor.h
```diff
@@ -0,0 +1,76 @@
...
+#ifndef __SYCLCTS_TESTS_MULTI_PTR_LOCAL_ACCESSOR_CONSTRUCTORS_H
+#define __SYCLCTS_TESTS_MULTI_PTR_LOCAL_ACCESSOR_CONSTRUCTORS_H
+
+#include "../common/common.h"
+#include "../common/section_name_builder.h"
+#include "multi_ptr_common.h"
+
+namespace multi_ptr_local_accessor_constructors {
+
+constexpr int expected_val = 42;
```
vasilytric, line 17: This variable used `expected_val` just once, can we avoid using this variable?

kopylovanat: Removed and replaced its usage with integer literal.

tests/multi_ptr/multi_ptr_local_accessor_constructor.h
kopylovanat: Changed to lowercase.

tests/multi_ptr/multi_ptr_local_accessor_constructor.h
andreyromanof: I think it's not a good idea. We don't get a lot of benefits after deleting the variable, but we get a "magic number" in our code. It's always hard to understand what this number does, especially after some time.

tests/multi_ptr/multi_ptr_common.h
```diff
@@ -11,6 +11,29 @@
...
+        return "access::address_space::global_space";
+      case type::local_space:
+        return "access::address_space::local_space";
+      case type::private_space:
+        return "access::address_space::private_space";
+      case type::generic_space:
+        return "access::address_space::generic_space";
+      default:
+        // no stringification for deprecated ones
+        return "unknown address_space";
```
andreyromanof, line 32:
```suggestion
        return "unknown address_space or adress_space is deprecated";
```
Possibly can help us in the future when some of the adress_space modes will fail to stringify.

tests/multi_ptr/multi_ptr_local_accessor_constructor.h
```diff
@@ -0,0 +1,74 @@
...
+#ifndef __SYCLCTS_TESTS_MULTI_PTR_LOCAL_ACCESSOR_CONSTRUCTORS_H
+#define __SYCLCTS_TESTS_MULTI_PTR_LOCAL_ACCESSOR_CONSTRUCTORS_H
+
+#include "../common/common.h"
+#include "../common/section_name_builder.h"
+#include "multi_ptr_common.h"
+
+namespace multi_ptr_local_accessor_constructors {
+
+template <typename T, typename Address_SpaceT, typename DimensionT>
```
andreyromanof, line 17:
```suggestion
template <typename T, typename AddrSpaceT, typename DimensionT>
```
Just a suggestion. Feel free to ignore)

tests/multi_ptr/multi_ptr_local_accessor_constructor.h
kopylovanat: I think you are right, returned the variable for better readability.

tests/multi_ptr/multi_ptr_local_accessor_constructor.h
kopylovanat: Done

tests/multi_ptr/multi_ptr_common.h
kopylovanat: Good idea, changed to "unknown or deprecated address_space".

bader (approved):
LGTM, but I would consider writing a test checking local_accessor properties like access to the memory shared among work-items. For example, run at least two work-items and read the value written by other work-item.

vasilytric:
> but I would consider writing a test checking local_accessor properties like access to the memory shared among work-items. For example, run at least two work-items and read the value written by other work-item.

@bader  Let's make it into another PR

keryell (approved):
Thanks.

PR #343, opened 2022-05-25
Streamline handling of CTS options

psalz:
This adds a new CMake function `add_cts_option` that replaces all manual handling of `SYCL_CTS_ENABLE_*` options.

For each option, a corresponding preprocessor definition is automatically created. This macro is set regardless of the option's
value, which means that `#if <OPTION>` should be used instead of e.g. `#ifdef <OPTION>` going forward. All relevant locations in existing tests have been updated.

All remaining options that used different names for CMake option and preprocessor macro have been unified to use a single name.

An overview of all CTS option values is now printed at the end of CMake configuration:

```
====================================
  SYCL 2020 Conformance Test Suite
       Configuration summary
====================================

 * SYCL_CTS_ENABLE_FULL_CONFORMANCE
   Enable full conformance with extensive tests: OFF

 * SYCL_CTS_ENABLE_EXT_ONEAPI_PROPERTIES_TESTS
   Enable extension oneAPI compile-time property list tests: OFF

 * SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS
   Enable tests for deprecated SYCL features: ON

 * SYCL_CTS_ENABLE_EXT_ONEAPI_SUB_GROUP_MASK_TESTS
   Enable extension oneAPI sub_group_mask tests: OFF

 * SYCL_CTS_ENABLE_EXT_ONEAPI_DEVICE_GLOBAL_TESTS
   Enable extension oneAPI device_global tests: OFF

 * SYCL_CTS_ENABLE_VERBOSE_LOG
   Enable debug-level logs (deprecated): OFF

 * SYCL_CTS_ENABLE_DOUBLE_TESTS
   Enable tests that require double precision floating point capabilities: ON

 * SYCL_CTS_ENABLE_HALF_TESTS
   Enable tests that require half precision floating point capabilities: ON

 * SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
   Enable OpenCL interoperability tests: ON

CMake Warning at cmake/AddCTSOption.cmake:49 (message):
  Full conformance mode (SYCL_CTS_ENABLE_FULL_CONFORMANCE) should be used for
  conformance submission
Call Stack (most recent call first):
  CMakeLists.txt:92 (print_cts_config_summary)
```

Additionally, this removes the `add_host_and_device_compiler_definitions` function as it modified global state (through `compile_definitions` and the `computecpp_device_compiler_defs` cache variable), which makes it susceptible to containing stale/duplicate definitions upon reconfiguration.

Device compiler options for ComputeCpp (which is the only SYCL implementation with a separate device compiler pass at this time) are now instead inferred from the test executable through generator expressions. This is a prerequisite for fixing the issues I've encountered in #331.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,10 +6,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON) # Required for hex floats in C++11 mode on gcc 6+
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
 
-list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
-include(AddSYCLExecutable)
-include(AddCTSOption)
-
 find_package(Threads REQUIRED)
 find_package(OpenCL REQUIRED)
 find_package(PythonInterp 3 REQUIRED)
@@ -28,6 +24,10 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
    ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     add_compile_options(-Wall -Wno-unused-variable)
 endif()
+
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+include(AddSYCLExecutable)
+include(AddCTSOption)
 
 add_cts_option(SYCL_CTS_ENABLE_FULL_CONFORMANCE
     "Enable full conformance with extensive tests" OFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ set(CMAKE_CXX_EXTENSIONS ON) # Required for hex floats in C++11 mode on gcc 6+
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+include(AddSYCLExecutable)
+include(AddCTSOption)
 
 find_package(Threads REQUIRED)
 find_package(OpenCL REQUIRED)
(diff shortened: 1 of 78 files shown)


CMakeLists.txt
```diff
@@ -27,92 +25,40 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
...
-    add_host_and_device_compiler_definitions(-DSYCL_CTS_ENABLE_FULL_CONFORMANCE)
-else()
-    message(STATUS "Full conformance mode: OFF")
-    message(WARNING
-            "Full conformance mode should be used for conformance submission")
-endif()
-# ------------------
+add_cts_option(SYCL_CTS_ENABLE_FULL_CONFORMANCE
+    "Enable full conformance with extensive tests" OFF
+    WARN_IF_OFF "Full conformance mode (SYCL_CTS_ENABLE_FULL_CONFORMANCE) should be used for conformance submission")
```
bader, line 34: In my day-to-day work, I don't run CTS in full mode, so I'll always see a warning. GitHub actions job logs also will have such warning in each log, which doesn't seem useful.

I think it would be nice a have an option to suppress this such warnings or just do not emit it.

bader (approved):

CMakeLists.txt
psalz: I personally don't really mind the warning, and I think with this change the situation is already improved as it is printed at the very end, below the configuration summary, making it easier to ignore as it doesn't just whiz by. However I do also question the usefulness of the warning, since to generate the report one would use the script (which sets the flag automatically) anyway. I have some improvements regarding conformance report generation planned, I suggest we revisit this then.

psalz:
@ProGTX if you have time, please take a look at my changes regarding device compiler option handling for ComputeCpp, since you were the last to touch this.

ProGTX (approved):

fraggamuffin:
LGTM

bader:
@psalz, sorry, I introduced merge conflicts with recent merges. Could you resolve them, please?

psalz:
Rebased onto current SYCL-2020 branch and converted new uses of `#ifd[n]def` to `#if`.

keryell (approved):
Nice!

PR #355, opened 2022-06-16
Require size member function for assign and are_equal

steffenlarsen:
Overloads of `value_operations::assign` and `value_operations::are_equal` for types with subscript operator uses the `size` member function for iterating the members. However it only checks for the subscript operator. This can cause compile-time errors when using these with types such as `user_struct` that define the subscript operator and a comparison operator but not the `size` member function.

This commit further restricts the aforementioned overloads by requiring the `size` member function. This is done through adding a new type trait for checking a type for the `size` member function, namely `has_size`, and a shortcut trait `has_subscript_and_size` for the conjunction of `has_subscript_operator` and `has_size`.

### DIFF
--- a/util/type_traits.h
+++ b/util/type_traits.h
@@ -101,6 +101,42 @@ template <typename T>
 inline constexpr bool has_subscript_operator_v =
     has_subscript_operator_t<T>::value;
 
+/**
+ * @brief Verify \c T has size member function
+ */
+template <typename T, typename = void>
+struct has_size : std::false_type {};
+
+template <typename T>
+struct has_size<
+    T, std::void_t<decltype(std::declval<T>().size())>>
+    : std::true_type {};
+
+/**
+ * @brief Shortcut for has_size::type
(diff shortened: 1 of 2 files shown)


bader (approved):

util/type_traits.h
```diff
@@ -101,6 +101,42 @@ template <typename T>
...
+ */
+template <typename T> struct has_subscript_and_size {
+    static constexpr bool value = has_subscript_operator_v<T> && has_size_v<T>;
+};
+
+/**
+ * @brief Shortcut for has_subscript_and_size::value
+ */
+template <typename T>
+constexpr bool has_subscript_and_size_v = has_subscript_and_size<T>::value;
```
ProGTX, line 138:
```suggestion
constexpr inline bool has_subscript_and_size_v = has_subscript_and_size<T>::value;
```

ProGTX (approved):

PR #354, opened 2022-06-15
Named requirement verification for accessors iterators

andreyromanof:
According to [spec](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_container_interface), all accessor iterator have to satisfy named requirement [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator). This PR provides a test that verifies conformity to the requirement.

As utility objects were also implemented classes that help to perform verifications.
Currently implemented checks for the following named requirements:
- [LegacyIterator](https://en.cppreference.com/w/cpp/named_req/Iterator)
- [LegacyInputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator)
- [LegacyOutputIterator](https://en.cppreference.com/w/cpp/named_req/OutputIterator)
- [LegacyBidirectionalIterator](https://en.cppreference.com/w/cpp/named_req/BidirectionalIterator)
- [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator)
- [EqualityComparable](https://en.cppreference.com/w/cpp/named_req/EqualityComparable)

The new test turned off with `DISABLE_FOR_TEST_CASE`.

In addition, tests for named requirement verifiers can be found in this [file](https://github.com/andreyromanof/SYCL-CTS/blob/e4a6519cbdbcee27b8e85c70a0c0b6009c5c060c/tests/sandbox/Tests.cpp) from [this](https://github.com/andreyromanof/SYCL-CTS/tree/draft/iterator/requirement_check/tests/sandbox) branch. These tests were used in the process of development.

PR #357, opened 2022-06-17
Extend value_operations

andreyromanof:
This PR adds new overloads to value_operations that allows to work with:
- [`std::pair`](https://en.cppreference.com/w/cpp/utility/pair)
- [`std::tuple`](https://en.cppreference.com/w/cpp/utility/tuple)
- [`std::variant`](https://en.cppreference.com/w/cpp/utility/variant)


vasilytric (approved):

vasilytric:
All is ok, but would you mind to leave a few comments for new functions that has been added in this PR

ProGTX (approved):

tests/common/value_operations.h
```diff
@@ -15,24 +15,67 @@
...
 namespace value_operations {
 
+namespace detail {
+
 template <typename T, size_t N>
 using ArrayT = T[N];
 
+/**
+ * @brief Function allows to make assignemt operations to elements of the
+ * container using std::get and std::index_sequence
```
bader, line 31:
```suggestion
 * @brief Function allows to make assignment operations to elements of the
 * container using std::get and std::index_sequence
```

vasilytric: This misprint has been fixed in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/354/commits/7276c4ed3f3f0e1bf0543845082d53a62c3d0ebd

PR #364, opened 2022-07-19
Add test on access to the memory shared among work-items

vasilytric:
These tests provides verification that local_accessor can
access to the memory shared among work-items

### DIFF
--- a/tests/accessor/local_accessor_access_among_work_items.h
+++ b/tests/accessor/local_accessor_access_among_work_items.h
@@ -41,7 +41,6 @@ class run_test {
       T values_arr[2] = {1, 2};
       const sycl::range range(2);
       constexpr T new_value = 5;
-      constexpr T new_assign_val = 20;
       bool is_acc_val_equal_to_expected = false;
       {
         sycl::buffer<bool> val_is_equal_to_expected_buffer(
@@ -58,18 +57,16 @@ class run_test {
                 // We going to set value in one work-item, synchronize
                 // work-items by sycl::group_barrier() calling and check that
                 // value have been set correctly from another work item.
-                if (item.get_global_id(0) != 0) {
+                if (item.get_local_id(0) != 0) {
                   acc_elem = new_value;
                 }
 
                 // Wait until the code above will be completed in all work-items
                 sycl::group_barrier(item.get_group());
 
-                if (item.get_global_id(0) == 0 && acc_elem == new_value) {
-                  acc_elem = new_assign_val;
+                if (item.get_local_id(0) == 0 && acc_elem == new_value) {
+                  val_is_equal_to_expected_acc[0] = true;
                 }
-
-                val_is_equal_to_expected_acc[0] = acc_elem == new_assign_val;
               });
         });
       }
--- a/tests/accessor/local_accessor_access_among_work_items.h
+++ b/tests/accessor/local_accessor_access_among_work_items.h
@@ -1,0 +1,91 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for the properties of sycl::local_accessor that can access to
+//  the memory shared among work-items
+//
+*******************************************************************************/
+#ifndef SYCL_CTS_LOCAL_ACCESSOR_ACCESS_AMONG_WORK_ITEMS_H
+#define SYCL_CTS_LOCAL_ACCESSOR_ACCESS_AMONG_WORK_ITEMS_H
+#include "accessor_common.h"
+
+namespace local_accessor_access_among_work_items {
+using namespace sycl_cts;
+using namespace accessor_tests_common;
+
+/**
+ * @brief Provides a functor that provides verification that local_accessor can
+ *        access to the memory shared among work-items
+ * @tparam T Current data type
+ * @tparam DimensionTypeT Current current dimension size
+ */
+template <typename T, typename DimensionTypeT>
+class run_test {
+  static constexpr int Dimension = DimensionT::value;
+
+ public:
+  /**
+   * @brief Functor that provides verification that local_accessor can access to
+   *        the memory shared among work-items
+   * @param type_name Current data type string representation
+   */
+  void operator()(const std::string& type_name) {
+    auto queue = util::get_cts_object::queue();
+
+    auto section_name = get_section_name<Dimension>(
+        type_name,
+        "Verify possibility to access to the memory shared among work-items. "
+        "[local_accessor]");
+    SECTION(section_name) {
+      T values_arr[2] = {1, 2};
+      const sycl::range range(2);
+      constexpr T new_value = 5;
+      constexpr T new_assign_val = 20;
+      bool acc_val_is_equal_to_expected = false;
+      {
+        sycl::buffer<bool> val_is_equal_to_expected_buffer(
+            &acc_val_is_equal_to_expected, sycl::range(1));
+        queue.submit([&](sycl::handler& cgh) {
+          auto val_is_equal_to_expected_acc =
+              val_is_equal_to_expected_buffer
+                  .template get_access<sycl::access_mode::write>(cgh);
+          cgh.parallel_for(
+              sycl::nd_range(range, range), [=](sycl::nd_item item) {
+                sycl::local_accessor<T, Dimension> acc(range, cgh);
+                T& acc_elem = array_acc[0];
+
+                if (item.get_global_id(0) != 0) {
+                  acc_elem = new_value;
+                }
+
+                // Wait until the code above will be completed in all work-items
+                sycl::group_barrier(item.get_group());
+
+                if (item.get_global_id(0) == 0 && acc_elem == new_value) {
+                  acc_elem = new_assign_val;
+                }
+
+                val_is_equal_to_expected_acc[0] = acc_elem == new_assign_val;
+              });
+        });
+      }
+      CHECK(acc_val_is_equal_to_expected);
+    }
+  }
+};
+
+template <typename T>
+class run_local_accessor_access_among_work_items_tests {
+ public:
+  void operator()(const std::string& type_name) {
+    // Type packs instances have to be const, otherwise for_all_combination will
+    // not compile
+    const auto dimensions = get_dimensions();
+
+    for_all_combinations<run_test, T>(dimensions, type_name);
+  }
+};
+}  // namespace local_accessor_access_among_work_items
+
+#endif  // SYCL_CTS_LOCAL_ACCESSOR_ACCESS_AMONG_WORK_ITEMS_H
--- a/tests/accessor/local_accessor_access_among_work_items.h
+++ b/tests/accessor/local_accessor_access_among_work_items.h
@@ -41,7 +41,7 @@ class run_test {
       T values_arr[2] = {1, 2};
       const sycl::range range(2);
       constexpr T valid_new_value = 5;
-      constexpr T invalid_new_value = 5;
+      constexpr T invalid_new_value = 6;
 
       bool is_acc_val_equal_to_expected = false;
       {
--- a/tests/accessor/local_accessor_access_among_work_items.h
+++ b/tests/accessor/local_accessor_access_among_work_items.h
@@ -40,8 +40,8 @@ class run_test {
     SECTION(section_name) {
       T values_arr[2] = {1, 2};
       const sycl::range range(2);
-      constexpr T valid_new_value = 5;
-      constexpr T invalid_new_value = 6;
+      constexpr T valid_value = 5;
+      constexpr T invalid_value = 6;
 
       bool is_acc_val_equal_to_expected = false;
       {
@@ -54,22 +54,23 @@ class run_test {
           sycl::local_accessor<T, Dimension> acc(range, cgh);
           cgh.parallel_for(
               sycl::nd_range(range, range), [=](sycl::nd_item item) {
-                T& acc_elem = acc[0];
-
-                // We going to set value in one work-item, synchronize
-                // work-items by sycl::group_barrier() calling and check that
-                // value have been set correctly from another work item.
-                if (item.get_local_id(0) != 0) {
-                  acc_elem = valid_new_value;
-                } else {
-                  acc_elem = invalid_new_value;
-                }
-
-                // Wait until the code above will be completed in all work-items
+                auto lid = item.get_local_id(0);
+                // Initialize local memory with invalid value
+                acc[lid] = invalid_value;
+                // Wait for work-items to finish initialization
                 sycl::group_barrier(item.get_group());
 
-                if (item.get_local_id(0) == 0 && acc_elem == valid_new_value) {
-                  val_is_equal_to_expected_acc[0] = true;
+                // Work-items with index greater than 0 writes to valid data to
+                // the first element of the local data
+                if (lid != 0) {
+                  acc[0] = valid_value;
+                }
+
+                // Wait for data store to finish
+                sycl::group_barrier(item.get_group());
+                // 0th work-item reports the result
+                if (lid == 0) {
+                  val_is_equal_to_expected_acc[0] = acc[0] == valid_value;
                 }
               });
         });
--- a/tests/accessor/local_accessor_access_among_work_items.h
+++ b/tests/accessor/local_accessor_access_among_work_items.h
@@ -26,7 +26,7 @@ class run_test {
 
  public:
   /**
-   * @brief Functor that provides verification that local_accessor can access to
+   * @brief Functor that provides verification that local_accessor can access
    *        the memory shared among work-items
    * @param type_name Current data type string representation
    */
--- a/tests/accessor/local_accessor_access_among_work_items.h
+++ b/tests/accessor/local_accessor_access_among_work_items.h
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-//  Provides tests that the sycl::local_accessor can access to the memory shared
+//  Provides tests that the sycl::local_accessor can access the memory shared
 //  among work-items.
 //
 *******************************************************************************/
@@ -16,7 +16,7 @@ using namespace accessor_tests_common;
 
 /**
  * @brief Provides a functor that provides verification that local_accessor can
- *        access to the memory shared among work-items
+ *        access the memory shared among work-items
  * @tparam T Current data type
  * @tparam DimensionTypeT Current current dimension size
  */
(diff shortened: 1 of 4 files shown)


tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,91 @@
...
+    auto section_name = get_section_name<Dimension>(
+        type_name,
+        "Verify possibility to access to the memory shared among work-items. "
+        "[local_accessor]");
+    SECTION(section_name) {
+      T values_arr[2] = {1, 2};
+      const sycl::range range(2);
+      constexpr T new_value = 5;
+      constexpr T new_assign_val = 20;
+      bool acc_val_is_equal_to_expected = false;
```
andreyromanof, line 45:
```suggestion
      bool is_acc_val_equal_to_expected = false;
```

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,91 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for the properties of sycl::local_accessor that can access to
```
andreyromanof, line 5: Do we really need word "properties" here?

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,91 @@
...
+      {
+        sycl::buffer<bool> val_is_equal_to_expected_buffer(
+            &acc_val_is_equal_to_expected, sycl::range(1));
+        queue.submit([&](sycl::handler& cgh) {
+          auto val_is_equal_to_expected_acc =
+              val_is_equal_to_expected_buffer
+                  .template get_access<sycl::access_mode::write>(cgh);
+          cgh.parallel_for(
+              sycl::nd_range(range, range), [=](sycl::nd_item item) {
+                sycl::local_accessor<T, Dimension> acc(range, cgh);
```
andreyromanof, line 55: I can't say for sure, but shouldn't we create `local_accessor` instance before `parallel_for` invocation?

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,91 @@
...
+                T& acc_elem = array_acc[0];
+
+                if (item.get_global_id(0) != 0) {
```
andreyromanof, line 58: Looks like you try to set value in one work item, synchronize work items with `group_barrier` and check that value have been set correctly from another work item. If you leave a comment to this algorithm it would be nice.

vasilytric: Comment has been added in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/5db87bf32e0364c53ccc4bcf6341571d4c8c3a19

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: Yes, I've done it in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/5db87bf32e0364c53ccc4bcf6341571d4c8c3a19

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: No, deleted in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/5db87bf32e0364c53ccc4bcf6341571d4c8c3a19

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: I've applied this comment in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/5db87bf32e0364c53ccc4bcf6341571d4c8c3a19

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+   *        the memory shared among work-items
+   * @param type_name Current data type string representation
+   */
+  void operator()(const std::string& type_name) {
+    auto queue = util::get_cts_object::queue();
+
+    auto section_name = get_section_name<Dimension>(
+        type_name,
+        "Verify possibility to access to the memory shared among work-items. "
+        "[local_accessor]");
```
bader, line 39:
```suggestion
        "Verify possibility to access the memory shared among work-items. "
        "[local_accessor]");
```

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+#define SYCL_CTS_LOCAL_ACCESSOR_ACCESS_AMONG_WORK_ITEMS_H
+#include "accessor_common.h"
+
+namespace local_accessor_access_among_work_items {
+using namespace sycl_cts;
+using namespace accessor_tests_common;
+
+/**
+ * @brief Provides a functor that provides verification that local_accessor can
+ *        access to the memory shared among work-items
```
bader, line 19:
```suggestion
 * @brief Provides a functor that provides verification that local_accessor can
 *        access the memory shared among work-items
```

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+ * @tparam T Current data type
+ * @tparam DimensionTypeT Current current dimension size
+ */
+template <typename T, typename DimensionTypeT>
+class run_test {
+  static constexpr int Dimension = DimensionT::value;
+
+ public:
+  /**
+   * @brief Functor that provides verification that local_accessor can access to
```
bader, line 29:
```suggestion
   * @brief Functor that provides verification that local_accessor can access
```

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+      constexpr T new_assign_val = 20;
+      bool is_acc_val_equal_to_expected = false;
+      {
+        sycl::buffer<bool> val_is_equal_to_expected_buffer(
+            &is_acc_val_equal_to_expected, sycl::range(1));
+        queue.submit([&](sycl::handler& cgh) {
+          auto val_is_equal_to_expected_acc =
+              val_is_equal_to_expected_buffer
+                  .template get_access<sycl::access_mode::write>(cgh);
+          sycl::local_accessor<T, Dimension> acc(range, cgh);
```
bader, line 53: I don't see this variable to be used in the kernel.

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+          cgh.parallel_for(
+              sycl::nd_range(range, range), [=](sycl::nd_item item) {
+                T& acc_elem = array_acc[0];
```
bader, line 56: What is array_acc?

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+
+                // We going to set value in one work-item, synchronize
+                // work-items by sycl::group_barrier() calling and check that
+                // value have been set correctly from another work item.
+                if (item.get_global_id(0) != 0) {
```
bader, line 61:
```suggestion
                if (item.get_local_id(0) != 0) {
```

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,94 @@
...
+                }
+
+                // Wait until the code above will be completed in all work-items
+                sycl::group_barrier(item.get_group());
+
+                if (item.get_global_id(0) == 0 && acc_elem == new_value) {
+                  acc_elem = new_assign_val;
+                }
+
+                val_is_equal_to_expected_acc[0] = acc_elem == new_assign_val;
```
bader, line 72: You have a data race here. Depending on work-item execution order you might get both values: `true` and `false`.

bader (changes requested):

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: This comment has been applied [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/bb0394a3416ce372412bd19c44da15e021bf8702#diff-cbe2db4042ddaa9f8e561ec7b67998e37f2ade0860b556328eacd9374694816cL38-R38)

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: This comment has been applied [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/bb0394a3416ce372412bd19c44da15e021bf8702#diff-cbe2db4042ddaa9f8e561ec7b67998e37f2ade0860b556328eacd9374694816cL19-R19)

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: This logic has been fixed [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/cfc7f28aead936ab09d3c940b40c70cb241110f8#diff-cbe2db4042ddaa9f8e561ec7b67998e37f2ade0860b556328eacd9374694816cL68-R68)
Now there are no race conditions

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: This comment has been applied [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/cfc7f28aead936ab09d3c940b40c70cb241110f8#diff-cbe2db4042ddaa9f8e561ec7b67998e37f2ade0860b556328eacd9374694816cL61-R60) and in other place

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: This has been misprint, fixed [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/bb0394a3416ce372412bd19c44da15e021bf8702#diff-cbe2db4042ddaa9f8e561ec7b67998e37f2ade0860b556328eacd9374694816cL56-R56)

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: Misprint in variable namin has been fixed in this commit: [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/bb0394a3416ce372412bd19c44da15e021bf8702#diff-cbe2db4042ddaa9f8e561ec7b67998e37f2ade0860b556328eacd9374694816cL56-R56)

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: This comment has been applied in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/b5dcd83f2da4e0927f9afc9bd62e8fd2f4c95a5f

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,91 @@
...
+          cgh.parallel_for(
+              sycl::nd_range(range, range), [=](sycl::nd_item item) {
+                T& acc_elem = acc[0];
+
+                // We going to set value in one work-item, synchronize
+                // work-items by sycl::group_barrier() calling and check that
+                // value have been set correctly from another work item.
+                if (item.get_local_id(0) != 0) {
+                  acc_elem = new_value;
+                }
```
bader, line 62: Here must be an else-branch which sets "invalid" value. If acc memory is not shared, the part of the memory accessed by work-item 0 is left uninitialized and behavior is undefined.

vasilytric: If I correctly understood you.
`is_acc_val_equal_to_expected` has `false` as default value, if `accessor` memory isn't shared, then `is_acc_val_equal_to_expected` will not been changed in `true`, so test will fail

bader: No, I'm talking about local memory accessed via `acc`.

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -40,7 +40,9 @@ class run_test {
     SECTION(section_name) {
       T values_arr[2] = {1, 2};
       const sycl::range range(2);
-      constexpr T new_value = 5;
+      constexpr T valid_new_value = 5;
+      constexpr T invalid_new_value = 5;
```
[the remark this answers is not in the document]
vasilytric: yes, you  are right. fixed in this commit https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/a63095af2112c787fa224d87c16e0af7f78ed02f

tests/accessor/local_accessor_access_among_work_items.h
vasilytric: I've aplied this comment in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/1b06b8c73ad78493043c4fcb9804392e6739e941

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,95 @@
...
+                } else {
+                  acc_elem = invalid_new_value;
+                }
+
+                // Wait until the code above will be completed in all work-items
+                sycl::group_barrier(item.get_group());
+
+                if (item.get_local_id(0) == 0 && acc_elem == valid_new_value) {
+                  val_is_equal_to_expected_acc[0] = true;
+                }
```
bader, line 73:
```suggestion
                auto lid = item.get_local_id(0);
                // Initialize local memory with invalid value
                acc[lid] = invalid_value;
                // Wait for work-items to finish initialization
                sycl::group_barrier(item.get_group());
                
                // Work-items with index greater than 0 writes to valid data to the first element of the local data
                if (lid != 0)
                  acc[0] = valid_value;

                // Wait for data store to finish
                sycl::group_barrier(item.get_group());

                // 0th work-item reports the result
                if (lid == 0)
                  val_is_equal_to_expected_acc[0] = acc[0] == valid_value;
```
I suggest we something like this. It's more clear and should have no UB. The latest version still has a data race as both work-items write to the same memory location before the barrier.

Note: I dropped "new" from the constants name as I didn't get the meaning of it, so please, update declarations as well.
Note2: the suggested code snipped is not clang-formatted.

vasilytric: I've done this in this commit: https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/ea3caa35615978348308d094a8710e8bdc9bc600
clang-format has been applied and `new_` from variable name has been deleted

bader (approved):
LGTM. Thanks for your patience.

tests/accessor/local_accessor_access_among_work_items.h
```diff
@@ -0,0 +1,96 @@
...
+                auto lid = item.get_local_id(0);
+                // Initialize local memory with invalid value
+                acc[lid] = invalid_value;
+                // Wait for work-items to finish initialization
+                sycl::group_barrier(item.get_group());
+
+                // Work-items with index greater than 0 writes to valid data to
+                // the first element of the local data
+                if (lid != 0) {
+                  acc[0] = valid_value;
```
keryell, line 66: It looks to me like a race condition.
Why not having each work-item writing its own id into `acc[(id + 1)%item.get_local_range(0)]` and 
checking later that `acc[(id + 7)%item.get_local_range(0)] == (id + 6)%item.get_local_range(0)` for example.

tests/accessor/local_accessor_access_among_work_items_fp16.cpp
```diff
@@ -0,0 +1,44 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for local_accessor.
+//
+//  This test provides verifications that local_accessor can access the memory
+//  shared among work-items.
```
keryell, line 8: + "in the case of fp16 accesses when available"

tests/accessor/local_accessor_access_among_work_items_fp64.cpp
```diff
@@ -0,0 +1,44 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provides tests for local_accessor.
+//
+//  This test provides verifications that local_accessor can access the memory
+//  shared among work-items.
```
keryell, line 8: + "in the case of fp64 accesses when available"

keryell (changes requested):

tests/accessor/local_accessor_access_among_work_items.h
bader: Why is this a data race?
There are just 2 work-items, so we are definitely safe. Even with the range size > 2, the result of this kernel must be stable.

keryell: Ah yes, good point I forgot there are only 2 work-items. :-)
If there are more than 2 work-items, where is it written in the memory model description that writing the same values in a data race is not a data race?
At least the test could be something like
```c++
acc[(id + 1)%item.get_local_range(0)] = id;
global_acc[(id + 1)%item.get_local_range(0)] = (id + 2)%item.get_local_range(0);
```
and check on the host that the global buffer has the right iota values to avoid a compliant implementation with any local accessor implemented with a global buffer with just an element.

bader: Sorry, I agree it's a data race, but I don't think it's a problem. The execution order doesn't make the result undefined.

On the other hand I don't mind changing the code to avoid the data race.

keryell: Anyway, all this might be some over-engineering an just the minimal test in this PR can be enough. As you want.

tests/accessor/local_accessor_access_among_work_items_fp16.cpp
vasilytric: I suggest add `For the double type.` instead. Sentences like this uses across all tests.
I've done it [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/a9ad5ace687645d284a860d46e9036888371a92c#diff-c4327d08876350fc635197385b6ecf8f41003273d63765aca038b903eb80a34aR8)

tests/accessor/local_accessor_access_among_work_items_fp64.cpp
vasilytric: I suggest add `For the double type.` instead. Sentences like this uses across all tests.
I've done it [here](https://github.com/KhronosGroup/SYCL-CTS/pull/364/commits/a9ad5ace687645d284a860d46e9036888371a92c#diff-336d0aecd22b6ac2375b4d894a47e5370c5e7553b1646c1a2d488cf7eb64126bR8)

keryell (approved):
Thanks!

ProGTX (approved):

PR #366, opened 2022-07-21
Remove outdated requirements.txt

psalz:
I just spotted this file, which was last updated in 2017. I think keeping this probably does more harm than good.


keryell (approved):
Long journey since. :-)

PR #367, opened 2022-07-26
Substitute deprecated feature kernel::get_work_group_info with kernel::get_info

raaiq1:
SYCL 2020 specification doesn't include kernel::get_work_group_info, thus calls to such methods in the tests are replaced with kernel::get_info


bader (approved):

bader (changes requested):
@raaiq1, please, disable build of failing tests in CI. It seems that ComputeCPP version we use doesn't support new SYCL 2020 API yet.

keryell (approved):

PR #371, opened 2022-08-01
Bump DPC++ version to 7ee07ba

bader:


keryell (approved):
I have no idea about the new taste of `7ee07ba` but you are more a connaisseur than me in this domain. :-)

PR #374, opened 2022-09-02
Fix usages of CHECK_THROWS_MATCHES macro

AlexeySachkov:
Added `()` to actually trigger execution of lambdas passed to the macro
or otherwise the macro didn't invoke anything. Apparently,
`static_cast<void>` on lambda compiles just fine.

### DIFF
--- a/tests/exceptions/exceptions_constructors.cpp
+++ b/tests/exceptions/exceptions_constructors.cpp
@@ -27,7 +27,7 @@ inline void check_exception(sycl::exception e, const std::error_code errcode,
   CHECK(e.category() == errcat);
   CHECK(e.what() != nullptr);
   CHECK(e.has_context() == false);
-  CHECK_THROWS_MATCHES([&e] { e.get_context(); }, sycl::exception,
+  CHECK_THROWS_MATCHES([&e] { e.get_context(); }(), sycl::exception,
                        sycl_cts::util::equals_exception(sycl::errc::invalid));
 }
 
@@ -48,7 +48,7 @@ inline void check_exception(sycl::exception e, const std::error_code errcode,
   CHECK(e.category() == errcat);
   CHECK(e.what() == what_arg);
   CHECK(e.has_context() == false);
-  CHECK_THROWS_MATCHES([&e] { e.get_context(); }, sycl::exception,
+  CHECK_THROWS_MATCHES([&e] { e.get_context(); }(), sycl::exception,
                        sycl_cts::util::equals_exception(sycl::errc::invalid));
 }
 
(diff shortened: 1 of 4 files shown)


AlexeySachkov:
Creating this as a draft for now, because I only tested optional kernel features test and don't know yet if the rest of changes are correct.

[`CHECK_THROWS_MATCHES`](https://github.com/catchorg/Catch2/blob/f25236ff430d98b3cba5495cc68235d1d7701b98/src/catch2/matchers/catch_matchers.hpp#L219) is forwarded to [`INTERNAL_CATCH_THROWS_MATCHES`](https://github.com/catchorg/Catch2/blob/f25236ff430d98b3cba5495cc68235d1d7701b98/src/catch2/matchers/internal/catch_matchers_impl.hpp#L66) where we have the following [line](https://github.com/catchorg/Catch2/blob/f25236ff430d98b3cba5495cc68235d1d7701b98/src/catch2/matchers/internal/catch_matchers_impl.hpp#L71):

```
static_cast<void>(__VA_ARGS__ );
```

`__VA_ARGS__` is what passed as the first argument `expr` to `CHECK_THROWS_MATCHES`.

So, apparently `static_cast<void>(lambda)` [compiles just fine](https://godbolt.org/z/hq41EEK8h), but that's essentially no-op, which causes tests do nothing. This PR fixes that

bader:
Tagging @psalz.

tests/exceptions/exceptions_constructors.cpp
```diff
@@ -27,7 +27,7 @@ inline void check_exception(sycl::exception e, const std::error_code errcode,
   CHECK(e.category() == errcat);
   CHECK(e.what() != nullptr);
   CHECK(e.has_context() == false);
-  CHECK_THROWS_MATCHES([&e] { e.get_context(); }, sycl::exception,
+  CHECK_THROWS_MATCHES([&e] { e.get_context(); }(), sycl::exception,
```
keryell, line 30: But if we are using macros, why using lambdas?
```suggestion
  CHECK_THROWS_MATCHES(e.get_context(), sycl::exception,
```
?

keryell (approved):
Good catch!

bader:
@AlexeySachkov, do you plan to make additional changes?
If it's ready, please, convert it from the draft and we will merge it right away.

tests/exceptions/exceptions_constructors.cpp
AlexeySachkov: Thanks, I haven't thought about it. I've applied the suggestion in 3106ff5980d406e1dd424a01db4d93b6a38f7754

AlexeySachkov:
> @AlexeySachkov, do you plan to make additional changes? If it's ready, please, convert it from the draft and we will merge it right away.

I've applied comment from @keryell, but I don't have plans for any other changes. Note: I still haven't tested locally suites other than optional kernel features, but I suppose that CI should cover for the rest (at least compilation phase)

PR #375, opened 2022-09-05
Switch to struct information descriptors

bader:
Change information descriptors from enum class elements to structs in
accordance with SYCL 2020. Additionally, drop some outdated descriptors
and bump the DPC++ version.

PR #376, opened 2022-09-05
Remove use of is_host and fix default device constructor check

steffenlarsen:
SYCL 2020 removes the is_host member function from SYCL classes in tandem with removing the requirement that the host device must be available. This commit removes the uses of the is_host member function and fixes the expected behavior of the default device constructor to expect the resulting device to be the same as one constructed with default_selector_v.

### DIFF
--- a/tests/context/context_constructors.cpp
+++ b/tests/context/context_constructors.cpp
@@ -35,6 +35,35 @@ class TEST_NAME : public util::test_base {
     set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
   }
 
+  void check_context_after_ctor(sycl::context &context,
+                                sycl::device &expectedDevice,
+                                util::logger &log) {
+    if (context.get_devices().size() != 1) {
+      FAIL(log, "context was not constructed correctly (get_devices size)");
+    }
+
+    if (context.get_devices()[0] != expectedDevice) {
+      FAIL(log, "context was not constructed correctly (device equality)");
+    }
+  }
+
+  void check_context_after_ctor(sycl::context &context,
+                                std::vector<sycl::device> &expectedDevices,
+                                util::logger &log) {
+    if (context.get_devices().size() != expectedDevices.size()) {
+      FAIL(log, "context was not constructed correctly (get_devices size)");
+    }
+
+    for (auto &device : context.get_devices()) {
+      if (std::find(expectedDevices.begin(), expectedDevices.end(), device) ==
+          expectedDevices.end()) {
+        FAIL(log,
+             "context was not constructed correctly (device not in passed "
+             "device list)");
+      }
+    }
+  }
+
   /** execute the test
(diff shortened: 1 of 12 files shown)


steffenlarsen:
Changes need DPC++ bump, which https://github.com/KhronosGroup/SYCL-CTS/pull/375 should bring.

tests/context/context_constructors.cpp
```diff
@@ -86,8 +94,23 @@ class TEST_NAME : public util::test_base {
...
+
+        for (auto &device : context.get_devices()) {
+          if (std::find(deviceList.begin(), deviceList.end(), device) ==
+              deviceList.end()) {
+            FAIL(log,
+                 "context was not constructed correctly (device not in passed "
+                 "device list)");
+          }
+
+          if (std::count(context.get_devices().begin(),
```
keryell, line 109: You could just count once and gives 2 different error messages according to the answer.

tests/context/context_constructors.cpp
```diff
@@ -100,8 +123,23 @@ class TEST_NAME : public util::test_base {
...
```
keryell, line 138: Idem.

tests/context/context_constructors.cpp
```diff
@@ -112,8 +150,24 @@ class TEST_NAME : public util::test_base {
...
+            FAIL(log,
+                 "context was not constructed correctly (device not in "
+                 "platform)");
+          }
+
+          if (std::count(context.get_devices().begin(),
+                         context.get_devices().end(), device) != 1) {
+            FAIL(log,
+                 "context was not constructed correctly (duplicate devices)");
+          }
```
keryell, line 170: What about using a lambda or function instead of all these copy-paste?

keryell (commented):
Great!
While you are here, perhaps you could factorize some code?

tests/context/context_constructors.cpp
keryell: With just this call you can also replace the previous `std::find`. If it is `>= 2` there are too many devices, if `== 0` the device is not found. Only 1 O(n) algorithm instead of 2.

steffenlarsen: I see! Technically it would change the check as one checks for the device in the device-list from the platform and the other counts it in the list of devices reported by the context. However, the duplicate check is somewhat redundant as it will be caught by the other two checks, unless the platform has duplicate devices too which would arguably be platform issue rather than a context issue.

bader:
@steffenlarsen, the changes from #375 has been merged. Could you resolve conflicts, please? Unfortunately GitHub doesn't allow me to do it via WebUI.

keryell (approved):
Thanks.

steffenlarsen:
It appears we may need another bump to the DPC++ version to get support for the new device selectors.

PR #373, opened 2022-09-02
Switch to struct information descriptors

sergey-semenov:
Change information descriptors from enum class elements to structs in
accordance with SYCL 2020. Additionally, drop some outdated descriptors
and bump the DPC++ version.

PR #369, opened 2022-08-01
Update copyright notices to Khronos copyright

npmiller:
This patch updates all the copyright notices to reflect that the CTS is
no longer under Codeplay copyright but under Khronos copyright, and adds
the relevant Khronos Copyright notice.

PR #377, opened 2022-09-08
Flesh out developer documentation on CI and category filters

psalz:
This includes a section on how to update the version of a SYCL implementation, so we hopefully avoid running into the same issue over and over again (last seen in #373 / #375).

I've also updated the section regarding the copyright header, as the header used in most files was recently changed in #369. @npmiller I assume that new files (which this section talks about) should not include the copyright for Codeplay - correct?

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -131,6 +131,18 @@ Each file in the SYCL CTS should be prefaced by the Khronos copyright header:
 //
 //  Copyright (c) <YEAR> The Khronos Group Inc.
 //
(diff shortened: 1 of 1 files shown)


bader (approved):

bader:
> Yes, I believe that's correct cc @AerialMantis

@AerialMantis, could you confirm that copyright header is correct, please?

AerialMantis (approved):

bader:
Thanks!

PR #379, opened 2022-09-08
CI: Summarize matrix build job results in final job

psalz:
This adds a new job to our CI pipeline that simply summarizes the results of the previous matrix job. This is useful because we can then use this job as a stand-in for the build step in GitHub's branch protection rules.

Currently, all entries in the build matrix need to be enumerated manually, and whenever the version of a SYCL implementation is updated, the rules need to be updated as well:

image

(Here, DPC++ was updated from caa696f to b3cbda5, so the former never runs and GitHub prevents the PR from being merged).

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -149,3 +149,19 @@ jobs:
         with:
           name: build-times-${{ matrix.sycl-impl }}
           path: ${{ env.container-workspace }}/build/build_times.log
+
+  # This job simply summarizes the results of the "compile-cts" matrix build job above.
+  # It can then be used in a branch protection rule instead of having to enumerate all
+  # entries of the matrix manually (and having to update it whenever the matrix changes).
+  cts-compiles-for-all-implementations:
+    needs: [compile-cts]
+    if: always()
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Summarize matrix build results
+        run: |
+          if [[ "${{ needs.compile-cts.result }}" == "success" ]]; then
+            exit 0
+          else
+            exit 1
+          fi


.github/workflows/cts_ci.yml
```diff
@@ -149,3 +149,19 @@ jobs:
         with:
           name: build-times-${{ matrix.sycl-impl }}
           path: ${{ env.container-workspace }}/build/build_times.log
+
```
bader, line 152: `compile-cts` job has following code:
```
    needs: build-image-for-sycl-impl
    # Wait for Docker image builds, but run even if they're skipped
    if: always()
```

I think we might need to check `build-image-for-sycl-impl` job status as well, as `compile-cts` is executed even when `build-image-for-sycl-impl` fails.

bader (commented):
The idea sounds good to me, but I think we must be sure that all jobs are **finished** successfully (or skipped).

.github/workflows/cts_ci.yml
psalz: What you mean is that a CI run should be considered failed when `build-image-for-sycl-impl` fails? In other words, `compile-cts` should run when `build-image-for-sycl-impl` is skipped, but not when it fails?

bader: I think so. According to my understanding `build-image-for-sycl-impl` is triggered when we update SYCL implementation version and it might fail (e.g. it builds some implementations from sources and we can fail at this step). I'm not sure if we get valid results from `compile-cts` in such case.

psalz: Now that I think about it, that can't happen. If the image build fails, the `compile-cts` will still run, but will fail once it tries to pull the new image.

bader: I see. Even though there is no job status check in GitHub Actions script, there is a dependency via dockerhub, when both jobs use the same SHA.

bader (approved):

.github/workflows/cts_ci.yml
psalz: Exactly. We could probably make this clearer in the workflow definition though, I'll take a note.

psalz:
Merged and updated the branch protection rule. All existing PRs will not pass this check unless back-merged / rebased, so we will just have to force-merge them.

PR #383, opened 2022-09-26
Fix strict aliasing violation in check_as_result

dwoodwor-intel:
This function writes values into a temporary array using one type and then reads them back from the same array using a different type to compare against the expected result. This is undefined behavior in C and C++, so this has been causing a hang when running these tests. The fix is to do this properly by using separate arrays for the differently-typed values and a memcpy to copy the bits from the input type to the output type.

### DIFF
--- a/tests/common/common_vec.h
+++ b/tests/common/common_vec.h
@@ -441,7 +441,8 @@ asType check_as_result(sycl::vec<vecType, N> inputVec,
   for (size_t i = 0; i < N; ++i) {
     tmp_ptr[i] = getElement(inputVec, i);
   }
-  asType* exp_ptr = reinterpret_cast<asType*>(tmp_ptr);
+  asType exp_ptr[asN];
+  memcpy(exp_ptr, tmp_ptr, sizeof(exp_ptr));
   for (size_t i = 0; i < asN; ++i) {
     if (exp_ptr[i] != getElement(asVec, i)) {
       return false;


tests/common/common_vec.h
```diff
@@ -441,7 +441,8 @@ asType check_as_result(sycl::vec<vecType, N> inputVec,
   for (size_t i = 0; i < N; ++i) {
     tmp_ptr[i] = getElement(inputVec, i);
   }
-  asType* exp_ptr = reinterpret_cast<asType*>(tmp_ptr);
+  asType exp_ptr[asN];
+  memcpy(exp_ptr, tmp_ptr, sizeof(exp_ptr));
```
bader, line 445: Can we use C++ std::variant here instead?

dwoodwor-intel: Does std::variant allow this kind of type punning? I thought it was designed to make it impossible to access the value using the wrong type like this.

bader: You are right. Sorry for a senseless question.

bader (approved):

tests/common/common_vec.h
keryell, line 445:
```suggestion
  std::memcpy(exp_ptr, tmp_ptr, sizeof(exp_ptr));
```
Do we include the right header too?
Just curious: do we define the layout of a `sycl::vec`?
Does this code works with little and big endian layout?

ProGTX: Maybe we could use `sycl::bit_cast` instead? `std::bit_cast` is intended to replace `reinterpret_cast` in cases like this, and it was pre-adopted in SYCL 2020.

ProGTX (approved):

fraggamuffin:
Will check endianness in spec. But this is ok.

keryell (approved):
Let's approve this for now and think further about the `std::vec` layout problem.

tests/common/common_vec.h
bader: @dwoodwor-intel, although the pull request is merged, I think Peter's suggestion is worth considering.

ProGTX:
I agree, it wouldn't solve the endianess problem, my suggestion was just to avoid doing a manual `memcpy`, but that's quite a minor thing. I also agree that the layout will probably need more discussion.

tests/common/common_vec.h
dwoodwor-intel: I've been looking into this, but there are a few issues that I've run into:
* `std::bit_cast` (and `sycl::bit_cast`) don't work with plain C arrays like these because those can't be returned from functions, and my understanding is that they aren't safe to use with `std::array` either because the standard doesn't have any layout requirements for that. It seems like the way around this is to define a wrapper struct around an array of the right type/size.
* `std::bit_cast` (and `sycl::bit_cast`) require that the input and output types are the same size, which is not always the case when this function is instantiated. It's possible to match the input/output array sizes similar to just copying up to the size of the output array with memcpy, but this is a little ugly because there might be partial elements of the input array read. My understanding based on `check_vectorN_as` is that these should really be the same size in any cases actually tested, so it seems like this can be worked around with a constexpr if too.

Overall, I think this function with `sycl::bit_cast` would look like this:

```c++
template <typename vecType, int N, typename asType, int asN>
asType check_as_result(sycl::vec<vecType, N> inputVec,
                       sycl::vec<asType, asN> asVec) {
  if constexpr (sizeof(asType[asN]) == sizeof(vecType[N])) {
    vecType tmp_ptr[N];
    for (size_t i = 0; i < N; ++i) {
      tmp_ptr[i] = getElement(inputVec, i);
    }
    struct WrappedBitCastArray {
      asType vals[asN];
    };
    auto exp_ptr = sycl::bit_cast<WrappedBitCastArray>(tmp_ptr);
    for (size_t i = 0; i < asN; ++i) {
      if (exp_ptr.vals[i] != getElement(asVec, i)) {
        return false;
      }
    }
  } else {
    return false;
  }
  return true;
}
```

Is this preferable to just using memcpy? If so, I can submit this as a follow-up change.

ProGTX: My aim was to simplify code, but it seems like using it would lead in the opposite direction, so I prefer the `memcpy` here in that case. Thanks for checking this, though, it's very informative.

PR #386, opened 2022-10-17
Change event::wait_and_throw test case

steffenlarsen:
One of the event::wait_and_throw test cases makes the assumption that queue async handlers are somehow propagated to contexts. However, according to [Section 4.13.1.3](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_priorities_of_async_handlers) there are priorities to which async handlers are to be used, where queue is the highest, assuming the runtime can make the association between events and queues. This commit changes the test to follow the aforementioned section, assuming the queues are associated with the events. See also https://github.com/KhronosGroup/SYCL-Docs/issues/299.

### DIFF
--- a/tests/event/event.cpp
+++ b/tests/event/event.cpp
@@ -348,11 +348,14 @@ TEST_CASE(
 }
 
 // TODO SPEC: It is unclear what "any unconsumed error [...] will be passed to
-// the async handler associated with the context" means when two queues override
-// their (shared) context's async handler.
+// the async handler associated with the context" means when the context does
+// not have an async handler, but the queues do. This test assumes that the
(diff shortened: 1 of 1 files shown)


psalz (approved):
Yes, makes sense to me!

keryell (approved):
Sounds good.
Looking at the test implementations, the coding style around the `std::move` looks a little bit weird and it lacks of comments.
But it does not prevent them to work for now.

psalz:
> Looking at the test implementations, the coding style around the `std::move` looks a little bit weird and it lacks of comments. But it does not prevent them to work for now.

You are right, it does look weird. Outright wrong, even. Fixed in #392.

PR #393, opened 2022-11-03
Remove library dependency on OpenCL

### LINKED ISSUE #347: Does `OpenCL` is really required?
Hi,

https://github.com/KhronosGroup/SYCL-CTS/blob/94a4b36923a730a1999f899bdb8bef1804ffee05/CMakeLists.txt#L12

Does OpenCL is really required?

I try to build the CTS on a system where OpenCL is not available.

psalz:
Include OpenCL headers via submodule (we still require these for reference math implementation), unless OpenCL interop testing is enabled.

Fixes #347.


cmake/FindDPCPP.cmake
```diff
@@ -55,7 +55,7 @@ endif()
 
 add_library(DPCPP::Runtime INTERFACE IMPORTED GLOBAL)
 set_target_properties(DPCPP::Runtime PROPERTIES
-  INTERFACE_LINK_LIBRARIES    OpenCL::OpenCL
+#   INTERFACE_LINK_LIBRARIES    OpenCL::OpenCL
```
bader, line 58:
```suggestion
```

keryell (approved):
Thanks.

bader (changes requested):

CMakeLists.txt
```diff
@@ -62,6 +65,17 @@ add_cts_option(SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
 
 add_cts_option(SYCL_CTS_ENABLE_CUDA_INTEROP_TESTS
     "Enable CUDA interoperability tests", OFF)
+
+# ------------------
+if(SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS OR SYCL_IMPLEMENTATION STREQUAL "ComputeCpp")
```
bader, line 70: Could you add a comment here with the clarification why ComputeCPP case is special?

bader (approved):

psalz:
Okay, I went back to the drawing board and chose a different route. I'm now defining a proxy library (`CTS::OpenCL_Proxy`), which allows me to transitively set `CL_TARGET_OPENCL_VERSION=120` on all targets that include the OpenCL header. This finally suppresses the warning `"CL_TARGET_OPENCL_VERSION is not defined. Defaulting to 300 (OpenCL 3.0)"` which we've been emitting for ages.

bader (approved):
Looks good to me.
Nitpick: "Remove library dependency on OpenCL" sounds too ambitious after recent two re-bases. :)

keryell (approved):
Nice!

tomdeakin:
@ProGTX please review, merge is accepted.

ProGTX (approved):

PR #394, opened 2022-11-04
Split fp64 case from usm_atomic_access and usm_atomic_access_atomic64

steffenlarsen:
The use of double in usm_atomic_access and usm_atomic_access_atomic64 is not guarded by a check for fp64. This commit splits these checks for double into a new test usm_atomic_access_atomic64_fp64. Note that we do not need usm_atomic_access_fp64 as testing for double atomics require both fp64 and atomic64.

### DIFF
--- a/tests/usm/usm_atomic_access.h
+++ b/tests/usm/usm_atomic_access.h
@@ -20,7 +20,7 @@
 
 namespace usm_atomic_access {
 
-static auto get_scalar_types() {
+inline auto get_nondouble_scalar_types() {
   static const auto scalar_types =
       named_type_pack<int, unsigned int, long,
                       unsigned long, float, double, long long,
@@ -28,6 +28,11 @@ static auto get_scalar_types() {
                                            "unsigned long", "float", "double",
                                            "long long", "unsigned long long");
   return scalar_types;
+}
+
+inline auto get_fp64_type() {
+  static const auto types = named_type_pack<double>::generate("double");
+  return types;
 }
 
 /** @brief Return atomic aspect depending on the type of allocated memory
--- a/tests/usm/usm_atomic_access_atomic64_fp64.cpp
+++ b/tests/usm/usm_atomic_access_atomic64_fp64.cpp
@@ -3,13 +3,13 @@
 //  SYCL 2020 Conformance Test Suite
 //
 //  Provide verification to atomic access for USM allocations that underlying
-//  type size lower than 64 byte.
+//  type size equal than 64 byte.
 //
 *******************************************************************************/
 
 #include "usm_atomic_access.h"
 
-#define TEST_NAME usm_atomic_access_core
+#define TEST_NAME usm_atomic_access_atomic64_fp64
 
 namespace TEST_NAMESPACE {
 using namespace sycl_cts;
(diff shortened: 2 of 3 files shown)


tests/usm/usm_atomic_access.h
```diff
@@ -20,7 +20,7 @@
 
 namespace usm_atomic_access {
 
-static auto get_scalar_types() {
+inline auto get_nondouble_scalar_types() {
   static const auto scalar_types =
       named_type_pack<int, unsigned int, long,
                       unsigned long, float, double, long long,
```
bader, line 26: Shouldn't we remove "double" from the line 28?

tests/usm/usm_atomic_access_atomic64_fp64.cpp
```diff
@@ -0,0 +1,47 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provide verification to atomic access for USM allocations that underlying
+//  type size equal than 64 byte.
```
bader, line 6:
```suggestion
//  type size equal to 64 byte.
```
Is this a typo?
If so, please, fix in tests/usm/usm_atomic_access_atomic64.cpp as well.

tests/usm/usm_atomic_access_atomic64_fp64.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+  /** execute the test
+   */
+  void run(util::logger &log) override {
+    {
+      auto queue{util::get_cts_object::queue()};
+      if (!queue.get_device().has(sycl::aspect::fp64)) {
+        log.note(
+            "Device does not support double precision floating point "
+            "operations");
```
bader, line 35:
```suggestion
        WARN(
          "Device does not support double precision floating point operations. "
          "Skipping the test case.");
```

tests/usm/usm_atomic_access_atomic64_fp64.cpp
```diff
@@ -0,0 +1,47 @@
...
+ public:
+  /** return information about this test
+   */
+  void get_info(test_base::info &out) const override {
+    set_test_info(out, TOSTRING(TEST_NAME), TEST_FILE);
+  }
+
+  /** execute the test
+   */
+  void run(util::logger &log) override {
```
bader, line 29: Please, use Catch2 testing framework instead.

bader (changes requested):

tests/usm/usm_atomic_access_atomic64_fp64.cpp
steffenlarsen: The comments have been reworded for both tests.

tests/usm/usm_atomic_access_atomic64_fp64.cpp
steffenlarsen: Do you have an example of a test-case that uses Catch2 for reference?

bader: https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/event/event.cpp
https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/docs/README.adoc

bader (approved):

tests/usm/usm_atomic_access_atomic64_fp64.cpp
bader: Great. Thanks a lot!

tests/usm/usm_atomic_access.h
```diff
@@ -20,16 +20,21 @@
 
 namespace usm_atomic_access {
 
-static auto get_scalar_types() {
+inline auto get_nondouble_scalar_types() {
```
keryell, line 23: Just curious about the role of inline here.

keryell (approved):

ProGTX (approved):

tests/usm/usm_atomic_access.h
steffenlarsen: `static` will cause warnings now that the different source files don't necessarily use the functions. Granted we may not need `inline` here, we will at least avoid duplicate definition errors if it ever becomes relevant.

keryell: I see. Thanks.

PR #398, opened 2022-11-16
Fix SYCL 2020 accessor::value_type tests

bader:
The accessor class's value_type member was updated recently to change its "const"-ness when in read-only mode (https://github.com/intel/llvm/pull/7096) for SYCL 2020 compliance as defined in [Section 4.7.6.9.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_command_accessors) The tests were not updated to reflect this change.

This patch specifically addresses the `accessor_api_buffer_core` test in the accessor_legacy directory and the error:

```
accessor_api_buffer_core - static assertion failed due to requirement 'std::is_same<const sycl::vec<int, 16>, sycl::vec<int, 16>>::value': value_type is of wrong type
```

In addition to the changes to accessor_api_common_all.h, it was necessary to resolve issues with accessor_api_image_common.h as there were tests in there for invalid targets. host_image, image, and image_array have been removed from the SYCL 2020 specification in favour of respectively specialised accessors. It was decided to simply remove these from the tests for that reason due to this being legacy testing.

This PR is cloned from https://github.com/KhronosGroup/SYCL-CTS/pull/395 to update DPC++ compiler.


tests/accessor_legacy/accessor_api_common_all.h
```diff
@@ -50,8 +52,12 @@ void check_accessor_members(sycl_cts::util::logger &log,
 
   using acc_t = sycl::accessor<T, dims, mode, target, placeholder>;
 
+  using t_type = typename std::conditional<mode == sycl::access_mode::read,
+                                           const T, T>::type;
```
keryell, line 56:
```suggestion
  using acc_t = sycl::accessor<T, dims, mode, target, placeholder>;

  using t_type = typename std::conditional_t<mode == sycl::access_mode::read,
                                           const T, T>;
```

tests/accessor_legacy/accessor_api_common_all.h
```diff
@@ -50,8 +52,12 @@ void check_accessor_members(sycl_cts::util::logger &log,
...
+
   using value_type = typename acc_t::value_type;
-  static_assert(std::is_same<value_type, T>::value,
+
+  static_assert(std::is_same<value_type, t_type>::value,
```
keryell, line 60:
```suggestion
  static_assert(std::is_same_v<value_type, t_type>,
```

keryell (changes requested):
Possible modernizing.

tests/accessor_legacy/accessor_api_common_all.h
ProGTX: You can get rid of `typename` in that case as well.

tests/accessor_legacy/accessor_api_image_common.h
```diff
@@ -1831,9 +1640,10 @@ class check_image_accessor_api_type {
     /**
      *  check image accessor api for all dimensions
      */
-    check_image_accessor_api_dim<T, 1>(log, typeName, count, size, queue);
-    check_image_accessor_api_dim<T, 2>(log, typeName, count, size, queue);
-    check_image_accessor_api_dim<T, 3>(log, typeName, count, size, queue);
+    // check_image_accessor_api_dim<T, 1>(log, typeName, count, size, queue);
```
ProGTX, line 1643: Why is this code commented out? It should be either fixed or removed.

tests/accessor_legacy/accessor_api_image_common.h
```diff
@@ -135,97 +112,44 @@ struct get<sycl::cl_half4> {
...
+sycl::range<imageDims> make_test_range(size_t count);
 
 /**
  * @brief Constructs a range for testing, specialization for 1D images
  * @param count Number of elements to store in entire range
  * @return 1D range
  */
 template <>
-sycl::range<1> make_test_range<1>(size_t count, bool /*isImageArray*/) {
+sycl::range<1> make_test_range<1>(size_t count) {
```
ProGTX, line 130: It seems like the main goal of this PR is to remove image arrays, fixing the `value_type` is a very minor part of the changes here. I think it would be better to split image array removal into a separate PR, or at the very least make it obvious in the commit message and description.

ProGTX (changes requested):

tests/accessor_legacy/accessor_api_common_all.h
bader: Done.

tests/accessor_legacy/accessor_api_image_common.h
bader: I'm not sure. I uncommented it, so let's see what CI has to say about it.
@andylshort, I'd like to hear your thoughts on this as this is your patch.

tests/accessor_legacy/accessor_api_image_common.h
```diff
@@ -1834,6 +1643,7 @@ class check_image_accessor_api_type {
     check_image_accessor_api_dim<T, 1>(log, typeName, count, size, queue);
     check_image_accessor_api_dim<T, 2>(log, typeName, count, size, queue);
     check_image_accessor_api_dim<T, 3>(log, typeName, count, size, queue);
```
bader, line 1645: `check_image_accessor_api_dim` template is removed, so this code doesn't compile.
@andylshort, please, clean-up all dependent code. 
Here if we remove `check_image_accessor_api_dim` here, `check_image_accessor_api_type` doesn't do anything useful anymore, so we need update it and its users.

tests/accessor_legacy/accessor_api_image_common.h
andylshort: As per the SYCL 2020 changes in [Appendix D.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:what-changed-between), "Support for the accessor targets `target::image` and `target::host_image` have been removed". This was causing these tests to fail, as shown in the above CI failure.

These `image` and `host_image` accessors aren't SYCL 2020-compliant and were causing various failures, hence why, even though these are categorised as "legacy" tests, I commented these out. Admittedly I could've provided better commit messages and done better, for which I apologise. This change fixed the bug and a discussion with my team thought this was sufficient enough due to these being legacy tests.

If a new PR is required to take this change further, or if changes need to be made here, do let me know.

tests/accessor_legacy/accessor_api_image_common.h
andylshort: The PR was to fix a bug related to image_array, image, and host_image not being SYCL 2020 compliant, and their accessor functions being replaced by specialised alternatives. This change was a necessary part of fixing this bug.

tests/accessor_legacy/accessor_api_image_common.h
andylshort: This is part of why I commented them out as I was unsure how to address this dependent code. The dependencies on this are `accessor_api_image_core.cpp` and `accessor_api_image_fp16`. Without the non-compliant image accessor code that's causing other tests to fail, as you said these also don't do anything useful anymore. Valid argument for removal of these legacy tests?

bader: Yes. I think we should remove `check_image_accessor_api_type` and all invocations of this template. If callers of this template become no-op, we should remove them too.

tests/accessor_legacy/accessor_api_image_common.h
bader: > I commented these out

I think it's a common practice in such cases to remove the code instead of commenting it out. Commented out code does more harm than good.

andylshort:
@bader I have pushed further commits to my fork to resolve these issues: https://github.com/andylshort/SYCL-CTS/tree/alamzeds/bug-accessor-value_type

bader:
> @bader I have pushed further commits to my fork to resolve these issues: https://github.com/andylshort/SYCL-CTS/tree/alamzeds/bug-accessor-value_type

Thanks! I cherry-picked https://github.com/andylshort/SYCL-CTS/commit/b80e92fab1544c39834a08e45bb88a204b5ff100.
@ProGTX, @keryell, please, take a look.

ProGTX (approved):

keryell (approved):
Thanks!

PR #400, opened 2022-11-17
Add test plan for group functions

Melirius:
Adds a test plan for group functions as described in sections 4.17.3 "Group functions" and 4.17.4 "Group algorithms library" of the SYCL 2020 specification.

### DIFF
--- a/test_plans/group_functions.asciidoc
+++ b/test_plans/group_functions.asciidoc
@@ -68,7 +68,8 @@ If function accepts another types, the following list of types is added:
 Pointers are taken from the same list of types.
 
 `bool Predicate(T)` functions are functions checking the ranges of `T` itself or
-the first element of `vec, marray, struct` type `T`.
+the first element of `vec, marray, struct` type `T`. They are choosen in such a way that
+functions using them should return both `true` and `false` in different tests.
 
 `binary_op` used in scans and reductions are addition and maximum of `T` itself or
 the first element of `vec, marray, struct` type `T`.
@@ -79,9 +80,9 @@ Data for the functions are sequences of 1, 2, 3, etc. corresponding to increasin
 
 === Execution
 
-Functions will be executed by `sycl::parallel_for` with `nd_range<Dims>` corresponding to
-one group of maximal group size available on device with `Dims = 1,2,3` (by querying 
-`info::device::max_work_item_sizes<Dims>` and `info::device::max_work_group_size`).
+Functions will be executed by a group using `sycl::parallel_for` with `nd_range<Dims>`
+corresponding to one group of maximal group size available on device with `Dims = 1,2,3`
+(by querying `info::device::max_work_item_sizes<Dims>` and `info::device::max_work_group_size`).
 Sub-group variants will execute functions on the first subgroup of this large group only.
 
 === Traits
@@ -120,8 +121,11 @@ For each possible `memory_scope` wider than `Group::fence_scope` in the list
 * `sycl::memory_scope::device`
 * `sycl::memory_scope::system`
 
-check existence of `void group_barrier(Group g, memory_scope fence_scope)`
-for both `group` and `sub_group`.
+check that invocation of `void group_barrier(Group g, memory_scope fence_scope)`
+for both `group` and `sub_group` behaves as expected: when each of workitems writes 1
+into zero-initialized local memory array (for `sub_group` and `work_group` memory scope)
+or zero-initialized global memory array (for `device` and `system` memory scope)
+after barrier all array values red by workitems in reverse order are equal to 1.
 
 ==== joint_any_of
 
--- a/test_plans/group_functions.asciidoc
+++ b/test_plans/group_functions.asciidoc
@@ -245,7 +245,7 @@ for each of the following use-cases:
 * using `OutPtr joint_inclusive_scan(Group g, InPtr first, InPtr last, OutPtr result,
 BinaryOperation binary_op);` with some `O binary_op(I, I)`
 * using `OutPtr joint_inclusive_scan(Group g, InPtr first, InPtr last, OutPtr result,
-T init, BinaryOperation binary_op);` with some fundamental type `T` and `T binary_op(T, I)`
+BinaryOperation binary_op, T init);` with some fundamental type `T` and `T binary_op(T, I)`
 
 check returned type and value for both `group` and `sub_group`.
 
@@ -255,7 +255,7 @@ For some fundamental types `T` and `V` for each of the following use-cases:
 
 * using `T exclusive_scan_over_group(Group g, T x, BinaryOperation binary_op);`
 with some `T binary_op(T, T)`
-* using `T exclusive_scan_over_group(Group g, V x, BinaryOperation binary_op, T init);`
+* using `T exclusive_scan_over_group(Group g, V x, T init, BinaryOperation binary_op);`
 with some `T binary_op(T, V)`
 
 check returned type and value for both `group` and `sub_group`.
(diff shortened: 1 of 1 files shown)


test_plans/group_functions.asciidoc
```diff
@@ -0,0 +1,272 @@
...
+=== Data for the functions
+
+Data for the functions are sequences of 1, 2, 3, etc. corresponding to increasing `linear_id`.
+
+=== Execution
+
+Functions will be executed by `sycl::parallel_for` with `nd_range<Dims>` corresponding to
+one group of maximal group size available on device with `Dims = 1,2,3` (by querying 
+`info::device::max_work_item_sizes<Dims>` and `info::device::max_work_group_size`).
+Sub-group variants will execute functions on the first subgroup of this large group only.
```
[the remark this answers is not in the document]
Melirius: Definitely tests should be run for group, I add a sentence about it also. BTW, am I right that group algorithms library functions (4.17.4) can be run also on subgroup?

test_plans/group_functions.asciidoc
```diff
@@ -0,0 +1,272 @@
...
+
+For each possible `memory_scope` wider than `Group::fence_scope` in the list
+
+* `sycl::memory_scope::sub_group`
+* `sycl::memory_scope::work_group`
+* `sycl::memory_scope::device`
+* `sycl::memory_scope::system`
+
+check existence of `void group_barrier(Group g, memory_scope fence_scope)`
+for both `group` and `sub_group`.
```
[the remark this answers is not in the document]
Melirius: No, we test that function exist and also that it accepts aforementioned memory scopes. I'll change the wording, thanks.

It is really difficult to meaningfully test barrier but minimum positive testing can be done. I'll add it.

test_plans/group_functions.asciidoc
```diff
@@ -0,0 +1,272 @@
...
+
+==== joint_any_of
+
+With several different pointer types `Ptr` and `bool Predicate(*Ptr)` check returned type and value
+of `bool joint_any_of(Group g, Ptr first, Ptr last, Predicate pred);` for
+both `group` and `sub_group`.
```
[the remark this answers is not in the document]
Melirius: Yep, I mean it when said about different `Predicate`,  but it should be expressed explicitly.

test_plans/group_functions.asciidoc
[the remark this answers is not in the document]
gmlueck: > Definitely tests should be run for group, I add a sentence about it also. BTW, am I right that group algorithms library functions (4.17.4) can be run also on subgroup?

Yes.  All of those functions have a statement like:

> Available only if `sycl::is_group_v<std::decay_t<Group>>` is true

And we say in [4.17.3.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_group_type_trait):

> `is_group<T>` is `std::true_type` if `T` is the type of a standard SYCL group (`group` or `sub_group`)

test_plans/group_functions.asciidoc
```diff
@@ -0,0 +1,276 @@
...
+* `sycl::memory_scope::sub_group`
+* `sycl::memory_scope::work_group`
+* `sycl::memory_scope::device`
+* `sycl::memory_scope::system`
+
+check that invocation of `void group_barrier(Group g, memory_scope fence_scope)`
+for both `group` and `sub_group` behaves as expected: when each of workitems writes 1
+into zero-initialized local memory array (for `sub_group` and `work_group` memory scope)
+or zero-initialized global memory array (for `device` and `system` memory scope)
+after barrier all array values red by workitems in reverse order are equal to 1.
```
gmlueck, line 128: I think you will need to write the test so that it checks that all these memory scopes are supported by using `info::device::atomic_fence_scope_capabilities`.  The specification says that only `memory_scope::work_group` is required to be supported.  The test should skip any memory scopes that are not supported according to that query.

Adding @pennycook here ... should the spec also guarantee that `memory_scope::sub_group` is supported?  The default memory  scope for `group_barrier<sub_group>()` is `memory_scope::sub_group`.  It seems weird to have a default value that might not be supported by the device.  In any case, isn't it trivial to support `memory_scope::sub_group` if you are required to support `memory_scope::work_group`?

Melirius: Yes, you're right, it should be tested before.

test_plans/group_functions.asciidoc
```diff
@@ -0,0 +1,277 @@
...
+==== group_barrier
+
+For each `memory_scope` wider than `Group::fence_scope` in the list
+
+* `sycl::memory_scope::sub_group`
+* `sycl::memory_scope::work_group`
+* `sycl::memory_scope::device`
+* `sycl::memory_scope::system`
+
+supported by the hardware (from query of `info::device::atomic_fence_scope_capabilities`)
```
gmlueck, line 124:
```suggestion
If supported by the hardware (from query of `info::device::atomic_fence_scope_capabilities`)
```

keryell (approved):
Looks  good.
@Pennycook by looking at all this list in a central place, I was wondering whether the naming is right in the specification for the presence of `group` or not and the place of `group` in the various functions or member functions.

fraggamuffin:
fix then can merge

test_plans/group_functions.asciidoc
bader: @gmlueck, @Pennycook, should I wait for @bashbaug to reply before merging this PR?

bashbaug: Note: I think this is one of the places where SYCL and OpenCL can be different, if desired, though we'll need some care if they are different.

In OpenCL in general, the "subgroup" memory scope is only required when subgroups are supported.

* In OpenCL C it only exists when subgroups are supported ([link](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#memory-scope) - "Requires support for OpenCL C 3.0 or newer and the __opencl_c_subgroups feature.").
* In SPIR-V, the subgroup memory scope is only valid when subgroups are supported ([link](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#validation-rules) - "For OpenCL 2.1, OpenCL 2.2, or OpenCL 3.0 devices supporting Subgroups (where CL_DEVICE_MAX_NUM_SUB_GROUPS is not 0)."  See note below...).
* This means that a device that does not support subgroups does not need to handle a subgroup scope memory fence, say by "upgrading" it to the wider work-group scope.

If SYCL chooses to require the subgroup scope when the work-group scope is also supported it's fine to do so, but it will be SYCL's responsibility to generate valid SPIR-V in this case, specifically by generating a work-group fence instead of a subgroup fence if/when subgroup fences aren't supported.

I agree with most of @pennycook's suggested clarifications above, especially that we should require that the subgroup scope is supported when subgroups are supported.  This is debatably already true, but it is not stated explicitly.

When we do this I also think it makes sense to describe the SPIR-V requirements in terms of the atomic fence capabilities query, rather than subgroups generically.  Again, this is debatably already true, but I think it's a bit clearer and consistent to describe the requirements in this way.

gmlueck: I think we can merge this test plan before this conversation is resolved.  The only impact on the test plan is whether the test can assume that `memory_scope::sub_group` is supported or if it must check to see if the device supports this memory scope.

I suggest that the CTS follow the existing wording of the SYCL spec, which does *not* guarantee that `memory_scope::sub_group` is supported.

I have opened the following issue to continue the conversation about whether the SYCL spec should be changed:

https://gitlab.khronos.org/sycl/Specification/-/issues/639

PR #404, opened 2022-11-21
Fix param type in function accessor_tests_common::get_subscript_overload

aleksmesh:
const attribute added to AccT& accessor parameter in get_subscript_overload function

### DIFF
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -843,7 +843,7 @@ void test_accessor_types_common() {
 }
 
 template <typename T, typename AccT, int dims>
-decltype(auto) get_subscript_overload(AccT& accessor, size_t index) {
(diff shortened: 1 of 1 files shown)


tests/accessor/accessor_common.h
```diff
@@ -843,7 +843,7 @@ void test_accessor_types_common() {
 }
 
 template <typename T, typename AccT, int dims>
-decltype(auto) get_subscript_overload(AccT& accessor, size_t index) {
+decltype(auto) get_subscript_overload(const AccT& accessor, size_t index) {
```
ProGTX, line 846: I'm curious why this was a problem?

ProGTX (approved):

tests/accessor/accessor_common.h
aleksmesh: @ProGTX It's a problem if accessor is const, since AccT is explicitly provided.

PR #408, opened 2022-11-24
Fix compiling bugs in accessor_default_values

aleksmesh:


tests/accessor/accessor_default_values.h
```diff
@@ -13,6 +13,8 @@
 #include <type_traits>
 
 namespace accessor_default_values_test {
+using namespace sycl_cts;
+using namespace accessor_tests_common;
```
ProGTX, line 17: It's usually better to avoid blanket `using namespace` statements, would it be possible to get rid of at least one of them and use proper namespace qualifiers where needed?

aleksmesh: @ProGTX  You are right, "using namespace sycl_cts" is not needed here. Removed.

ProGTX (approved):

PR #410, opened 2022-11-24
Buffers 2020 plan

MathiasMagnus:
Test plan for buffer interface test updates relating to SYCL 1.2.1 >> SYCL 2020.

### DIFF
--- a/test_plans/buffers.asciidoc
+++ b/test_plans/buffers.asciidoc
@@ -3,9 +3,15 @@
 
 = Test plan for buffer (2020 update)
 
-This is a test plan for the changes in APIs described in SYCL 2020 section 4.7.2 compared to SYCL 1.2.1
+This is a test plan for the changes in APIs described in SYCL 2020 compared to SYCL 1.2.1 relating to buffers. These tests don't cover all changes, but a subset.
 
 == Testing scope
+
+Tests intend to cover:
+
+* 4.7.2.1 Buffer interface
+* 4.7.6.4 Properties on buffers
+* 4.7.6.7 Container interface
 
 === Device coverage
 
--- a/test_plans/buffers.asciidoc
+++ b/test_plans/buffers.asciidoc
@@ -1,0 +1,80 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for buffer (2020 update)
+
+This is a test plan for the changes in APIs described in SYCL 2020 section 4.7.2 compared to SYCL 1.2.1
+
+== Testing scope
+
+=== Device coverage
+
+Device coverage does not change. All of the tests described below are performed only on the default device that is selected on the CTS command line.
+
+=== Buffer types
+
+Type coverage changes! All of the tests described below are performed using each of the following `typename T`:
+
+In Regular mode:
+
+* `int`
+* `float`
+* A user-defined class with several scalar member variables and a user-defined default constructor.
+
+In full conformance mode:
+
+* `char`
+* `signed char`
+* `unsigned char`
+* `short int`
+* `unsigned short int`
+* `unsigned int`
+* `long int`
+* `unsigned long int`
+* `long long int`
+* `unsigned long long int`
+* `bool`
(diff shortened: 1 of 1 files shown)


test_plans/buffers.asciidoc
```diff
@@ -0,0 +1,94 @@
...
+For `Dimensions == {2;3}` check that:
+
+* An appropriately sized multi-dimensional buffer initialized with multi-dim indices when read back to host is linearized according to 3.11.1
+
+=== no_init property
+
+For `Dimensions == 1` check that:
+
+* When fully initializing a buffer on device, a subsequent kernel adding a requirement on the buffer by creating a `no_init` `write_only` accessor really calls the only trivial CTOR for every element, even without explictly filling the buffer.
+* When fully initializing a buffer on device, a subsequent kernel adding a requirement on the buffer by creating a `no_init` `write_only` ranged accessor really calls the only trivial CTOR for the covered range, even without explictly filling the buffer.
```
gmlueck, line 93: I don't understand these tests.  A typical SYCL implementation does not ever call the `T` constructor when copying data between host and device.  This is the reason we limit `T` to "device copyable" types.  The idea is that an implementation can just move data via direct memcpy.  This behavior is the same regardless of whether the `no_init` property is specified.

Even when `no_init` is specified, it's only a hint to the implementation that *can* enable optimization.  A conformant implementation is allowed to ignore the property if it wants.  The SYCL spec doesn't provide any guarantee that an optimization occurs.

The spec does provide a guarantee related to `no_init` that would be interesting to test:

> If this is a ranged accessor, this applies only to the elements within the accessor’s range. The values of unwritten elements outside of this range are preserved.

Therefore, an interesting test could be:

* Construct a buffer with initialized values.
* Create a *ranged* accessor to that buffer with `no_init`.
* Pass the accessor to a kernel that does not write any values through the accessor.
* Make sure that the elements *outside* of the range still have their initial values.

test_plans/buffers.asciidoc
```diff
@@ -0,0 +1,94 @@
...
+* After modifying its contents on device, updated contents are written back to the vector upon buffer destruction.
+* The logging allocator's log isn't empty.
+
+=== Linearization
+
+For `Dimensions == {2;3}` check that:
+
+* An appropriately sized multi-dimensional buffer initialized with multi-dim indices when read back to host is linearized according to 3.11.1
+
+=== no_init property
```
gmlueck, line 88: Note that `no_init` is a property on accessors, not buffers.  Only noting this because it seemed a little weird that you have a test for `no_init` in a test plan that is for buffers.

test_plans/buffers.asciidoc
```diff
@@ -0,0 +1,94 @@
...
+
+In addition, if the device has `aspect::fp16`, the following type is tested:
+
+* `sycl::half`
+
+== Tests
+
+=== SYCL1.2.1 compatible interface
+
+CTS contains tests for `buffer` SYCL1.2.1 interface. Some functions have been marked as deprecated in SYCL2020. These tests should be marked as deprecated and be enrolled into the machinery disabling legacy tests. These functions are the following:
```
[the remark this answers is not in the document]
MathiasMagnus: That I know, but there's a specific CMake option `SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS` disabling these tests. Not all features marked deprecated in 2020 have been subjected to this option yet, such as the ones listed here.

test_plans/buffers.asciidoc
MathiasMagnus: Intel seems to have a working set of no_init tests internally that'll get upstreamed sometime in the future, so it's safe to remove it from this plan.

test_plans/buffers.asciidoc
MathiasMagnus: ditto

keryell (approved):
That looks good, even if it would be clearer to repeat the content of *In Regular mode:*  into *In full conformance mode:* part.

gmlueck (approved):

PR #411, opened 2022-11-25
 Fix tests for accessor iterators requirements

aleksmesh:
Fix tests for iterators requirements for sycl::accessor and sycl::host_accessor.

### DIFF
--- a/util/type_traits.h
+++ b/util/type_traits.h
@@ -254,8 +254,8 @@ struct addition : std::false_type {};
 
 template <typename LeftOperand, typename RightOperand>
 struct addition<LeftOperand, RightOperand,
-                std::void_t<decltype(std::declval<LeftOperand>() +
-                                     std::declval<RightOperand>())>>
+                std::void_t<decltype(std::declval<LeftOperand&>() +
+                                     std::declval<RightOperand&>())>>
     : std::true_type {};
 
 template <typename LeftOperand, typename RightOperand>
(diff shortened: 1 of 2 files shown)


aleksmesh:
CI failed due to old version of DPC++ compiler. I don't have permission to change cts_ci.yml, @bader could you uplift DPC++ compiler to newer version to check this PR?

util/type_traits.h
```diff
@@ -254,8 +254,8 @@ struct addition : std::false_type {};
 
 template <typename LeftOperand, typename RightOperand>
 struct addition<LeftOperand, RightOperand,
-                std::void_t<decltype(std::declval<LeftOperand>() +
-                                     std::declval<RightOperand>())>>
+                std::void_t<decltype(std::declval<LeftOperand&>() +
+                                     std::declval<RightOperand&>())>>
```
ProGTX, line 258: This is a strange change, why is it needed? I'm guessing because the type is not defined yet? If that's the case, this is probably preferable:
```suggestion
                std::void_t<decltype(std::declval<const LeftOperand&>() +
                                     std::declval<const RightOperand&>())>>
```

util/type_traits.h
```diff
@@ -266,8 +266,8 @@ struct subtraction : std::false_type {};
 
 template <typename LeftOperand, typename RightOperand>
 struct subtraction<LeftOperand, RightOperand,
-                   std::void_t<decltype(std::declval<LeftOperand>() -
-                                        std::declval<RightOperand>())>>
+                   std::void_t<decltype(std::declval<LeftOperand&>() -
+                                        std::declval<RightOperand&>())>>
```
ProGTX, line 270:
```suggestion
                   std::void_t<decltype(std::declval<const LeftOperand&>() -
                                        std::declval<const RightOperand&>())>>
```

util/type_traits.h
aleksmesh: Actually only change for LeftOperand is needed, removed others.
It is needed similarly to reference on line 242 - to handle rval of first operand for subtraction operator.

ProGTX: The problem I see is that binary addition is a non-modying operation, so a non-const reference should not be needed or requested.

aleksmesh: @ProGTX You are right, it shouldn't work like that. Removed this change and revert enabling this test for now.

keryell (approved):
So at the end it is just modernizing C array to `std::array`.

ProGTX (approved):

PR #424, opened 2022-12-07
Find to adapt

MathiasMagnus:
This PR is a simplification of how different toolchains being verified by the CTS are discovered and used. There are two important aspects to the changes proposed here:

- separation of concern
  - Currently the onus of detecting toolchains and adapting them to CTS use is conflated in the self-hosted FindXYZ.cmake scripts. Not only does it run the risk of becoming out of sync with distributions (as it already has), but it also complicates the scripts for not much added value (IMHO).
  - The proposed setup leverages detection scripts shipping with toolchain installations for detecting the toolchain components.
  - It is only Intel SYCL which ships no CMake support for compiling SYCL applications, therefore `FindDPCPP.cmake` has been retained for use with Intel SYCL and DPCPP installations.
- consolidation
  - The proposed setup is a stepping stone towards nearly eliminating all helper scripts from the CTS itself. It leverages the fact that ComputeCpp and hipSYCL both expose an `add_sycl_to_target` function which can be used to deduplicate a lot of script logic currently in use.

It is up to the user to supply the same canonical variables driving usual toolchain detection, such as `hipSYCL_DIR`, `ComputeCpp_DIR` or `DPCPP_INSTALL_DIR`. With those in place, configuration and builds should pass.

Tested on:
- Linux (ComputeCpp, hipSYCL, Intel SYCL)
- Windows (ComputeCpp)

_Note: the current implementation depends on ComputeCpp shipping `ComputeCppConfig.cmake` with their compiler which seems to not being the case for the 2.8.0 install in CI but is true for more recent packages. 2.11.0 has them (although the experimental package is missing them in Windows, but "luckily" there's no Windows CI)._


ProGTX (approved):
We're looking to update the Docker image, this looks good.

bader:
@MathiasMagnus, please, update your branch to pull new ComputeCPP version.
Please, take a look at the failure for hipSYCL implementation.

MathiasMagnus:
The failures on hipSYCL (@psalz) and ComputeCpp (@ProGTX) are for the following reasons:

- The hipSYCL failure stems from `configure.sh` having `CMAKE_PREFIX_PATH=/sycl/bin` instead of `CMAKE_PREFIX_PATH=/sycl/bin`, which was mildly redundant, because `FindhipSYCL.cmake` did have a `PATH_SUFFIXES bin` in its call to `find_program()`. I could hotfix it by sprinkling ``if [[ "${{matrix.sycl-impl}}" == "hipsycl" ]]; then echo -DhipSYCL_ROOT=/sycl; fi;` \` into the CI script where `configure.sh` instead of asking for an image update, but updating `configure.sh` would be cleaner, especially how it wouldn't even break existing CI, given how `FindhipSYCL.cmake` is written.
- The ComputeCpp failure stemmed from not configuring with `-DSYCL_LANGUAGE_VERSION=2020` which I could hardcode into `AdaptComputeCpp.cmake` when defining the target `SYCL::SYCL`, however then there's a redifinition of macro for any target (all the tests) not just linking to SYCL::SYCL but also calling `add_sycl_to_executable` which have already received integration headers setting `2017`. I could come up with a hack to set this variable prior to calling `find_package`, but if the macro `add_sycl_to_executable` needs this value to be set prior to looking for ComputeCpp, it would be best if it was set on the command-line. Once again, it could be hacked into the CI scripts, as a hotfix.

In the spirit of not leaking things into `cts_ci.yml` that are likely better be placed in `configure.sh`, should we put these additions there?

psalz:
Thanks for looking into this! I've made the suggested changes to the configure scripts in #444.

> The proposed setup is a stepping stone towards nearly eliminating all helper scripts from the CTS itself. It leverages the fact that ComputeCpp and hipSYCL both expose an add_sycl_to_target function which can be used to deduplicate a lot of script logic currently in use.

Last I checked I think the problem with using `add_sycl_to_target` was that it doesn't really work with the "build once, link into multiple executables" approach currently used by the CTS. But we can revisit that.

MathiasMagnus:
@bader @ProGTX I rebased on top of #444 but ComputeCpp still fails, because the image didn't get the new `config.sh`:
```
PS C:\Users\mate> docker run --interactive --tty --rm scottstraughan/sycl-cts-ci:computecpp-2.11.0
root@69755b077ad8:/# cat /scripts/configure.sh
#!/bin/bash

set -o errexit -o pipefail -o noclobber -o nounset

cmake . -G Ninja -B build \
    -DSYCL_IMPLEMENTATION=ComputeCpp \
    -DCMAKE_PREFIX_PATH=/sycl \
    -DCMAKE_BUILD_TYPE=Release \
    -DSYCL_CTS_ENABLE_FULL_CONFORMANCE=0 \
    $@
```
The change `-DSYCL_LANGUAGE_VERSION=2020` is still missing from the configuration script, hence the build errors.

keryell (commented):
Would it be possible to factorize out most of `cmake/AdaptDPCPP.cmake` and `cmake/AdaptComputeCpp.cmake` ?
These files are pretty similar.

MathiasMagnus:
> Would it be possible to factorize out most of `cmake/AdaptDPCPP.cmake` and `cmake/AdaptComputeCpp.cmake` ? These files are pretty similar.

Most certainly. I don't know if that's the direction we want to go. My understanding of the status quo is that:
- `add_sycl_to_executable` is the _de facto_ support implementations give as CMake support. This is mostly a gentlemen's agreement.
- `add_sycl_executable_implementation` is currently a mostly superfluous extra escape hatch serving as the entrypoint for the CTS to translate it's own options to the implementation at hand.
  - This currently is very lightly used, but until implementers don't agree on a CMake interface or upstream some form of abstraction, these files will soon diverge. Take the only meaningful difference for eg. `target_compile_definitions(SYCL::SYCL INTERFACE SYCL_LANGUAGE_VERSION=${SYCL_LANGUAGE_VERSION})` of ComputeCpp. Currently only it has the ability to operate in both 1.2.1 and 2020 modes. The CTS (for whatever reason) has no option to control which standard version it's testing. I expect such an option be introduced shortly. How standard versions are set isn't part of the `add_sycl_to_executable` gentlement's agreement, so it needs to be translated to the impl's liking. Whether that's a preprocessor definition or a compilation option `-std=sycl2020` is implementation defined.

While it hurts my eyes too, I don't feel confident enough yet to abstract the current of interfaces. As soon as the 2020 CTS is finished, this "cosmic alignment" of interfaces will immediately be challanged.

MathiasMagnus:
>  I'm not sure if this image should include configure.sh, but if so, we need Codeplay to update the image. @ProGTX, could you help with this?
We probably can update this script from sources to make sure that the script version is latest. @psalz, what do you think?

Do note that mildly related are the changes [proposed](https://github.com/codeplaysoftware/computecpp-sdk/pull/291) for ComputeCppConfig.cmake that fix Windows use (the import detection was borked), but the CTS also outright failed with Ninja, because
> [IMPLICIT_DEPENDS and DEPFILE can not both be specified.](https://github.com/Kitware/CMake/blob/cda4310a82f2a6e107f7ddd00cfd434ca31321b4/Source/cmAddCustomCommandCommand.cxx#L302-L307)

This latter is likely fixed by [this](https://github.com/codeplaysoftware/computecpp-sdk/pull/291/commits/d52090789d763e0fafdf7c7b6b0c802d0df490c9) commit which uses either IMPLICIT_DEPENDS or DEPFILE but never both. Moving to using ComputeCppConfig.cmake also means having to clean it up minimally for use by the CTS.

bader:
> * The CTS (for whatever reason) has no option to control which standard version it's testing.

The CTS uses git branches to control the standard version it's testing. `SYCL-2020` branch tests SYCL 2020 version, `SYCL-1.2.1/master` tests SYCL 1.2.1.

MathiasMagnus:
> > * The CTS (for whatever reason) has no option to control which standard version it's testing.
> 
> The CTS uses git branches to control the standard version it's testing. `SYCL-2020` branch tests SYCL 2020 version, `SYCL-1.2.1/master` tests SYCL 1.2.1.

I did not mean to sound condescending, the "whatever reason" can be multitude of things, including not _yet_ having transitioned to a unified codebase (like the OpenCL CTS), or not seeing the engineering effort justified for a standard that's a language, not a library.

keryell:
On the other hand, does it makes sense to add an option to support SYCL 1.2.1 nowadays? Only SYCL 2020 and the future matter. :-)

MathiasMagnus:
That was just an example, but reduced and full feature set may become options.

keryell (approved):
OK, let's try this and we can improve even more later.

keryell:
What we doing on this? Ready to merge?

bader:
> What we doing on this? Ready to merge?

We need to fix `SYCL CTS CI / compile-cts (computecpp, 2.11.0) ` check before the merge. I think @MathiasMagnus is still working on it. @MathiasMagnus, am I right?

MathiasMagnus:
I'm still waiting on the resolution of [this](https://github.com/KhronosGroup/SYCL-CTS/pull/424#issuecomment-1363541240). I can apply a hotfix in the CI script if we want to push it through:
```
run: |
          bash /scripts/configure.sh \
            -DSYCL_CTS_EXCLUDE_TEST_CATEGORIES="${{ env.container-workspace }}/ci/${{ matrix.sycl-impl }}.filter" \
            `if [[ "${{matrix.sycl-impl}}" == "computecpp" ]]; then echo -DSYCL_LANGUAGE_VERSION=2020; fi;` \
            -DSYCL_CTS_MEASURE_BUILD_TIMES=ON
```
Either way it's fine by me. I checked and the changes [proposed](https://github.com/codeplaysoftware/computecpp-sdk/pull/291) for the ComputeCpp SDK config script don't need to be in the image. That's only critical to get Windows working. So it's either to hotfix above, a similar hotfix going into the CMake scripts, or getting the `config.sh` script to update inside the image.

keryell:
Would it be solved at the same time the CI is updated to use Clang as a host compiler for ComputeCpp as discussed along https://github.com/KhronosGroup/SYCL-CTS/pull/483 ?

MathiasMagnus:
@bader @keryell CI should be passing now (it already did in my branch). The only change are the last two commits. As the long commit message mentions, unfortunately the [Update implicit dependence on headers](https://github.com/codeplaysoftware/computecpp-sdk/pull/291/commits/d52090789d763e0fafdf7c7b6b0c802d0df490c9) commit inside my batch of [Various CMake fixes](https://github.com/codeplaysoftware/computecpp-sdk/pull/291) is required for ComputeCpp to configure with Ninja. The hotfix of conditionally specifying `-DSYCL_LANGUAGE_VERSION=2020` in the CI script had to be turned into a patch file, because the Ninja generator is hardwired into `/scripts/configure.sh`, which needs to be patched. Instead of submitting an alternative configure script, I figured a patch is nicer, as it clearly shows the delta from the image.

bader (approved):
LGTM, but I'd like @illuhad to approve hipSYCL changes.

cmake/AdapthipSYCL.cmake
```diff
@@ -25,6 +17,10 @@ function(add_sycl_executable_implementation)
     add_library(${object_lib_name} OBJECT ${test_cases_list})
     add_executable(${exe_name} $<TARGET_OBJECTS:${object_lib_name}>)
 
+    # hipSYCL needs the macro to be called on the executable target. Setting
+    # properties on source files is insufficient.
+    add_sycl_to_target(TARGET ${exe_name} SOURCES ${test_cases_list})
```
illuhad, line 22: I think we also need to `add_sycl_to_target()` on the target `object_lib_name`. It seems that the executable is only involved when everything is linked together, while the compilation target is `object_lib_name`. Not setting this causes the system C++ compiler to be used to compile CTS tests, instead of hipSYCL, which would only be invoked for linking. Clearly this is not what we want ;)

illuhad (changes requested):

ProGTX:
Quick note that the ComputeCpp Docker image has been updated, including the patch proposed here.

.github/workflows/cts_ci.yml
```diff
@@ -128,6 +128,7 @@ jobs:
     container:
       # image: khronosgroup/sycl-cts-ci:${{ matrix.sycl-impl }}-${{ matrix.version }}
       # ComputeCpp images are hosted by Codeplay
+      # TODO: remove patching ComputeCpp configuration when image gets updated
```
bader, line 131: Do we still need any changes in this file?

MathiasMagnus: True... removed. (waiting for CI to finish before blasting it again with a comment-only change)

cmake/AdapthipSYCL.cmake
MathiasMagnus: I fixed which compiler/linker gets invoked. This actually surfaced the fact that the `test_all` target did not use the compiler launcher previously, or did so in an underhanded way. It is a linkage of object files containing SYCL code, so `add_sycl_to_target` should be in order.

.github/workflows/cts_ci.yml
bader: You shouldn't worry about CI. CI will run tests for all commits, and you will be able to review all results.
BTW, here are results for https://github.com/KhronosGroup/SYCL-CTS/pull/424/commits/a0e720dc28c3126a28de7b6f6e83d01dc07d9443 - https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4116155955/jobs/7105833324. Patched ComputeCPP build fails.

MathiasMagnus: Pushed removal.

.github/workflows/cts_ci.yml
```diff
@@ -136,6 +136,7 @@ jobs:
           submodules: true
       - name: Configure CMake
         working-directory: ${{ env.container-workspace }}
+        shell: bash
         run: |
           bash /scripts/configure.sh \
```
bader, line 141: Technically, this is not needed too. `run` command invokes configuration script from bash.
As an alternative, we can keep shell configuration in GitHub Actions script and simplify the `run` command by dropping `bash` command.

MathiasMagnus: Yeah, another leftover from from the patching, which used bash `if [[]]` syntax.

illuhad (approved):
I think it's good now, thank you!

bader:
@illuhad, please, check if your comments were addressed.

PR #425, opened 2022-12-07
Windows, MSVC, Clang[-cl] fixes

MathiasMagnus:
Builds are extremely noisy on Windows, even with moderate diagnostic levels, but on higher, warnings are outright unusable. I started chipping away at the most noisy sources, typically found in `common/common.h` and other universally included headers or TUs which are related to testing utilities, that are built on every rebuild, regardless of which test one is developing.

Most of the warnings are related to not conforming to ISO, unwanted casting, some are toolchain related limitations, but a few seem quite serious and was just buried among other warnings. My favorite:
```
double_double r = {-0.0 -0.0 };
// vs.
double_double r = {-0.0, -0.0 };
```
which seems to be overly explicit in wanting to init both fields of `double_double` to negative zero, but because of a missing comma, the second field remains to be just zero initialized.

_Note: some formatting changes made it into the change set, resulting from auto-configured format-on-save in IDE. If it's an issue, I can revisit those changes and revert formatting changes._


ci/computecpp.filter
```diff
@@ -20,7 +20,7 @@ nd_item
 opencl_interop
 platform
 reduction
-specialization_constants
+spec_const
```
bader, line 23: What is the reason for the renaming? Does using `specialiation_constants` name causes MSVC to issue warnings?

oclmath/reference_math.cpp
```diff
@@ -2469,6 +2472,7 @@ static inline void mul128( cl_ulong a, cl_ulong b, cl_ulong *hi, cl_ulong *lo )
 
 // Move the most significant non-zero bit to the MSB
 // Note: not general. Only works if the most significant non-zero bit is at MSB-1
+/*
```
bader, line 2475: Please, do not leave commented out code. Please, remove it if it's not used by CTS.

tests/spec_const/spec_const_exceptions_throwing_common.h
```diff
@@ -3,21 +3,22 @@
 //  SYCL 2020 Conformance Test Suite
 //
 //  Common code for expected exceptions throwing by specialization constants.
-//  In this tests we check that exception with code sycl::errc::invalid is thrown
+//  In this tests we check that exception with code sycl::errc::invalid is
+thrown
```
bader, line 7: I suggest using clang-format for code formatting. Looks like `//` is missing here.

tests/spec_const/spec_const_exceptions_throwing_core.cpp
```diff
@@ -2,15 +2,16 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-//  Provides tests for specialization constants throwing exceptions when expected
+//  Provides tests for specialization constants throwing exceptions when
+expected
```
bader, line 6: ditto

tests/spec_const/spec_const_exceptions_throwing_fp16.cpp
```diff
@@ -2,16 +2,17 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-//  Provides tests for specialization constants throwing exceptions for sycl::half
+//  Provides tests for specialization constants throwing exceptions for
+sycl::half
```
bader, line 6: ditto

tests/spec_const/spec_const_multiple_via_kb_fp64.cpp
```diff
@@ -10,17 +10,17 @@
...
+#define TEST_NAME spec_const_multiple_via_kb_fp64
 
 namespace TEST_NAMESPACE {
 using namespace sycl_cts;
 
 /** test specialization constants for double
  */
 class TEST_NAME : public sycl_cts::util::test_base {
-public:
+ public:
```
bader, line 23: I suggest we keep existing formatting and move this change to a separate PR, unless adding 1-space indentation fixes some warnings on Windows which will be surprising (😮).

util/sycl_exceptions.h
```diff
@@ -86,6 +86,7 @@ class matcher_exception_category : public Catch::Matchers::MatcherGenericBase {
 #if SYCL_CTS_SUPPORT_HAS_EXCEPTION_CATEGORY == 0
     // There should be no compilation failures, but every CHECK_THROWS_MATCHES
     // should fail with this matcher
+    (void)other;
```
bader, line 89:
```suggestion
    std::ignore = other;
```

bader (changes requested):

oclmath/reference_math.cpp
MathiasMagnus: This code wasn't used by anything, that's why I got a linker warning. (Unreferenced function). However the logic seemed tricky/hard enough to cause trouble when trying to reproduce. It's still in Git history, but it's less apparent to a future user that at one point we had such a function. My rationale in commenting was trying to avoid someone reinventing this wheel in the future, but it can be removed if you insist.

tests/spec_const/spec_const_exceptions_throwing_common.h
MathiasMagnus: Do correct me if I'm wrong, (I'm still fairly new to clang-format), but this formatting mishap (like all of the following ones) are a result of applying clang-format on the files. I have VS Code configured to run clang-format on save. Applying the `.clang-format` file found in the repo puts "thrown" on the next line without a leading `//`. If that is an issue, someone should look into the `.clang-format` file. I checked being super specific:
```
C:\Kellekek\LLVM\15.0.3\bin\clang-format.exe --style=file:C:\Users\mate\Source\Repos\sycl-cts\.clang-format -i .\tests\spec_const\spec_const_exceptions_throwing_common.h
```
and it redoes this change.

tests/spec_const/spec_const_exceptions_throwing_core.cpp
MathiasMagnus: ditto

tests/spec_const/spec_const_exceptions_throwing_fp16.cpp
MathiasMagnus: ditto

util/sycl_exceptions.h
MathiasMagnus: If we're aiming for beauty and setting precedence, how about [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) on the function arg?

ci/computecpp.filter
ProGTX: Would `spec_constants` work? Minor change, but slightly more understandable.

oclmath/reference_math.cpp
ProGTX: It's better to remove it IMO. We can add it somewhere for reference (a GitHub issue?).

tests/spec_const/spec_const_exceptions_throwing_common.h
ProGTX: It's because the whole thing is already in a `/**/` comment block. It's strange, we might want to change that, but for now it's probably safer to add `//`

ProGTX (approved):

tests/spec_const/spec_const_exceptions_throwing_common.h
bader: > It's because the whole thing is already in a `/**/` comment block. It's strange, we might want to change that, but for now it's probably safer to add `//`

I didn't notice that. +1 for not mixing two commenting styles. It's very confusing.

util/sycl_exceptions.h
bader: The attribute works for this case too.

tests/spec_const/spec_const_multiple_via_kb_fp64.cpp
MathiasMagnus: I'm fine either way, but this formatting comes from the format file: `AccessModifierOffset: -1` which pulls it in one character as opposed to regular indentation in the scope, which would be two. I've come to set my IDE to format changes only, but it is mildly annoying that formatting isn't universal. (I do understand the drawback of losing history just because enforcing formatting in a retrospecitve manner.)

MathiasMagnus:
I triaged all outstanding comments. I may be a terrible engineer, but I don't get how formatting is supposed to work on this repo.

My last [push](https://github.com/KhronosGroup/SYCL-CTS/pull/425/commits/9581b35001693169a5a2c8c8b6ff19e6bb0d2db7) triggered an action, that has [this](https://github.com/KhronosGroup/SYCL-CTS/actions/runs/3732641656#summary-10184960106) summary on the check-clang-format step. However, if I issue the very same command on my Windows dev box, I get no diffs. In CI
```
DIFF=$( git diff -U0 --no-color 5928b3b750fd2518a913d0d2476d618c55475337 | clang-format-diff-14 -p1 )
  if [ ! -z "$DIFF" ]; then
    echo 'The following changes are not formatted according to `clang-format`:' >> $GITHUB_STEP_SUMMARY
    ....
```
On my box:
```
PS C:\Users\mate\Source\Repos\sycl-cts [win-msvc-clangcl-fixes ≡]> git diff -U0 --no-color 5928b3b750fd2518a913d0d2476d618c55475337 | python.exe C:\Kellekek\LLVM\15.0.6\share\clang\clang-format-diff.py -p1 -binary C:\Kellekek\LLVM\15.0.6\bin\clang-format.exe
PS C:\Users\mate\Source\Repos\sycl-cts [win-msvc-clangcl-fixes ≡]> $LASTEXITCODE
0
```
How am I supposed to format my code?? I'll be burning a lot of CI time this way, given how I can't detect formatting errors locally, especially how unlike in the OpenCL ecosystem repos, there's no button to cancel a running job.

bader:
> How am I supposed to format my code?

You can go the summary of GitHub actions and copy the patch prepared by clang-format tool. Here is the link to the latest CI run - https://github.com/KhronosGroup/SYCL-CTS/actions/runs/3732641656#summary-10184960106.

bader:
@MathiasMagnus, please, avoid force-pushing. It's hard to track what changes you applied after first round of code review.

bader (approved):

bader:
@MathiasMagnus, please, address regressions with computecpp and hipsycl implementations.

psalz:
> How am I supposed to format my code?? I'll be burning a lot of CI time this way, given how I can't detect formatting errors locally, especially how unlike in the OpenCL ecosystem repos, there's no button to cancel a running job.

It's because you're running clang-format 15, while CI is using 14. There are these nasty breaking changes in the way formatting is done with pretty much every LLVM release, and I'm afraid there is no other way around it except everybody using the same version (or manually applying the patch, as suggested by @bader).

oclmath/reference_math.cpp
```diff
@@ -71,15 +71,27 @@
 
 // Declare Classification macros for non-C99 platforms
 #ifndef isinf
-    #define isinf(x)    (	sizeof (x) == sizeof(float )	?	fabsf(x) == INFINITY  	\
-                        :	sizeof (x) == sizeof(double)	?	fabs(x) == INFINITY  	\
-                        :	fabsl(x) == INFINITY)
+#define isinf(x)                                    \
```
keryell, line 74: Why not defining a function instead of using a macro calling a lambda?

MathiasMagnus: I didn't want to challange too many things. [`isinf`](https://en.cppreference.com/w/c/numeric/math/isinf) et al are supposed to be macros, and I didn't know if it would trigger allergic reactions if I changed that. The issue with the previous macro/implementation was that while the optimizer can always dead-code eliminate the branch of the select operator that's not taken, but the front-end still parses it and issues the diagnostic of implicit down-casting. `if constexpr` on the other hand is free of that.

MathiasMagnus:
I updated the branch. (Apologies @bader for the force push, but it makes a generally cleaner history and removes purely formatting commits and noise due to initially working with clang-format-15 set as the formatting tool.)

The changes that happened in this push compared to the previous version:
- Less formatting noise.
- Removed Catch2 hash bump. Will revisit this once proper Windows CI is in place or if I find the time to repro the compilation issue caused by it's absence.
- Removed "Incorrect no. of arguments to function like macro" commit which became obsoleted by work on `accessor_iterator_requirement.cpp`
- "Remove dead code" commit is atomically just removing the code, instead of commenting it out in another then later removing it, as per the suggested resolution by @ProGTX [here](https://github.com/KhronosGroup/SYCL-CTS/pull/425#discussion_r1043581688)

ci/computecpp.filter
MathiasMagnus: Applied.

tests/spec_const/spec_const_exceptions_throwing_common.h
MathiasMagnus: Fixed. There should be no mixed commenting styles anymore.

bader:
@MathiasMagnus, latest rebase haven't launched CI jobs. I'll try close/re-open to trigger CI.

oclmath/reference_math.cpp
bader: @keryell, are you okay to address this comment later?

keryell: Sure. No problem.

keryell (approved):
Thanks.

PR #429, opened 2022-12-09
Configure OpenCL proxy before configuring SYCL executable

steffenlarsen:
This commit moves the OpenCL proxy configuration into a separate cmake file (cmake/AddOpenCLProxy.cmake) and includes it before including AddSYCLExecutable. This is motivated by CMake on Windows failing to configure for DPC++ due to the OpenCL proxy picking Clang as a C compiler due to FindDPCPP overwriting the compilers.

Additionally, since the OpenCL proxy needs options to be correctly configured the CTS options are declared before including both the OpenCL proxy and the SYCL executable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,6 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
 endif()
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
-include(AddSYCLExecutable)
 include(AddCTSOption)
 
 add_cts_option(SYCL_CTS_ENABLE_FULL_CONFORMANCE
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -66,25 +65,8 @@ add_cts_option(SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
...
-if(SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS OR SYCL_IMPLEMENTATION STREQUAL "ComputeCpp")
-    find_package(OpenCL 1.2 REQUIRED)
-    target_link_libraries(OpenCL_Proxy INTERFACE OpenCL::OpenCL)
-else()
-    add_submodule_directory(vendor/OpenCL-Headers)
-    target_link_libraries(OpenCL_Proxy INTERFACE OpenCL::Headers)
-endif()
-# ------------------
+include(AddOpenCLProxy)
+include(AddSYCLExecutable)
```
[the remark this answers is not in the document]
steffenlarsen: Good idea! I have added such a check in AddSYCLExecutable.

bader (approved):

psalz (commented):
Why is it a problem that CMake uses Clang for the proxy library? I find that doubly confusing actually, given that it's only an interface library which doesn't need to be compiled in the first place.

steffenlarsen:
> Why is it a problem that CMake uses Clang for the proxy library? I find that doubly confusing actually, given that it's only an interface library which doesn't need to be compiled in the first place.

Something inside the OpenCL headers configuration makes it identify it as a C project and will therefore look for a C compiler. Some internal logic picks Clang as the C compiler for Windows after DPC++ has been configured, which in turn makes CMake panic because of the MSVC compiler selected when the language is first set in CMake.

cmake/AddSYCLExecutable.cmake
```diff
@@ -5,6 +5,14 @@ if (NOT ${SYCL_IMPLEMENTATION} IN_LIST KNOWN_SYCL_IMPLEMENTATIONS)
         "-DSYCL_IMPLEMENTATION=[Intel_SYCL,DPCPP;ComputeCpp,hipSYCL]")
 endif()
 
+if(NOT DEFINED SYCL_CTS_OPENCL_PROXY_CONFIGURED)
```
psalz, line 8: Can't you just do `if(NOT TARGET OpenCL_Proxy)` here?

steffenlarsen: You're right, that is much better! Thanks! 😄

psalz (approved):
LGTM!

ProGTX (approved):

PR #432, opened 2022-12-13
Check if tests for `sycl::cl_`, `sycl::half` and `sycl::byte` should be disabled in scalars tests

nilsfriess:
The tests in `tests/scalars` did not take the CMake variables `SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS`, `SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS`, and `SYCL_CTS_ENABLE_HALF_TESTS` into account. Now
- the tests for the `sycl::cl_` types are only compiled if `SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS` is defined,
- the tests using `sycl::byte` are only compiled if `SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS` is defined (since `sycl::byte` is deprecated according to [section 4.14.1 of the SYCL 2020 spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_scalar_data_types)), and
- the tests using `sycl::half` are only compiled if `SYCL_CTS_ENABLE_HALF_TESTS` is defined.

### DIFF
--- a/tests/scalars/CMakeLists.txt
+++ b/tests/scalars/CMakeLists.txt
@@ -1,8 +1,8 @@
 
-if(SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS)
-  file(GLOB test_cases_list *.cpp)
-else()
-  file(GLOB test_cases_list scalars_sycl_types.cpp)
+file(GLOB test_cases_list *.cpp)
+
+if(NOT SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS)
+  list(FILTER test_cases_list EXCLUDE REGEX scalars_interopability_types.cpp)
 endif()
 
 add_cts_test(${test_cases_list})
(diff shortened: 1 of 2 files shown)


tests/scalars/CMakeLists.txt
```diff
@@ -1,3 +1,8 @@
-file(GLOB test_cases_list *.cpp)
+
+if(SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS)
+  file(GLOB test_cases_list *.cpp)
+else()
+  file(GLOB test_cases_list scalars_sycl_types.cpp)
+endif()
```
bader, line 6: I think it's better to apply the opposite strategy and explicitly remove source files with OpenCL interop tests. E.g. https://github.com/KhronosGroup/SYCL-CTS/blob/62155afca2514a208324a9df5d8fd39eddb3caee/util/CMakeLists.txt#L4-L6.

bader (changes requested):

tests/scalars/CMakeLists.txt
nilsfriess: Done

bader (approved):

bader:
@nilsfriess, please, apply clang-format suggestions.

ProGTX (approved):

PR #433, opened 2022-12-13
Add test plan for stream class

nolmoonen:
Adds a test plan for the stream class as described in Section 4.16.1. of the SYCL 2020 specification.

### DIFF
--- a/test_plans/stream_class.asciidoc
+++ b/test_plans/stream_class.asciidoc
@@ -1,0 +1,33 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for stream class interface
+
+This is a test plan for the interface of the stream class as described in Section 4.16.1 of the SYCL 2020 specification. The existing tests are updated from SYCL 1.2.1 to SYCL 2020. The estimated development time is half a day.
+
+== Testing scope
+No negative test are included. Specifically, the statement
+
+[source]
+----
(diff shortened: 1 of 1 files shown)


test_plans/stream_class.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+
+=== Backend coverage
+
+All the tests described below are not backend-specific and are performed for any SYCL backend.
+
+=== Device coverage
+
+All tests construct a test device for which conformance is assessed. All the tests described below are performed once for that test device.
+
+== Tests
```
gmlueck, line 25: I'm not sure if you are trying to capture existing things that should be *removed* from the CTS.  SYCL 1.2.1 allowed the following stream operand types, but these have all been removed from SYCL 2020.  The CTS should not test that these types can be operands of the stream operator:

* `cl_char`
* `cl_uchar`
* `cl_int`
* `cl_uint`
* `cl_short`
* `cl_ushort`
* `cl_long`
* `cl_ulong`
* `sycl::byte`
* `cl_float`
* `cl_double`
* `cl_half`

nolmoonen: The tests added by this test plan are not run for each type, but I have now specified in the test plan that the deprecated types should be removed from the existing stream tests.

gmlueck (approved):

fraggamuffin:
ok to commit

PR #435, opened 2022-12-13
Add test plan for group class

nolmoonen:
Adds a test plan for the stream class as described in Section 4.9.1.7. of the SYCL 2020 specification.

### DIFF
--- a/test_plans/group.asciidoc
+++ b/test_plans/group.asciidoc
@@ -1,0 +1,71 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for group class
+
+This is a test plan for the group class as described in Section 4.9.1.7. of the SYCL 2020 specification.
+
+Estimated development time is two and a half days.
+
+== Testing scope
+
+=== Backend coverage
+
+All the tests described below are not backend-specific and are performed for any SYCL backend.
+
+=== Device coverage
+
+All tests construct a test device for which conformance is assessed. All the tests described below are performed once for that test device.
+
+== Tests
+
+=== Existing test
+The existing test in `group/group_api.cpp` defines some greater than zero group counts (2, 4, 8) with greater than zero group sizes (4, 3, 2), in three dimensions. A kernel is dispatched that calls the below functions, gathers the results for one work-item in each work-group and checks the return types.
+
+- `group.get_id().get(int)`
+- `group.get_id(int)`
+- `group.get_local_range().get(int)`
+- `group.get_local_range(int)`
+- `group.get_global_range().get(int)`
+- `group.get_global_range(int)`
+- `group.get_group_range().get(int)`
+- `group.get_group_range(int)`
+- `group.operator[]`
+
+After the terminates, the results and the return type check is verified.
+
+=== Modification
+
+The test is modified in the following way. The results are gathered for each work-item in every work-group, instead of for one work-item in every work-group. Furthermore, the expected results are initialized to some special value (not zero), to make sure the value is overwritten by the kernel. The following function tests are marked as deprecated:
+
+- `group.get_id().get(int)`
+- `group.get_id(int)`
+- `group.get_global_range().get(int)`
+- `group.get_global_range(int)`
+
+The following functions are added, and their return types and returned values are checked:
+
+- `group.get_group_id().get(int)` and `group.get_group_id(int)`:
++
+Return the work-group index (same as the deprecated `get_id`).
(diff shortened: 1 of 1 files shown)


test_plans/group.asciidoc
```diff
@@ -0,0 +1,71 @@
...
+After the terminates, the results and the return type check is verified.
+
+=== Modification
+
+The test is modified in the following way. The results are gathered for each work-item in every work-group, instead of for one work-item in every work-group. Furthermore, the expected results are initialized to some special value (not zero), to make sure the value is overwritten by the kernel. The following function tests are marked as deprecated:
+
+- `group.get_id().get(int)`
+- `group.get_id(int)`
+- `group.get_global_range().get(int)`
+- `group.get_global_range(int)`
```
gmlueck, line 44: None of these functions exist in the SYCL 2020 specification, not even as deprecated.  I agree that they were all defined in SYCL 1.2.1.

I was not part of the WG when this decision was made, so I don't know if it was an oversight.  Do other members of the WG think that we intentionally removed these functions (vs. deprecating them)?  Or, was this an oversight?

If we intentionally removed them from the spec, then the CTS should not test them at all.

test_plans/group.asciidoc
```diff
@@ -0,0 +1,71 @@
...
+Return a range with the maximum number of work-items in any work group.
+- `group.get_group_linear_id()`
++
+Linear version of `get_group_id`, calculated following Section 3.11.1 of the SYCL 2020 specification.
+- `group.get_local_linear_id()`
++
+Linear version of `get_local_id`, calculated following Section 3.11.1 of the SYCL 2020 specification.
+- `group.get_group_linear_range()`
++
+Linear version of `get_group_range`, calculated following Section 3.11.1 of the SYCL 2020 specification.
```
gmlueck, line 65: The definition in the spec seems clearer to me:

> Return the total number of workgroups in the nd_range.

test_plans/group.asciidoc
```diff
@@ -0,0 +1,71 @@
...
+- `group.get_local_linear_range()`
++
+Linear version of `get_local_range`, calculated following Section 3.11.1 of the SYCL 2020 specification.
```
gmlueck, line 68: Here also, the definition in the spec seems clearer:

> Return the total number of workitems in the work-group.

test_plans/group.asciidoc
```diff
@@ -0,0 +1,71 @@
...
+- `group.leader()`
++
+Returns `true` only for the work-item within each work-group with a local id of zero.
```
gmlueck, line 71: Are the `async_work_group_copy` and `wait_for` member functions covered in a different test plan?  (There were all in SYCL 1.2.1, so maybe they are just inherited from the existing CTS?)

Note, though, that SYCL 1.2.1 had a ` mem_fenc` member function, which does not exist in SYCL 2020.  Again, I'm not sure if this was intentionally removed (vs. deprecated).  If it was intentionally removed, we should remove testing from the CTS.

test_plans/group.asciidoc
nolmoonen: I have updated the description.

test_plans/group.asciidoc
nolmoonen: I have updated the description.

test_plans/group.asciidoc
gmlueck: I think we should write the CTS to the current state of the SYCL 2020 spec.  Since the functions I list above are not in the SYCL 2020 spec, we should remove testing for them from the CTS.

test_plans/group.asciidoc
gmlueck: Same here for `mem_fence`.  This function is not in the SYCL 2020 spec, so we should remove testing from the CTS.

test_plans/group.asciidoc
nolmoonen: Okay that makes sense, I have changed it in the test plan.

test_plans/group.asciidoc
nolmoonen: Fixed.

gmlueck (approved):

gmlueck (approved):

fraggamuffin:
1.2.1 and 2020 has some parts deprecated, should that be part of CTS? remove from CTS if not part of 2020.

fraggamuffin:
LGTM

PR #438, opened 2022-12-14
Add tests for ranged accessors requirement for the entire buffer

vladimirkhashev:
According to [4.7.6.8. Ranged accessors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:accessors.ranged)
> A ranged accessor still creates a requisite for the entire underlying buffer, even for the portions not within the range. For example, if one command writes through a ranged accessor to one region of a buffer and a second command reads through a ranged accessor from a non-overlapping region of the same buffer, the second command must still be scheduled after the first because the requisites for the two commands are on the entire buffer, not on the sub-ranges of the ranged accessors.

This test check this requirement.

### DIFF
--- a/tests/accessor/accessor_requisite_entire_buffer.cpp
+++ b/tests/accessor/accessor_requisite_entire_buffer.cpp
@@ -1,0 +1,67 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Provide checks that ranged accessor still creates a requisite for the entire
+//  underlying buffer
+//
+*******************************************************************************/
+
+#include "../common/disabled_for_test_case.h"
+#include "../common/get_cts_object.h"
+#include "catch2/catch_test_macros.hpp"
+
+#include "accessor_common.h"
+
+namespace accessor_requisite_entire_buffer {
+
+DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp)
+("requisite for the entire underlying buffer for sycl::accessor ",
+ "[accessor]")({
+  auto q = sycl_cts::util::get_cts_object::queue();
+
+  constexpr size_t buffer_size = 10;
+  constexpr size_t offset_size = 7;
+  int data[buffer_size];
+  std::iota(data, (data + buffer_size), 0);
+  {
+    sycl::buffer<int, 1> data_buf(data, sycl::range(buffer_size));
+    // create two commands that uses ranaged accessors that access to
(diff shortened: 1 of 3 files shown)


tests/accessor/accessor_requisite_entire_buffer.cpp
```diff
@@ -0,0 +1,67 @@
...
+//  Provide checks that ranged accessor still creates a requisite for the entire
+//  underlying buffer
+//
+*******************************************************************************/
+
+#include "../common/disabled_for_test_case.h"
+#include "../common/get_cts_object.h"
+#include "catch2/catch_test_macros.hpp"
+
+#include "accessor_common.h"
```
bader, line 14: Some of these header files use features not supported by certain implementations. Typical way to fix this issue is:

```c++
// FIXME: re-enable when <FEATURE X> is implemented
#if !SYCL_CTS_COMPILING_WITH_HIPSYCL && !SYCL_CTS_COMPILING_WITH_COMPUTECPP
// includes like 
// #include "accessor_common.h"
#endif
```

bader (changes requested):

tests/accessor/accessor_requisite_entire_buffer.cpp
vladimirkhashev: Thank you, done

tests/accessor/accessor_requisite_entire_buffer.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "catch2/catch_test_macros.hpp"
+
+// FIXME: re-enable when sycl::accessor is implemented
+#if !SYCL_CTS_COMPILING_WITH_HIPSYCL && !SYCL_CTS_COMPILING_WITH_COMPUTECPP
+#include "accessor_common.h"
+#endif
+
+namespace accessor_requisite_entire_buffer {
+
+DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp)
```
bader, line 21: Could clarify why the test is disabled for these implementations?

bader (approved):

tests/accessor/accessor_requisite_entire_buffer.cpp
vladimirkhashev: Code that doesn't work with these implementations is included via common file, now this code is disabled in common header file for hipSYCL and function that was disabled for ComputeCpp is enabled. So the test is enabled for all implementations now.

ProGTX (approved):

PR #445, opened 2022-12-27
Fix exception constructor tests

vladimirkhashev:
According to [4.13.2. Exception class interface](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:exception.class)

> The string returned by what() is guaranteed to contain what_arg as a substring

This patch changes test to check that what_arg is substring of what() instead of checking their equality.


bader (approved):

keryell (approved):
Good point!

PR #413, opened 2022-11-29
Test plan 4.15.3. Atomic references

bibrakc:
This pull request aims at test plan for 4.15.3. Atomic references.

PR #450, opened 2023-01-04
Test plan for sycl::atomic_ref

vladimirkhashev:
This test plan based on PR: https://github.com/KhronosGroup/SYCL-CTS/pull/413 and comments from @gmlueck.

### DIFF
--- a/test_plans/atomic_ref.asciidoc
+++ b/test_plans/atomic_ref.asciidoc
@@ -54,31 +54,9 @@ All non static tests run on following cases:
 
 Note: all tests run with different values of `DefaultOrder` and `DefaultScope` template parameters, for usupported values tests are skiped. To check if the special values of `DefaultOrder` parameters are supported should use `info::device::atomic_memory_order_capabilities`, `memory_order::relaxed` should be supported on all devices. To check if the special values for `DefaultScope` parameter are supported should use `info::device::atomic_memory_scope_capabilities`.
 
+Note: legal address spaces are not yet defined for atomic_ref used on the host, it's possible that will be decided that atomic_ref isn't allowed in host code at all. It can be tracked in this issue https://gitlab.khronos.org/sycl/Specification/-/issues/637, test plan and tests should be updated if it will be needed according to decision disscussed in the issue.
+
 === SYCL2020 interface
-
-==== Memory order traits
-
-For the following `ReadModifyWriteOrder` template parameter values:
-
-* `memory_order::relaxed`
-
-** Check the following:
-`memory_order_traits<memory_order::relaxed>::read_order` is equal to `memory_order::relaxed`,
-`memory_order_traits<memory_order::relaxed>::write_order` is equal to `memory_order::relaxed`
-
-* `memory_order::acq_rel`
-
-** Check the following:
-`memory_order_traits<memory_order::acq_rel>::read_order` is equal to `memory_order::acquire`,
-`memory_order_traits<memory_order::acq_rel>::write_order` is equal to `memory_order::release`
-
-* `memory_order::seq_cst`
-
-** Check the following:
-`memory_order_traits<memory_order::seq_cst>::read_order` is equal to `memory_order::seq_cst`,
-`memory_order_traits<memory_order::seq_cst>::write_order` is equal to `memory_order::seq_cst`
-
-Check that `read_order` and `write_oder` members of `memory_order_traits<ReadModifyWriteOrder>` are const.
 
 ==== Members
 
@@ -104,9 +82,6 @@ And the following values as `Space` template parameter:
 
 Check the following:
 
-* `default_read_order` member is equal to `memory_order_traits<DefaultOrder>::read_order`,
-* `default_write_order` member is equal to `memory_order_traits<DefaultOrder>::write_order` and
-* `default_read_modify_write_order` member is equal to `DefaultOrder`
 * `default_read_order`, `default_write_order` and `default_read_modify_write_order` members are const
 * `required_alignment` memeber is const and is equal to or grater than `alignof(ReferencedType)`
 * `is_always_lock_free` member is const and is equal to `true` of `false`
@@ -121,6 +96,24 @@ as `atomic_ref::value_type`.
 
 If `T` is `ReferencedType*` check that `atomic_ref::difference_type` is same
 as `ptrdiff_t`.
+
+For different cases of `DefaultOrder` template parameter check the following:
+
+* `memory_order::relaxed`
+** `default_read_order` is `memory_order::relaxed`
+** `default_write_order` is `memory_order::relaxed`
+** `default_read_modify_write_order` is `memory_order::relaxed`
+
+
+* `memory_order::acq_rel`
+** `default_read_order` is `memory_order::acquire`
+** `default_write_order` is `memory_order::release`
+** `default_read_modify_write_order` is `memory_order::acq_rel`
+
+* `memory_order::seq_cst`
+** `default_read_order` is `memory_order::seq_cst`
+** `default_write_order` is `memory_order::seq_cst`
+** `default_read_modify_write_order` is `memory_order::seq_cst`
 
 ==== Common constructors
 
--- a/test_plans/atomic_ref.asciidoc
+++ b/test_plans/atomic_ref.asciidoc
@@ -5,72 +5,99 @@
 
 This is a test plan for the APIs described in SYCL 2020 section 4.15.3. Atomic references
 
-Estimated development time is three days.
-
 == Testing scope
+
+The test plan covers non-atomic behavior, the latter will be covered in a separate test plan
 
 === Device coverage
 
-All of the tests described below are performed only on the default device that
-is selected on the CTS command line.
+All of the tests described below are performed
+
+* on the default device that is selected on the CTS command line
+* on the host side
 
 == Type coverage
+
+All of the tests described below are performed using each of the
+following types as the `ReferencedType`.
+
+In Regular mode:
+
+* `int`
+* `float`
+
+In full conformance mode:
 
 * `int`
 * `unsigned int`
 * `long`
 * `unsigned long`
+* `float`
+
+In addition, if the device has `aspect::atomic64`, the following types are tested:
+
 * `long long`
 * `unsigned long long`
+* `double`
 
-Implementation dependend types:
+Note: for the following types should use an `epsilon` parameter for comparison operators to take into account possible different rounding mode:
 
 * `float`
 * `double`
 
-The member functions of the atomic_ref class may be emulated and may use a different floating-point environment to those defined by `info::device::single_fp_config` and `info::device::double_fp_config`.
-Therefore, the floating-point atomics may use different rounding modes and may have different exception behavior.
-
 == Tests
+
+Note: all test run with different values of `DefaultOrder` and `DefaultScope` template parameters, for usupported values tests are skiped. To check if the special values of `DefaultOrder` parameters are supported should use `info::device::atomic_memory_order_capabilities`, `memory_order::relaxed` should be supported on all devices. To check if the special values for `DefaultScope` parameter are supported should use `info::device::atomic_memory_scope_capabilities`.
 
 === SYCL2020 interface
 
 ==== Members
 
-For the following values as DefaultOrder parameter:
+Create an atomic_ref object in host and device code with the following `DefaultOrder` parameter values:
 
 * `memory_order::relaxed`
+
 Check the following:
-`default_read_order` is `memory_order::relaxed` and
-`default_write_order` is `memory_order::relaxed`
+`default_read_order` member is `memory_order::relaxed` and
+`default_write_order` member is `memory_order::relaxed` and
+`default_read_modify_write_order` member is `memory_order::relaxed`
 
 
 * `memory_order::acq_rel`
+
 Check the following:
-`default_read_order` is `memory_order::acquire` and
-`default_write_order` is `memory_order::release`
+`default_read_order` member is `memory_order::acquire` and
+`default_write_order` member is `memory_order::release` and
+`default_read_modify_write_order` member is `memory_order::acq_rel`
 
 * `memory_order::seq_cst`
+
 Check the following:
-`default_read_order` is `memory_order::seq_cst` and
-`default_write_order` is `memory_order::seq_cst`
+`default_read_order` member is `memory_order::seq_cst` and
+`default_write_order` member is `memory_order::seq_cst` and
+`default_read_modify_write_order` member is `memory_order::seq_cst`
 
-
-Check whether the following members exist: Print their values.
+Check whether the following members exist:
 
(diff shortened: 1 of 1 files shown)


test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,471 @@
...
+
+Additionaly run same tests for `ReferencedType*`.
+
+Note: using T as `ReferencedType` or `ReferencedType*`.
+
+== Tests
+
+All non static tests run on following cases:
+
+* host code with referenced value is simple variable - should be skipped for `address_space::local_space`
```
gmlueck, line 51: We have not yet decided which address spaces are legal when `atomic_ref` is used on the host.  In fact, it's possible that we will decide that `atomic_ref` isn't allowed in host code at all.  This is being discussed in the Khronos issue (https://gitlab.khronos.org/sycl/Specification/-/issues/637).

I think what you have here is fine for now, but can you add a note in this test plan with a pointer to that Khronos issue?  If this is still not resolved when you write the test, please add a comment in the test with a pointer to that issue, so we know to update the test when we resolve it.

vladimirkhashev: Thank you for information! That seems I don't have access to this issue, maybe there is some public tracker that can be added to the test plan for future update?

vladimirkhashev:
Deleted test for memory_order_traits beacuase it's implementation dependent  and tagged as "Exposition only" in spec. Updated test for static member,  now "default_read_order", "default_write_order", "default_read_modify_write_order"  should be compared with corresponding values of memory_order enum.

gmlueck:
> Deleted the test for memory_order_traits beacuase it's implementation dependent and tagged as "Exposition only" in spec.

I'm not sure about this.  @Pennycook was it our intention to say that `sycl::memory_order_traits` is not part of the specification?  If that is the case, I'd like to clarify the spec to make that more obvious.  I think the confusion comes from the comment "Exposition only" in the synopsis.  What do we mean by that exactly?

test_plans/atomic_ref.asciidoc
gmlueck: The Khronos working group is still undecided about using `sycl::atomic_ref` in host code.  We plan to discuss it in the meeting on January 19, so I might have more information at that time.

If you want to make progress on this test, I suggest splitting it into two pieces.  The first part can test `sycl::atomic_ref` in device code.  This part is not controversial, so we can approve that plan whenever it is ready.  Once the working group decides about host code, we can either write a second test plan just for that case or we can update the test plan to include host testing.

Pennycook:
> I'm not sure about this. @Pennycook was it our intention to say that `sycl::memory_order_traits` is not part of the specification?

Yes, it was our intention to say it is not part of the specification.

The class exists only to simplify the definition of these two members of `sycl::atomic_ref` (which _are_ part of the specification):
```c++
  static constexpr memory_order default_read_order = memory_order_traits<DefaultOrder>::read_order;
  static constexpr memory_order default_write_order = memory_order_traits<DefaultOrder>::write_order;
```

An implementation should be free to use some other mechanism to calculate the default order values as long as they are set to the correct value.  For example, an implementation could specialize `sycl::atomic_ref` for every `memory_order` template parameter, or could use an inline constant expression.  But the specification needs some way to show what the default order values are expected to be, and so we introduced `memory_order_traits`.

>  If that is the case, I'd like to clarify the spec to make that more obvious. I think the confusion comes from the comment "Exposition only" in the synopsis. What do we mean by that exactly?

We used the phrase "exposition only" because it appears in the C++ specification to describe some implementation details (e.g.  the definition of [`std::atomic_ref`](http://eel.is/c++draft/atomics.ref.generic.general) uses a private `T*` to describe the behavior of the class).

The definition of "exposition only" is given [here](https://eel.is/c++draft/expos.only.entity), as:

> "Several entities and [typedef-name](https://eel.is/c++draft/dcl.typedef#nt:typedef-name)s defined in [[support]](https://eel.is/c++draft/support) through [[thread]](https://eel.is/c++draft/thread) and [[depr]](https://eel.is/c++draft/depr) are only defined for the purpose of exposition[.](https://eel.is/c++draft/expos.only.entity#1.sentence-1) The declaration of such an entity or [typedef-name](https://eel.is/c++draft/dcl.typedef#nt:typedef-name) is followed by a comment ending in exposition only[.](https://eel.is/c++draft/expos.only.entity#1.sentence-2)"

I agree we should clarify the SYCL specification.  I see a few options:

- Formally import the definition of "exposition only", and correct our usage of it.  Our "exposition only" comments currently come before the definition, but I don't think that's compatible with the definition above.
- Adopt a different convention like `__memory_order_traits__` or `sycl::detail::memory_order_traits`.
- Move `memory_order_traits` inside of `atomic_ref` as a `private` member so it's clearly not visible to developers.

Whatever we do, we should apply the same fix to the [reducer](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#reducer-class) class, which is also defined as "exposition only".

gmlueck:
> Adopt a different convention like `__memory_order_traits__`

We do something like this  elsewhere, generally in conjunction with a comment and / or some explanatory text after the synopsis.  For example the definition of [buffer accessor](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_command_accessors) has:

```
using value_type = // const DataT for read-only accessors, DataT otherwise
      __value_type__;
using accessor_ptr =   // multi_ptr to value_type with target address space,
      __pointer_class__; //   unspecified for access_mode::host_task
```

And then after the synopsis there is [table 55](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.accessors.command.buffer.types) and [table 79](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.accessors.common.types) which describe in words what the type aliases mean.

> Whatever we do, we should apply the same fix to the [reducer](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#reducer-class) class, which is also defined as "exposition only".

Yes, this is unclear.  For example, are the members `value_type`, `binary_operation`, and `dimensions` part of the spec, or are they only exposition?  I was assuming that we intended them to be part of the spec.

What is our intention here about "exposition only"?  For example, I was assuming that `reducer` must be a templated class.  Are we just saying that (some of) the template parameters are unspecified?

Pennycook:
> And then after the synopsis there is [table 55](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.accessors.command.buffer.types) and [table 79](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.accessors.common.types) which describe in words what the type aliases mean.

I see.  I personally prefer C++'s way of doing things here: describing something like `default_read_order` will require a lot of words, and I don't think the tables are an obvious place to look.  But consistency is important, so we should adopt this style if that's what the rest of the spec does.

> Yes, this is unclear. For example, are the members `value_type`, `binary_operation`, and `dimensions` part of the spec, or are they only exposition? I was assuming that we intended them to be part of the spec.

All of the member types, variables and functions are intended to be part of the specification.

> What is our intention here about "exposition only"? For example, I was assuming that `reducer` must be a templated class. Are we just saying that (some of) the template parameters are unspecified?

The intent was to say that some of the template parameters are unspecified _and_ there's no guarantee that a class called `reducer` actually exists.  Developers aren't allowed to create reducers directly, and the specification says to use `auto` or a template parameter everywhere that a reducer is accepted as an argument.

Implementations can implement reducers any way they like, as long as their reducer class(es) provide the expected members.

test_plans/atomic_ref.asciidoc
vladimirkhashev: Got it, removed information concerning of testing `sycl::atomic_ref` on host side from this test plan. When the decision about using `sycl::atomic_ref` in host code have been done we update this test plan or add a new one for test case of using `sycl::atomic_ref` in host code.

gmlueck (approved):

fraggamuffin:
LGTM

keryell (approved):
Thanks!

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+* `unsigned int`
+* `long`
+* `unsigned long`
+* `float`
+
+In addition, if the device has `aspect::atomic64`, the following types are tested:
+
+* `long long`
+* `unsigned long long`
+* `double`
```
AlexeySachkov, line 38: Are those extra types tested in both modes or only in full mode?

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+
+Additionaly run same tests for `ReferencedType*`.
+
+Note: using T as `ReferencedType` or `ReferencedType*`.
+
+== Tests
+
+All non static tests run on following cases:
```
AlexeySachkov, line 46: Do we have a definition of "non static test" somewhere?

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+
+* single_task with referenced value is accessor element - should be skipped for `address_space::local_space`
```
AlexeySachkov, line 48: Suggestion: "single_task" -> "`single_task`" (add backticks to tweak highlighting). The same can be applied for the next bullet.

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+* parallel_for with nd_range parameter and referenced value local_accessor element - should be skipped for `address_space::global_space`
+
+Note: all tests run with different values of `DefaultOrder` and `DefaultScope` template parameters, for unsupported values tests are skiped. To check if the special values of `DefaultOrder` parameters are supported should use `info::device::atomic_memory_order_capabilities`, `memory_order::relaxed` should be supported on all devices. To check if the special values for `DefaultScope` parameter are supported should use `info::device::atomic_memory_scope_capabilities`.
```
AlexeySachkov, line 51: I don't know which formatting standards are used in the repo or whether they exist at all, but it would be great if we had a certain limit on a line width: something like 80 or 120 symbols to improve readability of those documents in console and make it easier to view diff of any changes.

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+
+* `access::address_space::global_space`
+* `access::address_space::local_space`
+* `access::address_space::generic_space`
+
+Check the following:
+
+* `default_read_order`, `default_write_order` and `default_read_modify_write_order` members are const
+* `required_alignment` memeber is const and is equal to or grater than `alignof(ReferencedType)`
+* `is_always_lock_free` member is const and is equal to `true` of `false`
```
AlexeySachkov, line 81: Perhaps this is nit, but `is_always_lock` free is defined as `bool`, it can't accept any other values except for `true` or `false`. Would it be better to explicitly state that we are simply checking the type of that member and not checking its value, because it is implementation-defined?

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+* `default_scope` member is const and is equal to `DefaultScope` template parameter was used when object was created
+
+If `T` is `ReferencedType` check that `value_type` is same as `ReferencedType`.
+
+If `T` is `ReferencedType*` check that `value_type` is same as `ReferencedType*`.
+
+If `T` is `Integral` or `Floating` check that `atomic_ref::difference_type` is same
+as `atomic_ref::value_type`.
```
AlexeySachkov, line 89: If I'm not mistaken, `atomic_ref::difference_type` only exists in partial specialization for pointers

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+** `default_read_modify_write_order` is `memory_order::acq_rel`
+
+* `memory_order::seq_cst`
+** `default_read_order` is `memory_order::seq_cst`
+** `default_write_order` is `memory_order::seq_cst`
+** `default_read_modify_write_order` is `memory_order::seq_cst`
+
+==== Common constructors
+
+Create an `atomic_ref` object using `atomic_ref(T&)` and an other object using copy constructor in device code using the following values as `DefaultOrder` template parameter:
```
AlexeySachkov, line 114: "an other" -> "another"?

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+* `memory_scope::device`
+* `memory_scope::system`
+
+And the following values as `Space` template parameter:
+
+* `access::address_space::global_space`
+* `access::address_space::local_space`
+* `access::address_space::generic_space`
+
+Note: for the member functions tests below that take `memory_order` and `memory_scope` as arguments, the following values of these parameters are used if the `FULL CONFORMANCE` mode is enabled:
```
AlexeySachkov, line 173: "if" -> "only if" ?

test_plans/atomic_ref.asciidoc
```diff
@@ -0,0 +1,480 @@
...
+
+It should be noted that `memory_order::acquire` is used only for read operations, `memory_order::release` for write operations and `memory_order::acq_rel` for read-modify-write operations.
+
+If the `FULL CONFORMANCE` mode is disabled only default values for functions arguments is used which are determined by `atomic_ref` template parameters that which object was created with.
+
+===== Common member functions
+
+`bool is_lock_free()`
+
+Check if the function exists and returns `atomic_ref::is_always_lock_free`.
```
AlexeySachkov, line 199: This doesn't look correct: if `atomic_ref::is_always_lock_free` is `false`, it doesn't mean that this particular instance of `atomic_ref` can't perform lock-free operations.

It seems to me that the check we want to see is that if `atomic_ref::is_always_lock_free` is `true`, then `is_lock_free()` must also return `true`. If `atomic_ref::is_always_lock_free` is `false`, then we can't make any assumptions about result of `is_lock_free()`.

AlexeySachkov (commented):
Mostly questions and minor comments, but I think that a couple of test cases may be invalid in the proposed test plan

test_plans/atomic_ref.asciidoc
gmlueck: The spec also defines it for the specializations for integral types and floating point types.

test_plans/atomic_ref.asciidoc
AlexeySachkov: To me either is fine, I just wanted some clarity here

test_plans/atomic_ref.asciidoc
AlexeySachkov: Note that I'm fine with term "[non] static test" if it is documented somewhere. To me, the most appealing description would be something like "runtime checks"/"compile-time checks" because I see all CTS as runtime tests. One more idea I have is to say something like "test cases which check behavior and not interface": interface could be checked with compile-time checks, whilst behavior can only be checked at runtime.

Issue #406, opened 2022-11-24
Potential bugs in CTS when running with hipSYCL


nilsfriess:
Hello,

I am currently trying to compile and run the tests with hipSYCL to see what is missing in hipSYCL but I came across a few tests that fail not because hipSYCL does not support certain things required by the standard but because the tests are (as far as I can tell) testing things that are in fact not required by the standard. I'd be very happy to contribute to the CTS but I first wanted to discuss these things here, since I might also just misunderstand parts of the SYCL specification. 

Here are a few of the things I found so far:
- Edit: I somehow missed the `SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS` flag, so ignore this point.
    ~Some tests use the `sycl::cl_` types (e.g. the `scalars` test and the `stream` tests) but I think SYCL implementations without an OpenCL backend do not require those types. So maybe these types should be tested in a separate test that can be enabled/disabled depending on the SYCL implementation that is used.~ 
- `marray` has its own test but a few other test make use of `marray` and hence fail if `marray` is not implemented (e.g., the `buffer` tests and the `multi_ptr` tests). We have not (yet) implemented `marray` in hipSYCL and so some tests now fail only because they depend on `marray` but might otherwise work.
- In the `device_selector` tests, there are a bunch of conversion errors (e.g. conversion from `gpu_selector` to `device_selector` fails). These arise since the method `get_cts_object::device` in `tests/common/get_cts_object.h` expects a `sycl::device_selector` as its argument but the SYCL spec does not require `[...]_selector` to be derived from `device_selector` (see Table 14 in Section 4.6.1.1 of the SYCL 2020 Specification). I guess this could easily be fixed by just replacing the `sycl::device_selector` argument by a generic template parameter, for example.
- In the `vector_alias` tests, there are two issues:
    - First, the tests do not take the environment variable `SYCL_CTS_ENABLE_HALF_TESTS` into account (but this is of course easily fixed).
    - Second, the tests expect `char2 == vec<char,2>` but the standard requires `char2 == vec<int8_t, 2>`, see Section 4.14.2.2 of the SYCL 2020 spec, and in general `char != int8_t`.

That's what I found so far, let me know what you think.

bader:
[...]

PR #452, opened 2023-01-05
Add tests for is_property

nolmoonen:
Adds tests for `is_property` and `is_property_of` according to test plan #447.

### DIFF
--- a/ci/hipsycl.filter
+++ b/ci/hipsycl.filter
@@ -21,6 +21,7 @@ math_builtin_api
 multi_ptr
 nd_item
 nd_range
+property
 queue
 range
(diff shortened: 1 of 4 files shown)


nolmoonen:
I added the tests to hipSYCL's CI filter. It seems that hipSYCL cannot compile the contents of `common/type_coverage.h` due to it's usage of `marray`, which hipSYCL does not support. This is related to #406. The failed CI run can be found here: https://github.com/KhronosGroup/SYCL-CTS/actions/runs/3847756725/jobs/6554625024

ProGTX (approved):

PR #454, opened 2023-01-06
Test plan for sycl::atomic_fence

aleksmesh:
Test plan for sycl::atomic_fence [4.15.1. Barriers and fences](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:barriers-fences)


test_plans/barriers_fences.asciidoc
```diff
@@ -0,0 +1,74 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for barriers and fences
```
gmlueck, line 4: We already have a test plan for barriers: https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/group_functions.asciidoc#252-group_barrier

If that part is duplicated, then this plan could cover just `atomic_fence`.

test_plans/barriers_fences.asciidoc
```diff
@@ -0,0 +1,74 @@
...
+=== Device coverage
+
+All of the tests described below are performed only on the default device that
+is selected on the CTS command line.
+
+== Tests
+
+=== SYCL2020 interface
+
+==== Testing `sycl::atomic_fence` function
```
gmlueck, line 20: I was having some difficulty following all the details of the `atomic_fence` test plan, but here are some general comments:

* I think we should concentrate on testing fence operations between work-items in a single kernel.  I think we can wait before adding tests for fence operations between different kernels, different devices, or between the host and device.
* We should avoid using USM allocations in this test because USM is an optional feature, which may not be supported by all implementations.  The test can use `local_accessor` memory to test synchronization between work-items in the same work-group and a buffer `accessor` to test synchronization between work-items in different work-groups.
* We can still test different memory scopes.  `memory_scope::system` and `memory_scope::device` can still be used to synchronize memory operations between work-items in the same kernel.
* Remember that all memory scopes aside from `memory_scope::work_group` are optional.  The test needs to use `info::device::atomic_fence_scope_capabilities` to see which other memory scopes are supported.
* The memory order `memory_order::seq_cst` is also an optional feature.  The test needs to use `info::device::atomic_fence_order_capabilities` to see if it is supported.
* I believe it is only safe to use `atomic_fence` in conjunction with atomic operations like `atomic_ref`.  Perhaps @Pennycook can provide a code snippet to use.

I also have a question for @Pennycook.  I see the spec guarantees that `atomic_fence` supports all memory orders except `memory_order::seq_cst`, which is an optional feature.  However, `atomic_ref` is only guaranteed to support `memory_order::relaxed`, and all the others are optional features.  Is that what we intended?  It seems a little weird that they are different.

One other suggestion ... I believe the existing CTS had a test for the SYCL 1.2.1 function `group::mem_fence`.  The semantics of that function are not exactly the same as `atomic_fence`, but you might be able to get some ideas from looking at that old test.

Pennycook: > I believe it is only safe to use atomic_fence in conjunction with atomic operations like atomic_ref. Perhaps @Pennycook can provide a code snippet to use.

Right.  C++ defines two ways that fences can synchronize with one another [here](http://eel.is/c++draft/atomics.fences), but both require an atomic operation to some memory object _M_.

I don't have an example code ready to go, but I think it should be possible to construct examples by following the wording in the C++ specification.  For example, to test a release fence synchronizing with an acquire fence, as described below:

> A release fence _A_ synchronizes with an acquire fence _B_ if there exist atomic operations _X_ and _Y_, both operating on some atomic object _M_, such that _A_ is sequenced before _X_, _X_ modifies _M_, _Y_ is sequenced before _B_, and _Y_ reads the value written by _X_ or a value written by any side effect in the hypothetical release sequence _X_ would head if it were a release operation.

...we could do something like:

```c++
// work-item 0
data[0] = 42; // Non-atomic write to data[0]
sycl::atomic_fence(sycl::memory_order::release, sycl::memory_scope::work_group); // Fence A
sycl::atomic_ref<int, sycl::memory_order::relaxed, sycl::memory_scope::work_group>(M) = 1; // Atomic X

// work-item 1
int val = sycl::atomic_ref<int, sycl::memory_order::relaxed, sycl::memory_scope::work_group>(M); // Atomic Y
sycl::atomic_fence(sycl::memory_order::acquire, sycl::memory_scope::work_group); // Fence B
x = data[0]; // Non-atomic read of data[0]
if (val == 1) {
  assert(x == 42);
}
```

Because _A_ synchronizes with _B_, the write to data[0] from work-item 0 must be visible to work-item 1 if atomic X executes before atomic Y.

...I'm not sure how to make this into a good test, however, because there's no guarantee in the code above that work-item 0 executes before work-item 1 does.

> I also have a question for @Pennycook. I see the spec guarantees that atomic_fence supports all memory orders except memory_order::seq_cst, which is an optional feature. However, atomic_ref is only guaranteed to support memory_order::relaxed, and all the others are optional features. Is that what we intended? It seems a little weird that they are different.

I agree it is a little weird, and I don't think I fully understand why they're different.

I remember that SYCL 1.2.1's `mem_fence` supported "read" and "write" fences, and that we decided these mapped to acquire and release fences in C++11 terminology.  A relaxed fence doesn't do anything, so any device supporting only relaxed fences effectively wouldn't support memory fences at all.  I think we decided this didn't make sense, because all devices support `group_barrier` and `group_barrier` implies an acquire-release fence at work-group scope; a device that didn't support acquire-release fences wouldn't be able to implement SYCL.

I don't know why this same logic doesn't extend to supporting acquire-release for atomic operations, especially if they can be emulated by generating an `atomic_fence` before or after a relaxed atomic.  But this is another situation where the SYCL behavior is the same as OpenCL, so I'm hoping @bashbaug knows :smile:.

bashbaug: Yes, this is slightly messy, but the minimum capabilities were determined based on the definitions of the older OpenCL 1.x atomic functions (e.g. `atomic_add`) and fence functions (e.g. `read_mem_fence`).  There is a detailed discussion about this here, with a reference to a related internal Khronos issue as well:

https://github.com/KhronosGroup/OpenCL-Docs/issues/354

In short, since one of the design goals of OpenCL 3.0 was to provide an easy path from OpenCL 1.2 to 3.0, we chose to separate the "atomic memory" capabilities from the "atomic fence" capabilities. This is another case where SYCL could choose to do something different and require the same (stronger) memory orders for both atomic_ref and atomic_fence, though it would then be SYCL's responsibility to insert the proper fences when a strong memory order is used with an atomic_ref that is only supported by an atomic_fence.

aleksmesh: In the test plan, I described similar to what @Pennycook is suggesting. But since there's indeed no guarantee that work-item 0 executes before work-item 1 does, I would suggest instead of:
x = data[0]; // Non-atomic read of data[0]
if (val == 1) {
  assert(x == 42);
}
Use loop:
while (1 != (val = sycl::atomic_ref<int, sycl::memory_order::relaxed, sycl::memory_scope::work_group>(M)));
x = data[0]; // Non-atomic read of data[0]
assert(x == 42);

@gmlueck 
Removed the statement about different kernels and devices from the test plan,
Removed the using of USM,
Added checks using accessor and local_accessor,
Added a statement about the requirement to check device capabilities.

test_plans/barriers_fences.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+
+Test should check which `memory_order` values are supported by the device using `info::device::atomic_fence_order_capabilities`.
+
+Next, with writing `work-item`, the value of `sync_flag` should be set to `true`.
+
+In the reading `work-item`, the sequence of operations is as follows:
+
+* store `sync_flag` in a local variable until it reaches the value true
+* call `atomic_fence`
+* check that the `data[i]` value is set to `expected_val`
```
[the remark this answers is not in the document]
Pennycook: > However, I think these tests are not valid because they assume a particular forward-progress guarantee among the work-items. Consider that the reading work-item enters the spin loop on sync_flag before the writer. The code above assumes that the writer will eventually run, but I think there is no such guarantee. @Pennycook has been working on the formalization about forward-progress guarantees, so he might have more to comment here.

I agree that this isn't safe.  It might work on a lot of hardware you run these tests on (because any vectorizing implementations that serialize branches would _probably_ choose to execute the first branch before the second one) but strictly speaking the specification does not guarantee that this code will complete.

The only way to guarantee that a SYCL implementation switches between the work-items in a group is to call `group_barrier`.  But we can't use it here, because the fence implied by `group_barrier` would mean we weren't testing `atomic_fence`.

test_plans/barriers_fences.asciidoc
```diff
@@ -0,0 +1,60 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for barriers and fences
```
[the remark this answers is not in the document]
aleksmesh: Done

test_plans/barriers_fences.asciidoc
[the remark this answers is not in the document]
aleksmesh: @gmlueck Thank you for the clear example to describe the checking algorithm.

I completely agree with you that there is no guarantee that the writing item will run first. But am I correct in guessing that items are executed in parallel execution as described in [4.9.4.2.2. parallel_for invoke](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_parallel_for_invoke) and thus the writing item will anyway execute the write and atomic_fence with atomic_ref will provide no data raising situation?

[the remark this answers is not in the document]
aleksmesh: Thanks for the clarification. I suppose replacing 'while' with 'if' would be correct in this case.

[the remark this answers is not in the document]
gmlueck: Oh, that's a good idea!  Yes, the reader work-item can do something like:

```
bool write_happened = (sync_flag == true);
sycl::atomic_fence(ORDER2, SCOPE);
if (write_happened) {
  if (*data != EXPECTED)
    /* test fails */
}
```

Or even something like:

```
bool write_happened = false;
for (int i = 0;  i < COUNT;  i++) {
  if (sync_flag == true) {
    write_happened = true;
    break;
}
sycl::atomic_fence(ORDER2, SCOPE);
if (write_happened) {
  if (*data != EXPECTED)
    /* test fails */
}
```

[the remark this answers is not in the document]
aleksmesh: Thank you for your proposed variants! I preferred the second of the two.

gmlueck (approved):
Looks good!

fraggamuffin:
LGTM

test_plans/atomic_fence.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+  sycl::atomic_fence(ORDER1, SCOPE);
+  sync_flag = true;
+}
+else {
+  bool write_happened = false;
+  for (int i = 0; i < RETRY_COUNT; i++) {
+    if (sync_flag == true) {
+      write_happened = true;
+      break;
+    }
```
keryell, line 97: Just curious: would it possible for an implementation to execute the `else` branch first and thus never getting the information from the then part?

keryell, line 97: To clarify I have this should be safer:
```suggestion
if (nditem.get_global_linear_id() == 0) {
  *data = EXPECTED;
  sycl::atomic_fence(ORDER1, SCOPE);
  sync_flag = true;
}
if (nditem.get_global_linear_id() != 0) {
  bool write_happened = false;
  for (int i = 0; i < RETRY_COUNT; i++) {
    if (sync_flag == true) {
      write_happened = true;
      break;
    }
```

keryell (changes requested):
I have the feeling the tests should be written in a safer way.

test_plans/atomic_fence.asciidoc
gmlueck: Answering your question in the other comment:

> Just curious: would it possible for an implementation to execute the else branch first and thus never getting the information from the then part?

Yes, it is absolutely possible for some work-items to execute the "else" part before the first work-item executes the "then" part.  We could not think of any portable way to avoid this race, and the best we could think of was the RETRY_COUNT loop.  Note, however, that the test does not fail when some work-items execute the "else" part first.

Regarding your suggested code change above, I'm not opposed, but I don't see how it improves anything.  Aren't the two code snippets equivalent?  The return value of `nditem.get_global_linear_id()` is the same every time it is called from any particular work-item.

aleksmesh: So, since there is no difference, can I keep the version suggested in the test plan or should I replace 'if else' with 'if if'??

keryell: Still thinking to it...  I have not convinced myself with this because of potentially random order of work-item execution... 
I have the feeling that the only way to have the right forward progress guarantee is to have a group barrier which does a... memory fence! :-|
Perhaps having a group barrier with relaxed memory order and having the memory fence in the example having stronger guarantees to be checked?

aleksmesh: @keryell  Do I understand correctly, that you're suggesting to use atomic_fence together with group_barrier, to get a stronger guarantee of verification?
If so, then I think that if the tests run successfully, we won't know definitely if the mem fences used in group_barrier didn't masks the atomic_fence's incorrect functionality.

The checks for group_barrier are described here https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/group_functions.asciidoc#252-group_barrier.

test_plans/atomic_fence.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+* `ORDER1` is `sycl::memory_order::release` and `ORDER2` is `sycl::memory_order::acquire`.
+* `ORDER1` and `ORDER2` are both `sycl::memory_order::acq_rel`
+* `ORDER1` and `ORDER2` are both `sycl::memory_order::seq_cst`
+
+===== Testing `atomic_fence` across `work-items` in different `work-groups`
+
+This test allocates memory in a buffer `accessor` that is shared by all `work-items` in the `kernel`.
+One `work-item` writes a value to this memory and all other `work-items` in the `kernel` read the memory,
+making sure it has the expected value. Because this test requires `sycl::memory_scope::device`,
+it is only run if the device supports that memory scope.
```
keryell, line 79: Since we still have spent quite some time in reviews and during the SYCL meeting today devising how we can design the following test, you should describe also here why the test is designed this way, what it can test and what it cannot test.
Then you can copy paste the text as a comment into https://github.com/KhronosGroup/SYCL-CTS/pull/464 :-)
At least we do not have found a convincing way to have a better test.

keryell (changes requested):
Looks good except the subtle test needs some explanations on what, why, how.

test_plans/atomic_fence.asciidoc
aleksmesh: The check algorithms proposed in the test plan check that using a synchronizing variable of atomic_ref type in combination with atomic_fence function provides no data racing and the specified order of instruction execution.

Also, in the algorithm the atomic_fence function check will be performed only if the leader writes a value to the sync_flag variable before all other items complete the loop with the check of sync_flag value.

It does not give a 100% guarantee that the check will be performed, but while there is no guarantee that the loop with the condition while (sync_flag != true); will always complete, a safe and less strict check algorithm has been chosen.

The same explanation I added to the test plan as an additional section.

fraggamuffin:
wait for ronan, else this could be ready

fraggamuffin:
actually, this is OK, LGTM

bader:
> actually, this is OK, LGTM

@keryell, I would prefer you to confirm that your latest request for changes is satisfied before merging.

keryell (approved):
Thanks for the clarification.

PR #458, opened 2023-01-11
Add tests for stream class

nolmoonen:
Adds tests for the stream class according to #433.

### DIFF
--- a/tests/stream/stream_api.cpp
+++ b/tests/stream/stream_api.cpp
@@ -3,7 +3,7 @@
 //  SYCL 2020 Conformance Test Suite
 //
 //  Copyright (c) 2017-2022 Codeplay Software LTD. All Rights Reserved.
-//  Copyright (c) 2022 The Khronos Group Inc.
+//  Copyright (c) 2022-2023 The Khronos Group Inc.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
--- a/tests/stream/stream_api.cpp
+++ b/tests/stream/stream_api.cpp
@@ -154,21 +154,52 @@ class TEST_NAME : public util::test_base {
           sycl::stream os(2048, 80, cgh);
 
           /** check get_size()
-          */
+           */
+#ifdef SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS
           {
             auto size = os.get_size();
-            check_return_type<size_t>(log, size,
-                                      "sycl::context::get_size()");
+            check_return_type<size_t>(log, size, "sycl::stream::get_size()");
           }
+#endif  // SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS
 
-          /** check get_max_statement_size()
-          */
+          /** check size()
+           */
+#ifndef SYCL_CTS_COMPILING_WITH_DPCPP
+          {
+            auto size = os.size();
+            check_return_type<size_t>(log, size, "sycl::stream::size()");
+            CHECK(noexcept(os.size()));
+          }
+#else
+          WARN(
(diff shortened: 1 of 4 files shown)


tests/stream/stream_api.cpp
```diff
@@ -146,29 +143,59 @@ class TEST_NAME : public util::test_base {
...
+          /** check size()
+           */
+#ifndef SYCL_CTS_COMPILING_WITH_DPCPP
+          {
+            auto size = os.size();
+            check_return_type<size_t>(log, size, "sycl::stream::size()");
+            CHECK(noexcept(os.size()));
+          }
+#else
+          WARN(
```
bader, line 170: NOTE: We are about to add a new macro - SKIP. If https://github.com/KhronosGroup/SYCL-CTS/pull/469 goes first, these WARN should be replaced with SKIP.

tests/stream/stream_api.cpp
```diff
@@ -146,29 +143,59 @@ class TEST_NAME : public util::test_base {
...
+          {
+            auto maxStatementSize = os.get_work_item_buffer_size();
             check_return_type<size_t>(
                 log, maxStatementSize,
-                "sycl::context::get_max_statement_size()");
+                "sycl::stream::get_work_item_buffer_size()");
           }
+#else
+          WARN(
+              "DPCPP does not define sycl::stream:get_work_item_buffer_size(). "
```
bader, line 196:
```suggestion
              "DPCPP does not define sycl::stream::get_work_item_buffer_size(). "
```

tests/stream/stream_constructors.cpp
```diff
@@ -80,6 +80,27 @@ class TEST_NAME : public util::test_base {
...
 
+      /** check (size_t, size_t, sycl::handler&, sycl::property_list&)
+       * constructor
+       */
+      {
+        queue.submit([&](sycl::handler &handler) {
+          sycl::property_list property_list{};
+          sycl::stream os(bufferSize, maxStatementSize, handler, property_list);
+
+          if (os.get_size() != bufferSize) {
```
bader, line 91: `get_size` is deprecated. Please, use `size` instead.

This must be applied to all other stream test cases.

bader (changes requested):

tests/stream/stream_constructors.cpp
nolmoonen: Fixed, also by replacing `get_max_statement_size` with `get_work_item_buffer_size`.

tests/stream/stream_api.cpp
nolmoonen: Fixed.

tests/stream/stream_api.cpp
nolmoonen: Acknowledged, that's a nice addition.

tests/stream/stream_constructors.cpp
nolmoonen: An unfortunate side effect is that the constructor test must now be disabled for DPCPP, as it does not yet implement the SYCL2020 API here.

tests/stream/stream_api_fp16.cpp
```diff
@@ -30,7 +30,7 @@ using namespace sycl_cts;
 class test_kernel;
```
bader, line 30: Should we update the copyright header in this file as well?

bader (approved):

keryell (approved):
Thanks.

tests/stream/stream_api_fp16.cpp
nolmoonen: Right, I forgot about that. Since `stream_api_fp64.cpp` had no license header, I added the one from `docs/README.adoc` with year 2023.

PR #459, opened 2023-01-11
Fix device_global tests

steffenlarsen:
This commit makes the following changes to device_global related tests:
* Moves arrow_operator_overloaded struct out of no_def_cnstr.
* Fix overloads of assign and are_equal for arrays and add device_global overload.
* Change SYCL_EXT_ONEAPI_PROPERTY_LIST to SYCL_EXT_ONEAPI_PROPERTIES after change to the corresponding extension.
* Add experimental to the namespace of properties and device_global.
* Fix failure reporting.
* Fix use of templates on `value_operations`.
* Prevent the use of different contexts expecting the same device_global.
* Fix uses of compile-time properties after changes to the extension.
* Additional small bugfixes.

### DIFF
--- a/tests/extension/oneapi_device_global/device_global_exceptions.cpp
+++ b/tests/extension/oneapi_device_global/device_global_exceptions.cpp
@@ -19,7 +19,7 @@ using namespace sycl_cts;
 using namespace sycl_cts::util;
 using namespace device_global_common_functions;
 
-#if defined(SYCL_EXT_ONEAPI_PROPERTY_LIST) && \
+#if defined(SYCL_EXT_ONEAPI_PROPERTIES) && \
     defined(SYCL_EXT_ONEAPI_DEVICE_GLOBAL)
 namespace oneapi = sycl::ext::oneapi;
 
@@ -31,13 +31,13 @@ namespace oneapi = sycl::ext::oneapi;
  * otherwise
  */
 inline bool is_errc_invalid(const sycl::exception& e) {
-  return e.code() != sycl::errc::invalid;
+  return e.code() == sycl::errc::invalid;
 }
 
 namespace handler_copy_exceptions {
 // Creating instance with default constructor
 template <typename T>
-oneapi::device_global<T> dev_global;
+oneapi::experimental::device_global<T> dev_global;
 
 /** @brief The function tests that .copy() member function overload throws an
  * exception with error code equal to errc::invalid if attempt to write beyond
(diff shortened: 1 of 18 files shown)


ProGTX (approved):

tests/extension/oneapi_device_global/device_global_exceptions.cpp
```diff
@@ -31,13 +31,13 @@ namespace oneapi = sycl::ext::oneapi;
  * otherwise
  */
 inline bool is_errc_invalid(const sycl::exception& e) {
-  return e.code() != sycl::errc::invalid;
+  return e.code() == sycl::errc::invalid;
```
bader, line 34: 😮

tests/extension/oneapi_device_global/device_global_queue_copy.cpp
```diff
@@ -162,36 +162,37 @@ void run_test_copy_from_device_global(util::logger& log,
...
+  }
 
   auto event3 = queue.copy(dev_global<T>, dst_data3,
                                 sizeof(T) / sizeof(element_type), 0, depEvents);
   event3.wait();
   // Reverse order is used to increase probability of data race.
   // Array in gens[0] is already copied after event2,
   // so it's not copied here to not to rewrite result for previous case.
-  for (size_t i = numEvents - 1; i > 1; --i) {
+  for (size_t i = numEvents - 1; i + 1 > 0; --i) {
```
bader, line 195: `i` type is unsigned (size_t is unsigned right?) i.e. `i >= 0`, so `i + 1` is always greater than 0. Isn't this an infinite loop now?

```suggestion
  for (size_t i = numEvents - 1; i - 1 > 0; --i) {
```
?

bader (approved):
LGTM, except suspicious code in tests/extension/oneapi_device_global/device_global_queue_copy.cpp and formatting suggestions.
Please, also apply clang-format patch https://github.com/KhronosGroup/SYCL-CTS/actions/runs/3904532277#summary-10611336793.

tests/extension/oneapi_device_global/device_global_queue_copy.cpp
steffenlarsen: Good catch! That's an oopsie on my part. I have changed it to reverse the range `[numEvents..1]` and instead subtract 1 in the subscript operator. Additionally I have removed the comment about the copy above as it is no longer valid.

bader: Thanks. There were two such bugs in this file. You fixed the first one but the second one is still there. Could you update the second one too? Can we refactor the code to avoid code/bugs duplication?

steffenlarsen: Ah, thanks for pointing this out. I have fixed the other issue as well. It may be possible to create a common function for this, but the copy operation uses device global as source in one and dst in the other, so it would be heavily templated and I am not sure it is worth it. If you think it is worth it, I will go refactor it.

PR #462, opened 2023-01-11
Fix reference builtins for mixing vectors and scalars

steffenlarsen:
The min, max, and clamp reference builtins currently assume that the type inside the vector argument and the scalar type of the other argument are the same. However, this is not necessarily the case when using the char vector alias and the char scalar type, as the former uses `int8_t` which may be defined as `signed char` or `unsigned char` which are considered different types from `char` without specified signedness.
This commit relaxes the reference implementations to allow different scalar type to the element type in the passed vector, requiring it is implicitly convertible to the element type.


ProGTX (approved):

Melirius:
Hmm, this is a tickling bomb, essentially. I'll file an issue

PR #448, opened 2023-01-02
Add test plan for context class and device class

nolmoonen:
Adds a test plan for the missing information descriptors of the context class as described in Section 4.6.3. of the SYCL 2020 specification and of the device class as described in Section 4.6.4. of the SYCL 2020 specification.

PR #463, opened 2023-01-11
Add tests for context class and device class

nolmoonen:
Implements test plan #448. Note: edits are mostly due to moving the test to the new test framework, and the formatting changes that implies.


ProGTX (approved):

keryell (approved):
Thanks!

PR #465, opened 2023-01-12
Create test plan for sycl::queue constructors

vladimirkhashev:
Test plan to cover gaps in queue constructors tests https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/queue/queue_constructors.cpp.

### DIFF
--- a/test_plans/queue_constructors.asciidoc
+++ b/test_plans/queue_constructors.asciidoc
@@ -28,7 +28,7 @@ For following constructors use created `syclContext`, selected device or `cts_se
                const async_handler& asyncHandler,
                const property_list& propList = {})`
 
-Note: test with and without both properties `enable_profiling` and `in_order`.
+Note: test without properties.
 
 === Check property `in_order`
 
@@ -53,11 +53,14 @@ Use property `in_order` with following constructors also using selected device, 
                const async_handler& asyncHandler,
                const property_list& propList = {})`
 
-If `device::has(sycl::aspect::usm_host_allocations)` is `false` this part of test is skipped.
-Submit two `single_task` to queue - first with long loop and second that changes a `usm` variable with host allocation.
+If `device::has(sycl::aspect::usm_device_allocations)` is `false` this part of test is skipped.
+Submit two `single_task` to queue - first with long loop and second that changes a `usm` variable with device allocation.
 In the first task check that after loop the variable is unchanged to verify that `in_order` property adds the requirement that the SYCL queue provides in-order semantics.
 
 === Check constructors with both properties
+
+If `device::has(sycl::aspect::queue_profiling)` is `false` these tests is skipped.
+The same condition should also be added to existing tests for queue constructors with `enable_profiling` property.
 
 Use `propList = {in_order, enable_profiling}` with following constructors also using selected device, corresponding context, `cts_selector`, `cts_async_handler` and check that resulting queue member finctions `has_property<enable_profiling>()`, `has_property<in_order>()` and `is_in_order()` return `true`:
 
--- a/test_plans/queue_constructors.asciidoc
+++ b/test_plans/queue_constructors.asciidoc
@@ -1,0 +1,82 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl::queue constructors
+
+This is a test plan for the for new constructors of class `sycl::queue` described in SYCL 2020.
+
+== Testing scope
+
+Tests intend to cover not covered constructors from 4.6.5.1. Queue interface that are not covered in https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/queue/queue_constructors.cpp
+
+=== Device coverage
+
+Device coverage does not change. All of the tests described below are performed only on the default device that is selected on the CTS command line.
+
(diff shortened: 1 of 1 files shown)


test_plans/queue_constructors.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+
+* `queue(const context& syclContext, const DeviceSelector& deviceSelector, const property_list& propList = {})`
+* `queue(const context& syclContext, const DeviceSelector& deviceSelector,
+               const async_handler& asyncHandler, const property_list& propList = {})`
+* `queue(const context& syclContext, const device& syclDevice, const property_list& propList = {})`
+* `queue(const context& syclContext, const device& syclDevice,
+               const async_handler& asyncHandler,
+               const property_list& propList = {})`
+
+Note: test with and without both properties `enable_profiling` and `in_order`.
```
gmlueck, line 31: There is no need to duplicate this test with these properties.  They will have no effect on the error condition that you are testing.

test_plans/queue_constructors.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+* `queue(const context& syclContext, const DeviceSelector& deviceSelector,
+               const async_handler& asyncHandler, const property_list& propList = {})`
+* `queue(const context& syclContext, const device& syclDevice, const property_list& propList = {})`
+* `queue(const context& syclContext, const device& syclDevice,
+               const async_handler& asyncHandler,
+               const property_list& propList = {})`
+
+If `device::has(sycl::aspect::usm_host_allocations)` is `false` this part of test is skipped.
+Submit two `single_task` to queue - first with long loop and second that changes a `usm` variable with host allocation.
+In the first task check that after loop the variable is unchanged to verify that `in_order` property adds the requirement that the SYCL queue provides in-order semantics.
```
gmlueck, line 58: It would be better to use `aspect::usm_device_allocations` and `malloc_device` for this test.  Implementations are more likely to support USM device allocations than USM host allocations.  I think the test works equally well with either type of USM.

test_plans/queue_constructors.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+
+=== Check constructors with both properties
+
+Use `propList = {in_order, enable_profiling}` with following constructors also using selected device, corresponding context, `cts_selector`, `cts_async_handler` and check that resulting queue member finctions `has_property<enable_profiling>()`, `has_property<in_order>()` and `is_in_order()` return `true`:
```
gmlueck, line 62: The `enable_profiling` property is an optional feature.  The test needs to check if the selected device has `aspect::queue_profiling` before constructing a queue with this property.  If the selected device doesn't have this aspect, this test should be skipped.

test_plans/queue_constructors.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+* `queue(const device& syclDevice, const property_list& propList = {})`
+* `queue(const device& syclDevice, const async_handler& asyncHandler,
+               const property_list& propList = {})`
+* `queue(const context& syclContext, const DeviceSelector& deviceSelector, const property_list& propList = {})`
+* `queue(const context& syclContext, const DeviceSelector& deviceSelector,
+               const async_handler& asyncHandler, const property_list& propList = {})`
+* `queue(const context& syclContext, const device& syclDevice, const property_list& propList = {})`
+* `queue(const context& syclContext, const device& syclDevice,
+               const async_handler& asyncHandler,
+               const property_list& propList = {})`
```
gmlueck, line 81: Do we already have a test of the profiling APIs?  If we do not already test this, can we add a simple test like this:

* Check if the selected device has `aspect::queue_profiling`.  Skip this test if it does not.
* Construct a `queue` from the selected device with the `enable_profiling` property.
* Submit a simple command with `single_task`, saving the returned `event`.
* Wait for the task to complete.
* Get the following timestamp values from the `event`:
   * `submit = event::get_profiling_info<info::event_profiling::command_submit>()`
   * `start = event::get_profiling_info<info::event_profiling::command_start>()`
   * `end = event::get_profiling_info<info::event_profiling::command_end>()`
* Check that:
   * `end >= start`
   * `start >= submit`

test_plans/queue_constructors.asciidoc
vladimirkhashev: Removed cases with properties.

test_plans/queue_constructors.asciidoc
vladimirkhashev: Changed to `usm_device_allocations` and device allocation.

test_plans/queue_constructors.asciidoc
vladimirkhashev: Added condition `device::has(sycl::aspect::queue_profiling)` for test case.

test_plans/queue_constructors.asciidoc
vladimirkhashev: It is tested here https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/event/event.cpp#L450

gmlueck (approved):

PR #468, opened 2023-01-12
Add tests for sub group

nolmoonen:
Implements test plan #436.

### DIFF
--- a/tests/sub_group/sub_group_api.cpp
+++ b/tests/sub_group/sub_group_api.cpp
@@ -40,16 +40,17 @@ enum struct member_function : int {
   local_linear_range,
   leader,
   group_group_linear_id,  // part of group, not part of sub-group like above
-  method_count            // defines size, should be last
+  member_function_count   // defines size, should be last
 };
 
-static constexpr auto method_cnt = to_integral(member_function::method_count);
+static constexpr auto member_function_cnt =
+    to_integral(member_function::member_function_count);
 
 struct kernel_name;
 
-/** Return a linear index for a global work-item index and a method. */
-static size_t get_idx(size_t id, member_function method) {
-  return id * method_cnt + to_integral(method);
+/** Return a linear index for a global work-item index and a member function. */
+static size_t get_idx(size_t id, member_function member_func) {
+  return id * member_function_cnt + to_integral(member_func);
 }
 
 /**
@@ -113,15 +114,6 @@ size_t to_size_t(sycl::range<1> returned_value) {
   return returned_value[0];
 }
 
-// ComputeCPP cannot implicitly convert sycl::id to size_t
-#ifdef SYCL_CTS_COMPILING_WITH_COMPUTECPP
-/** sycl::id cannot be implicitly converted to size_t. */
-template <>
-size_t to_size_t(sycl::id<1> returned_value) {
-  return returned_value[0];
-}
-#endif
-
 /**
  * Registers the result of calling a member function.
  * The returned value \p returned_value is converted to \p size_t and stored
@@ -138,8 +130,14 @@ void register_member_function_no_type_check(size_t linear_execution_range,
   }
   // calculate the unique index of this work-item and member function
   const size_t idx = get_idx(linear_global_id, member);
+#ifdef SYCL_CTS_COMPILING_WITH_COMPUTECPP
+  // ComputeCPP cannot implicitly convert sycl::id to size_t.
+  // Set the result to some bogus value so that the workaround is removed.
+  accessor[idx] = std::numeric_limits<size_t>::max() - 1;
+#else
   // returned value of the function is converted to size_t for uniform storage
   accessor[idx] = to_size_t(returned_value);
+#endif
 }
 
 /**
@@ -158,6 +156,12 @@ void register_member_function(size_t linear_execution_range,
 }
 
 TEST_CASE("sub-group api", "[sub_group]") {
+#ifdef SYCL_CTS_COMPILING_WITH_COMPUTECPP
+  WARN(
+      "ComputeCpp cannot implicitly convert sycl::id to size_t. "
+      "The test that requires this functionality is set to fail.");
+#endif
+
   sycl::device device = sycl_cts::util::get_cts_object::device();
   sycl::queue queue = sycl_cts::util::get_cts_object::queue();
   const sycl::range<3> local_range{2, 3, 5};
@@ -178,7 +182,7 @@ TEST_CASE("sub-group api", "[sub_group]") {
 
   // allocate space for results to be gathered,
   // initialize arrays to special value to ensure a write happens
-  const size_t result_count = linear_execution_range * method_cnt;
+  const size_t result_count = linear_execution_range * member_function_cnt;
   constexpr size_t empty = std::numeric_limits<size_t>::max();
   std::vector<size_t> results(result_count, empty);
   {
@@ -237,7 +241,7 @@ TEST_CASE("sub-group api", "[sub_group]") {
                                            member_function::leader,
                                            accessor_value, sub_group.leader());
             // group.get_group_linear_id()
-            // method belongs to sycl::group, no return type check needed
+            // member function belongs to sycl::group, no type check needed
             register_member_function_no_type_check(
                 linear_execution_range, lgid,
                 member_function::group_group_linear_id, accessor_value,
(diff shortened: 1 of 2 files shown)


ProGTX (approved):

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+                  [=](size_t size) { return size == first_max_local_range; }));
+
+  // work-items are assigned to sub-groups in an implementation-defined way,
+  // and sub_group.get_group_id() cannot be used to identity work-items in the
+  // execution range. the below checks verify properties without depending on
+  // implementation details. note therefore that "global index" as used below
+  // is calculated rather than expected based on the work-item id
+
+  // check that all work-items in the same work-group have the same number
+  // of sub-groups. use REQUIRE as sub_group_counts is used by next checks
```
keryell, line 263:
```suggestion
  // Work-items are assigned to sub-groups in an implementation-defined way,
  // and sub_group.get_group_id() cannot be used to identity work-items in the
  // execution range. The checks below verify properties without depending on
  // implementation details. Note therefore that "global index" as used below
  // is calculated rather than expected based on the work-item id.

  // Check that all work-items in the same work-group have the same number
  // of sub-groups. Use REQUIRE as sub_group_counts is used by next checks
```

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+  // ---------------------------------------------------------------------------
+
+  // for each work-group, the number of sub-groups it contains
+  std::vector<size_t> sub_group_counts(work_group_count, special_val);
+  // populate and check sub_group_counts by iterating over the work-items
+  for (size_t i = 0; i < linear_execution_range; i++) {
+    const size_t wg_id = results[get_idx(i, method::group_group_linear_id)];
```
keryell, line 270: Just curious, what would happen if `wg_it` returned by the implementation were wrong? Would we have some UB everywhere?

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+    const size_t local_sg_id = results[get_idx(i, method::group_id)];
+    // #sub-groups before this work-group + sub-group index in this work-group
+    //  = global sub-group index
+    const size_t global_sg_id = work_group_offsets[wg_id] + local_sg_id;
+    const size_t item_id = results[get_idx(i, method::local_id)];
+    // #items before this sub-group + local item index in this sub-group
+    //  = global item index
+    const size_t global_item_id = sub_group_offsets[global_sg_id] + item_id;
+    item_seen[global_item_id] = true;
```
keryell, line 364: Same here. What happens if the id are completely wrong?

keryell (commented):
I am a little bit lost to understand all the bits here. Of course, this part of the spec is complex too.
Pedantic remark: since this is C++, `method` should be renamed `member function` or `member_function` according to the context.

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+std::string format_range(const sycl::range<3>& range) {
+  std::ostringstream ss;
+  ss << "(" << range[0] << ", " << range[1] << ", " << range[2] << ")";
+  return ss.str();
+}
+
+TEST_CASE("sub-group api", "[sub_group]") {
+  sycl::device device = sycl_cts::util::get_cts_object::device();
+  sycl::queue queue = sycl_cts::util::get_cts_object::queue();
+  const sycl::range<3> local_range{2, 2, 2};
```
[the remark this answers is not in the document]
nolmoonen: Changed it to `{2, 3, 5}`.

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+      cgh.parallel_for<kernel_name>(
+          execution_range, [=](sycl::nd_item<3> item) {
+            const size_t lgid = item.get_global_linear_id();
+            const sycl::sub_group sub_group = item.get_sub_group();
+
+            {  // get_group_id()
+              const size_t idx = get_idx(lgid, method::group_id);
+              accessor_value[idx] = sub_group.get_group_id()[0];
+              ASSERT_RETURN_TYPE(sycl::id<1>, sub_group.get_group_id(),
+                                 "sycl::sub_group.get_group_id()");
```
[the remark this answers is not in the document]
nolmoonen: Done as requested with `register_member_function`.

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+static size_t get_work_group_size(const sycl::device& device) {
+#if !(defined(SYCL_CTS_COMPILING_WITH_HIPSYCL) || \
+      defined(SYCL_CTS_COMPILING_WITH_COMPUTECPP))
+  const sycl::id<3> max_work_item_sizes =
+      device.get_info<sycl::info::device::max_work_item_sizes<3>>();
+#else
+  const sycl::id<3> max_work_item_sizes{1, 1, 1};
+  WARN(
+      "Implementation does not define device.get_info<max_work_item_sizes>. "
+      "Using work-group of size (1, 1, 1).");
```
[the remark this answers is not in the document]
nolmoonen: Why do we care about testing `sub_group`'s API when `info::device::max_work_item_sizes` is not defined? Because the sub-group may be developed independently of the information descriptors and therefore it would be nice to still be able to run the sub-group test when information descriptors are not implemented. Also, it is relatively little effort to put this workaround in place.

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+  for (size_t i = 0; i < linear_execution_range; i++) {
+    INFO("index: " << i);
+    const auto get = [=](method method) -> size_t {
+      return results[get_idx(i, method)];
+    };
+
+    // group_id, group_linear_id
+    size_t idx_group_id = get(method::group_id);
+    CHECK((idx_group_id == get(method::group_linear_id)));
+    CHECK((idx_group_id < get(method::group_linear_range)));
```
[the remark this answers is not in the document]
nolmoonen: Added some more comments.

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,368 @@
...
+  // execution range. the below checks verify properties without depending on
+  // implementation details. note therefore that "global index" as used below
+  // is calculated rather than expected based on the work-item id
+
+  // check that all work-items in the same work-group have the same number
+  // of sub-groups. use REQUIRE as sub_group_counts is used by next checks
+  // ---------------------------------------------------------------------------
+
+  // for each work-group, the number of sub-groups it contains
+  std::vector<size_t> sub_group_counts(work_group_count, special_val);
```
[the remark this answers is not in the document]
nolmoonen: Changed to `empty`.

tests/sub_group/sub_group_api.cpp
nolmoonen: Good point! For each of these instances, I have added a `REQUIRE` to check the bounds.

tests/sub_group/sub_group_api.cpp
nolmoonen: Same solution.

fraggamuffin:
LGTM

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,430 @@
...
+  // ---------------------------------------------------------------------------
+
+  // for each sub-group, the number of work-items it contains
+  std::vector<size_t> sub_group_sizes(sub_group_count, empty);
+  // populate and check sub_group_sizes by iterating over the work-items
+  for (size_t i = 0; i < linear_execution_range; i++) {
+    const size_t wg_id =
+        results[get_idx(i, member_function::group_group_linear_id)];
+    REQUIRE(wg_id < work_group_count);  // bounds check
+    const size_t local_sg_id = results[get_idx(i, member_function::group_id)];
```
keryell, line 376: You could simplify the code everywhere with
```suggestion
    const size_t local_sg_id = result(i, member_function::group_id);
```
by defining the right `result` function or lambda.

keryell (changes requested):
Thanks.
Almost there.

tests/sub_group/sub_group_api.cpp
```diff
@@ -0,0 +1,430 @@
...
+  local_range,
+  group_range,
+  max_local_range,
+  group_linear_id,
+  local_linear_id,
+  group_linear_range,
+  local_linear_range,
+  leader,
+  group_group_linear_id,  // part of group, not part of sub-group like above
+  method_count            // defines size, should be last
```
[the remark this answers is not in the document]
nolmoonen: Right my bad, fixed properly this time.

tests/sub_group/sub_group_api.cpp
nolmoonen: Nice suggestion, I've applied it.

fraggamuffin:
wait for Ronan

keryell (approved):
Thanks!

PR #469, opened 2023-01-12
Bump Catch2 to d548be26 for new SKIP macro

psalz:
I'm glad to announce that support for the long-awaited `SKIP` macro [has been merged](https://github.com/catchorg/Catch2/pull/2360) into Catch2!

This PR bumps the revision of the Catch2 submodule to current `devel` and adjusts *most* tests to use the new macro instead of manually printing a message and returning. The main advantage of using the macro is that test cases are reported explicitly as "skipped" instead of "failed" or "passed". This in turn will allow us to handle skipped tests separately in conformance report generation, which will become important once we start to decide on which test cases are required and which ones are optional for conformance.

The `SKIP` macro behaves essentially like a manual `FAIL`, in that it is the last thing executed in a test case. For this reason some of the legacy tests with more complex control flow still print manual skip messages, as they require some restructuring first (or rather, they should be converted to Catch2).

I've also updated the developer documentation to recommend the use of the `SKIP` macro when testing optional features.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -345,6 +345,12 @@ where `<impls...>` is a comma-separated list of `hipSYCL`, `ComputeCpp` and/or `
 
 NOTE: #**TODO:** Custom matchers.#
 
+=== Testing Optional Features
+
+The CTS may include tests that cannot be executed in all circumstances.
+Examples of such tests include tests for optional features, tests that depend on certain device capabilities, tests that require multiple devices as well as tests for vendor extensions.
+In such scenarios, Catch2's `SKIP` macro should be used to explicitly report a test case as skipped.
+
 === Best Practices
 
 Here is a list of best practices for writing test cases.
--- a/tests/cuda_interop/cuda_backend_avail.cpp
+++ b/tests/cuda_interop/cuda_backend_avail.cpp
@@ -27,6 +27,6 @@ TEST_CASE("CUDA backend availability test") {
   INFO("Checking sycl::backend::cuda is available");
   REQUIRE(q.get_backend() == sycl::backend::cuda);
 #else
-  INFO("The test is skipped because CUDA back-end is not supported");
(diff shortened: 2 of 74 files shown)


ProGTX (approved):
Congratulations on getting `SKIP` merged into Catch2, this looks like a nice cleanup!

tests/kernel_bundle/has_kernel_bundle_throws_diff_dev_from_one_platform_dev.cpp
```diff
@@ -36,8 +36,7 @@ class TEST_NAME : public sycl_cts::util::test_base {
     for (const auto &platform : platforms) {
       auto devices{platform.get_devices()};
       if (devices.size() <= 1) {
-        log.note("Platform skipped due to one device was found");
-        continue;
+        SKIP("Requires at least two devices");
       }
```
bader, line 40: Shouldn't we change the logic of this test?
Currently this test validates all platforms with 2+ devices. If I understand the sematic of SKIP correctly, new version will validate platforms with 2+ devices before the first one with one (or less?) device.

psalz: Whoops - good catch, thanks. I made those changes somewhat mechanically and I did not see that loop there. Reverted.

---

As an aside, the logic here is not ideal. We shouldn't be performing tests on random devices that happen to be available. I think it's time we start to think about how conformance should work in regards to features that require multiple devices. I'll open an issue.

bader (approved):
Looks good to me.
I merged a few PR, which introduced merge conflicts. Could you check if we can apply SKIP macro to recently committed code?

psalz:
Rebased onto current SYCL-2020 branch and resolved conflicts. There were no new opportunities for converting to SKIP.

PR #472, opened 2023-01-13
Add test plan for common reference semantics

nolmoonen:
Adds a test plan for the common reference semantics as described in Section 4.5.2. of the SYCL 2020 specification.

### DIFF
--- a/test_plans/reference_semantics.asciidoc
+++ b/test_plans/reference_semantics.asciidoc
@@ -24,8 +24,7 @@ The common reference semantics are defined for the following runtime classes in 
 
 Of these, the following classes are legal kernel parameter types (Section 4.12.4.): `accessor` (when templated with `target::device` or the deprecated `target::global_buffer`, `target::constant_buffer`, or `target::local`), `local_accessor`, `unsampled_image_accessor` (when templated with `image_target::device`), `sampled_image_accessor` (when templated with `image_target::device`), `stream`.
 
-
-The below tests are performed on the host for the following types:
+The following "host types" are identified:
 
 - `accessor<int, 1, access_mode::read_write, target::host_task>`
 - `buffer<int>`
@@ -44,7 +43,7 @@ The below tests are performed on the host for the following types:
 - `sampled_image`
 - `unsampled_image`
 
-The below tests are performed in a kernel for the following types:
+The following "kernel types" are identified:
 
 - `accessor<int, 1, access_mode::read_write, target::device>`
 - `local_accessor<int, 0>`
@@ -52,30 +51,31 @@ The below tests are performed in a kernel for the following types:
 - `sampled_image_accessor<int, 1>`
 - `stream`
 
-=== Type traits
+=== Host application tests
 
-For each runtime class `T`, check on the host that the following statements evaluate to `true`.
+These tests are run from a host application for all host types and kernel types.
 
-- `std::is_copy_constructible_v<T>`.
-- `std::is_copy_assignable_v<T>`
-- `std::is_destructible_v<T>`
-- `std::is_move_constructible_v<T>`
-- `std::is_move_assignable_v<T>`
+==== Equality via copy, symmetry
 
-=== Equality comparable and standard hash
-
-For each runtime class `T`, perform the below tests.
-
-==== Equality via copy and symmetry
-
-This test is repeated for the copy constructor and for copy assignment.
+This test is repeated for the copy constructor and for copy assignment. For class `T`:
 
 1. Create an instance `t0` of `T`.
 2. Create a copy of `t0`, `t1`.
 3. Check that `t0 == t1` and `t1 == t0`.
 4. Check that `std::hash<T>{}(t0) == std::hash<T>{}(t1)`.
 
+==== Non-equality via copy, symmetry
+
+This test is not executed for: `device_image`, `device` if only one device is available and `platform` if only one platform is available. For class `T`:
+
+1. Create an instance `t0` of `T`.
+2. Create an instance `t1` of `T`, which is distinct from `t0`.
+3. Check that `t0 != t1` and `t1 != t0`.
+4. Check that `std::hash<T>{}(t0) != std::hash<T>{}(t1)`.
+
 ==== Reflexivity
+
+For class `T`:
 
 1. Create an instance `t0` of `T`.
 2. Check that `t0 == t0`.
@@ -83,10 +83,24 @@ This test is repeated for the copy constructor and for copy assignment.
 
 ==== Transitivity
 
-This test is repeated for the copy constructor and for copy assignment.
+This test is repeated for the copy constructor and for copy assignment. For class `T`:
 
 1. Create an instance `t0` of `T`.
 2. Create a copy of `t0`, `t1`.
 3. Create a copy of `t1`, `t2`.
 4. Check that `t0 == t2` and `t2 == t0`.
 5. Check that `std::hash<T>{}(t0) == std::hash<T>{}(t2)`.
+
+=== Shared tests
+
+These tests are run both from a host application for all host types and kernel types, and from a kernel function for all kernel types.
+
+==== Type traits
+
+For class `T`, check that the following statements evaluate to `true`.
+
+- `std::is_copy_constructible_v<T>`.
+- `std::is_copy_assignable_v<T>`
+- `std::is_destructible_v<T>`
+- `std::is_move_constructible_v<T>`
+- `std::is_move_assignable_v<T>`
--- a/test_plans/reference_semantics.asciidoc
+++ b/test_plans/reference_semantics.asciidoc
@@ -93,7 +93,9 @@ This test is repeated for the copy constructor and for copy assignment. For clas
 
 === Shared tests
 
-These tests are run both from a host application for all host types and kernel types, and from a kernel function for all kernel types. Some of the below tests require mutation of an instance `T`. This is not always possible. When it is not possible, do not execute those tests.
+The tests in this section tests are run from a host application for all host types and kernel types, and from a kernel function for all kernel types.
+
+The tests below assert equality without the equality operator, such that they may also be executed in a kernel function. The "state" denotes the results of calling member functions of an object.
 
 ==== Type traits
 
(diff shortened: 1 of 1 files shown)


test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+- `sampled_image`
+- `unsampled_image`
+
+The below tests are performed in a kernel for the following types:
+
+- `accessor<int, 1, access_mode::read_write, target::device>`
+- `local_accessor<int, 0>`
+- `unsampled_image_accessor<int, 1, access_mode::read>`
+- `sampled_image_accessor<int, 1>`
+- `stream`
```
gmlueck, line 53: All of these types can be used to construct an object on the host too.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+
+- `std::is_copy_constructible_v<T>`.
+- `std::is_copy_assignable_v<T>`
+- `std::is_destructible_v<T>`
+- `std::is_move_constructible_v<T>`
+- `std::is_move_assignable_v<T>`
+
+=== Equality comparable and standard hash
+
+For each runtime class `T`, perform the below tests.
```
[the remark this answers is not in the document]
nolmoonen: Good observation, I misread that part. I have changed the plan to only execute the equality and `std::hash` checks on a host application.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+
+==== Transitivity
+
+This test is repeated for the copy constructor and for copy assignment.
+
+1. Create an instance `t0` of `T`.
+2. Create a copy of `t0`, `t1`.
+3. Create a copy of `t1`, `t2`.
+4. Check that `t0 == t2` and `t2 == t0`.
+5. Check that `std::hash<T>{}(t0) == std::hash<T>{}(t2)`.
```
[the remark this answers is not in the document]
nolmoonen: Good point, I have added this test. I agree with you about `device_image`.

test_plans/reference_semantics.asciidoc
nolmoonen: I have re-written the tests: there are "host types" and "kernel types". The tests for the equality operators and `std::hash` are done in a host application for host and kernel types. The tests for type traits are done in a host application for host and kernel types and in a kernel function for kernel types only.

I do not think it is useful to call the copy/move constructor/assignment functions from a kernel function, since the return value cannot be checked as equality is not guaranteed to be available.

psalz:
You are probably aware of this but there already is some precedent for testing common reference semantics [here](https://github.com/KhronosGroup/SYCL-CTS/blob/9a2072f5861e19936ea3a59c82a90c300b599ee9/tests/common/common_by_reference.h) and [here](https://github.com/KhronosGroup/SYCL-CTS/blob/9a2072f5861e19936ea3a59c82a90c300b599ee9/tests/event/event.cpp#L68) (and possibly elsewhere, I'm not sure). Maybe there's a way of salvage/unify some of these approaches.

gmlueck (approved):

nolmoonen:
> You are probably aware of this but there already is some precedent for testing common reference semantics [here](https://github.com/KhronosGroup/SYCL-CTS/blob/9a2072f5861e19936ea3a59c82a90c300b599ee9/tests/common/common_by_reference.h) and [here](https://github.com/KhronosGroup/SYCL-CTS/blob/9a2072f5861e19936ea3a59c82a90c300b599ee9/tests/event/event.cpp#L68) (and possibly elsewhere, I'm not sure). Maybe there's a way to salvage/unify some of these approaches.

@psalz I was only aware of the tests in `event.cpp`, which is a subset of the proposed test plan. Next to the common header you mention (which only seems to be used for `kernel_id`), there is also reference semantics checking in `accessor_legacy/accessor_constructors_(buffer|image|local)_utility.h`. I have updated the plan to include the existing functionality and to specify some implementation details to consolidate the existing tests. 

The change consists of checking copy and move on a host application and in a kernel function by storing the results of member functions. It also includes tests for the modification of instances to check if these propagate through copies.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,148 @@
...
+4. Check that the member functions of `t1` return the same values as stored in Step 2.
+
+==== Const-correctness
+Check that a `const` copy of an instance is not affected by the mutations to the instance that is copied.
+
+1. Create an instance `t0` of `T`.
+2. Make a `const` copy of `t0`, `t1`.
+3. Store the properties of `t1` by calling its member functions.
+4. Mutate `t0`.
+5. Check that the member functions of `t1` return the same values as stored in Step 2.
```
gmlueck, line 124:
```suggestion
5. Check that the member functions of `t1` return the same values as stored in Step 3.
```

I assume "step 2" was just a typo.

This test looks wrong to me, though.  [Section 4.5.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics) makes this guarantee about objects with common reference semantics:

> Any instance of T that is constructed as a copy of another instance, via either the copy constructor or copy assignment operator, **must behave as-if it were the original instance and as-if any action performed on it were also performed on the original instance** ...

There is no exception here for `const` objects.  In addition, the spec guarantees that `t0 == t1`, so it would be very weird if `t0` and `t1` didn't return the same values from their member functions.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,148 @@
...
+
+==== Mutation propagation
+Check that mutations are applied both to the instance and to the copy of the instance.
+
+1. Create an instance `t0` of `T`.
+2. Make a copy of `t0`, `t1`.
+3. Mutate `t1`.
```
gmlueck, line 131: What does "mutate" mean?  Step 4 below says "store properties", so I thought at first that "mutate" means "change the properties".  However, SYCL does not provide any way to change the properties of an object once it is created.  The common properties interface in [section 4.5.4.1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_properties_interface) has only functions to query the properties (`has_property` and `get_property`).  There is no `set_property`.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,148 @@
...
+
+1. Create an instance `t0` of `T`.
+2. Create a copy of `t0`, `t1`.
+3. Create a copy of `t1`, `t2`.
+4. Check that `t0 == t2` and `t2 == t0`.
+5. Check that `std::hash<T>{}(t0) == std::hash<T>{}(t2)`.
+
+=== Shared tests
+
+These tests are run both from a host application for all host types and kernel types, and from a kernel function for all kernel types. Some of the below tests require mutation of an instance `T`. This is not always possible. When it is not possible, do not execute those tests.
```
gmlueck, line 96: It would be nice to clarify when mutation is not possible.  I presume this will be the case for SYCL types that do not provide any "mutation" member functions?  If so, can we list the SYCL types that will be tested here?

See also my comment below asking what "mutation" means.

test_plans/reference_semantics.asciidoc
nolmoonen: Yes, indeed a typo. Okay I see, I have changed this test to check that a "mutation" is also applied to the `const` copy.

Interestingly, the test in `common_by_reference.h` is:
```c++
  const auto &const_obj = a;
  storage.store(const_obj);
  mutator(a, mutation::const_correctness); // changes instance a
  // failure if const_obj not the same as what was stored before the mutation
  util::run_check(log, mkInfo("Const-correcntess."), [&]{
    if (!storage.check(const_obj)) { 
      FAIL(log, testName + " ignores const correctness (storage::check())");
    }
  });
```

Unless I am missing something, this test is wrong.

test_plans/reference_semantics.asciidoc
nolmoonen: I replaced "properties" with "state" and clarified that the state of an object is the result of calling its member functions. The purpose of this is to check equality without relying on the presence of the equals operator.

test_plans/reference_semantics.asciidoc
nolmoonen: Essentially, it is a change to an object that can be "detected" in the copy of that object. I have added a better explanation and a list of types for which I think this holds.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,191 @@
...
+* detection: read from the accessor.
+- `accessor<int, 1, access_mode::read_write, target::device>` (in a kernel function):
+* mutation: write to the accessor.
+* detection: read from the accessor.
+- `buffer<int>` by obtaining an accessor and writing to the accessor (in the host application):
+* mutation: create an accessor, write to the accessor.
+* detection: create an accessor, read from the accessor.
+- `event` by creating an event with multiple dependent events (in the host application):
+* mutation: resolve an event on the wait list.
+* detection: query the size of the wait list.
```
gmlueck, line 146: I'm  not sure how this will work.  There is no SYCL API to create an event with dependent events.  I suppose you could submit a `single_task` that has dependent events.

However, I'm not sure how you would "resolve an event on the wait list".  There's no way to manually resolve an event.

test_plans/reference_semantics.asciidoc
gmlueck: Yes, that does seem wrong.  I wonder if that test failed before?

In any case, the new test plan you have written looks correct w.r.t. `const` copies.

nolmoonen: No, the `common_by_reference.h` header is only used by `kernel_id`, which passes a no-op `mutator`.

I must add that it is (apparently) hard to get the tests for the reference semantics right with regards to the propagation of changes. However, I feel that this is quite a crucial feature of these semantics.

test_plans/reference_semantics.asciidoc
```diff
@@ -0,0 +1,191 @@
...
+- `host_accessor` (in the host application):
+* mutation: write to the accessor.
+* detection: read from the accessor.
+- `host_sampled_image_accessor<int, 1>` (in the host application):
+* mutation: write to the accessor.
+* detection: read from the accessor.
```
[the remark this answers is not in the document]
nolmoonen: Okay, I have corrected these.

gmlueck (approved):

test_plans/reference_semantics.asciidoc
TApplencourt: You should be able to do a single task with a  `while !signaled[0] {}` and then use the host to modify the  `signaled` variable. Allow you to trigger an event on the wait list.

Issue #365, opened 2022-07-20
Add support for reduced feature sets.


bader:
A number of SYCL-CTS tests require "full" feature set. For instance, multi_ptr test requires implementation to support generic address space and id test requires implementation to support unnamed SYCL kernel function feature.

fraggamuffin:
should we have flag to switch between full and reduce? 
is reduced feature set real? Need feedback from CP.

alycm:
Below is the current state of reduced feature set support on ComputeCpp to provide some context.  Answering based on the using the "experimental" ComputeCpp compiler, which is where the primary SYCL 2020 support is being developed.

**Un-named SYCL kernel functions**.   ComputeCpp can support unnamed lambas when using ComputeCpp as both the host and device compiler.  For using a separate host compiler it depends on whether you get lucky with how that compiler's ABI encodes lambda names.  I suspect that the program will compile without error, but then it _might_ not run.  There is currently no flag to force this support off.

**Address space mode**.  ComputeCpp is always compiling programs using the inferred address space mode, even if the target hardware can support the generic address space.  If a program violates the inferred address space rules then ComputeCpp will fail to compile the program with some kind of "Address space conflict" error.  This will happen for very few programs.  There is no flag for this because ComputeCpp currently only supports the single mode.

There is a third point in the reduced feature set, but it's taking me a little longer to understand that part.

PR #473, opened 2023-01-13
Add test plan for feature sets

nolmoonen:
Adds a test plan to modify the CTS to take into account the full and reduced feature sets of Appendix B of the SYCL 2020 specification. It addresses #365.

### DIFF
--- a/test_plans/reduced_feature_set.asciidoc
+++ b/test_plans/reduced_feature_set.asciidoc
@@ -42,5 +42,5 @@ A simple test is constructed that ODR-uses a variable with static storage durati
 
 Repeat the following test both for `static const int value = <value>;` and `static constexpr int value = <value>;`.
 
-- Launch a kernel that takes the address of `value` (which ODR-uses the variable) and stores the result.
-- Check that the result does not evaluate to `nullptr`.
+- Launch a kernel that takes the address of `value` (which ODR-uses the variable) and locally stores the result. Dereference the pointer in a way that taking the address of `value` is not optimized out. For instance, by passing `0` in a buffer and using that to index the pointer. Store the result in a buffer.
+- On the host application, check that the result is equal to `value`.
--- a/test_plans/reduced_feature_set.asciidoc
+++ b/test_plans/reduced_feature_set.asciidoc
@@ -17,7 +17,7 @@ Currently, three full features are defined. For each of these features, the requ
 
 == Un-named SYCL kernel functions
 === Required behavior
-Every kernel launch has a kernel name, e.g. `parallel_for<kernel_name>`.
+Every kernel launch that does not accept `sycl::kernel` as argument, must have a kernel name, e.g. `parallel_for<kernel_name>`.
 
 === Exceptions
 The existing tests for `sycl::handler` in `handler/handler_invoke_api.cpp` are modified such that they are also run without a kernel name, if `SYCL_CTS_ENABLE_FEATURE_SET_FULL` is defined.
@@ -40,6 +40,7 @@ No variables with static storage duration may be ODR-used inside a device functi
 === Exceptions
 A simple test is constructed that ODR-uses a variable with static storage duration and produces some side effect. The test is guarded with `SYCL_CTS_ENABLE_FEATURE_SET_FULL`.
 
-- Declare `static const int value = <value>;`
+Repeat the following test both for `static const int value = <value>;` and `static constexpr int value = <value>;`.
+
 - Launch a kernel that takes the address of `value` (which ODR-uses the variable) and stores the result.
 - Check that the result does not evaluate to `nullptr`.
--- a/test_plans/reduced_feature_set.asciidoc
+++ b/test_plans/reduced_feature_set.asciidoc
@@ -1,0 +1,45 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for reduced and full feature set
+
+This is a test plan for the full and reduced feature sets as described in Appendix B.1. and Appendix B.2. of the SYCL 2020 specification.
+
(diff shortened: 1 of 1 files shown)


test_plans/reduced_feature_set.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+
+In all existing tests, the full features are removed and replaced with a workaround. Since the CTS was not written with the feature sets in mind, a "best effort" is made to remove the full features in the CTS. New additions should not use full features unless guarded with `SYCL_CTS_ENABLE_FEATURE_SET_FULL`.
+
+An exception to this are the tests that explicitly test full features. For example, when launching a kernel to test `sycl::id` 's interface, the kernel should be launched with a kernel name instead of using the full feature to not specify a name. However, the test that verifies that the full feature that allows for kernels to be launched without a kernel name works correctly, should still exist and be guarded with `SYCL_CTS_ENABLE_FEATURE_SET_FULL`.
+
+Currently, three full features are defined. For each of these features, the required behavior as a result from removing full features is stated, in addition to the tests which are the exceptions (or the proposal for a simple test if none exist).
+
+== Un-named SYCL kernel functions
+=== Required behavior
+Every kernel launch has a kernel name, e.g. `parallel_for<kernel_name>`.
```
gmlueck, line 20: I think this is a good goal.  I wonder how much work it will be, though.  Adding a unique kernel name might be difficult in tests where the kernel is defined as part of a complex template.

test_plans/reduced_feature_set.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+
+=== Exceptions
+The existing tests for `sycl::handler` in `handler/handler_invoke_api.cpp` are modified such that they are also run without a kernel name, if `SYCL_CTS_ENABLE_FEATURE_SET_FULL` is defined.
+
+== Generic address space mode
+=== Required behavior
+Section 5.9.4. of the SYCL 2020 specification defines several rules for the deduction of the address space of pointer types and reference types. It is illegal when following these rules results in a pointer value being assigned to a pointer variable that addresses a different addressing space. Tests that depend on such behavior should be rewritten.
```
gmlueck, line 27: How will we decide if any of the existing tests violate the inferred address space rules?  Can we run ComputeCpp and use it's error checking to find violations?

test_plans/reduced_feature_set.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+== ODR-usage of static declarations
+=== Required behavior
+No variables with static storage duration may be ODR-used inside a device function.
+
+=== Exceptions
+A simple test is constructed that ODR-uses a variable with static storage duration and produces some side effect. The test is guarded with `SYCL_CTS_ENABLE_FEATURE_SET_FULL`.
+
+- Declare `static const int value = <value>;`
+- Launch a kernel that takes the address of `value` (which ODR-uses the variable) and stores the result.
+- Check that the result does not evaluate to `nullptr`.
```
gmlueck, line 45: If we return the address back to the host (e.g. via a buffer), then the host can check that the address is the same as `&value`.

Shall we also do this test for a `static constexpr` variable?

test_plans/reduced_feature_set.asciidoc
nolmoonen: From what I can tell, a good chunk of the tests already use a kernel name. Creating a name for a kernel that is part of a complex template will not be that hard since kernel names themselves can also be templatized e.g. [buffer_constructors_common.h:L67](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/buffer/buffer_constructors_common.h#L67).

test_plans/reduced_feature_set.asciidoc
nolmoonen: The problem with that is that ComputeCpp's CI filter contains quite a lot of tests, which might be disabled for different reasons. Of course, compilation succeeds on the enabled tests, so these do not contain disallowed behavior according to ComputeCpp's error checking. From a manual check, I so far only found `multi_ptr`'s conversion and assignment checks to break this rule.

This is exactly why I put "best effort" in the test plan: I suspect there is no automatic way to check whether disallowed behavior is present in the CTS. However, if the assumption is made that it's there, the reduced feature set cannot be tested. The plan suggest to assume (after making a best effort to remove any) it is not there, and fix on a case-by-case basis if problems do occur.

test_plans/reduced_feature_set.asciidoc
nolmoonen: > If we return the address back to the host (e.g. via a buffer), then the host can check that the address is the same as `&value`.

That was my assumption as well, but I tested it with DPCPP and found that these are not equal. I think that makes sense, since the address spaces of the device and host are not (necessarily) the same. I too find it unfortunate that the value cannot be checked. I could not come up with a test that 1. uses a variable with static storage duration, 2. the kernel ODR-uses the variable, and 3. the result can be checked.

> Shall we also do this test for a `static constexpr` variable?

I have added it to the test plan.

gmlueck: Maybe this is a better test about ODR-using a static variable that is `static const` or `static constexpr`:

* Define a kernel that takes the address of `value` and stores the result.
* From the same kernel, read the stored address, dereference the pointer, and make sure the value returned is `<value>`.  Structure the code such that the compiler cannot determine statically that the stored address is really the address of `value`.

I think our intent is that it is legal to take the address of the variable from device code and using it within device code, but we do not make any guarantee about using the address on the host.

nolmoonen: Okay, that seems to work. I have updated the test plan.

gmlueck (approved):

PR #475, opened 2023-01-18
Add test for sycl::is_compatible

aleksmesh:
Add test for sycl::is_compatible according to [2.1. Functions is_compatible](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/kernel_bundle.asciidoc#21-finctions-is_compatible)

### DIFF
--- a/tests/kernel_bundle/sycl_is_compatible.cpp
+++ b/tests/kernel_bundle/sycl_is_compatible.cpp
@@ -41,6 +41,9 @@ TEST_CASE(
   const std::vector<sycl::kernel_id> builtinKernelIds =
       device.get_info<sycl::info::device::built_in_kernel_ids>();
 
+  if (builtinKernelIds.empty())
+    SKIP("No built-in kernels available for selected device.");
+
   CHECK(sycl::is_compatible(builtinKernelIds, device));
 }
 
@@ -63,8 +66,9 @@ TEST_CASE(
       }
     }
   }
-  if (!builtinKernelIds.empty())
-    CHECK(!sycl::is_compatible(builtinKernelIds, device));
+  if (builtinKernelIds.empty())
+    SKIP("No built-in kernels available for non-tested devices.");
+  CHECK(!sycl::is_compatible(builtinKernelIds, device));
 }
 
 template <typename KernelDescriptorT>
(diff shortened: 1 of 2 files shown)


ProGTX (approved):

fraggamuffin:
LGTM

tests/kernel_bundle/sycl_is_compatible.cpp
```diff
@@ -0,0 +1,240 @@
...
+}
+
+TEST_CASE(
+    "Check is_compatible(const std::vector<kernel_id>, const device) for "
+    "built-in kernel ids for the selected device",
+    "[kernel_bundle]") {
+  const sycl::device device = sycl_cts::util::get_cts_object::device();
+  const std::vector<sycl::kernel_id> builtinKernelIds =
+      device.get_info<sycl::info::device::built_in_kernel_ids>();
+
```
bader, line 43: Shouldn't we check that `builtinKernelIds` is not empty? 
Otherwise, this becomes a duplicate of the previous test.

tests/kernel_bundle/sycl_is_compatible.cpp
```diff
@@ -0,0 +1,240 @@
...
+      if (d != device) {
+        std::vector<sycl::kernel_id> newKernelIds =
+            d.get_info<sycl::info::device::built_in_kernel_ids>();
+        builtinKernelIds.insert(builtinKernelIds.end(), newKernelIds.begin(),
+                                newKernelIds.end());
+      }
+    }
+  }
+  if (!builtinKernelIds.empty())
+    CHECK(!sycl::is_compatible(builtinKernelIds, device));
```
bader, line 67: This seems to be a good case for SKIP macro.
```suggestion
  if (builtinKernelIds.empty())
    SKIP("No built-in kernels available for non-tested devices.");
  CHECK(!sycl::is_compatible(builtinKernelIds, device));
```

bader (approved):
LGTM. Just a couple of suggestions.

tests/kernel_bundle/sycl_is_compatible.cpp
aleksmesh: You are right, added check with SKIP.

tests/kernel_bundle/sycl_is_compatible.cpp
aleksmesh: Changed to SKIP.

PR #481, opened 2023-01-19
Fix USM queue shortcuts check for bool

sergey-semenov:
The copied values are filled using a customized version of iota to allow for bool type. This ends up filling every element after the 0th with ones, but they are then compared with a sequence of integers. Fix this by casting the expected value to the type being checked.


sergey-semenov:
Could someone please take a look?

keryell (approved):
Thanks.

PR #482, opened 2023-01-19
Fix array initialization in multi_ptr for no default ctor

steffenlarsen:
The `multi_ptr` tests may run with `no_def_cnstr` but will fail to compile due to the existence of uninitialized arrays in certain tests requiring the default constructor. This commit adds an initializer for these arrays, skipping initialization for `def_cnstr` as it has no constructor accepting an integer.

### DIFF
--- a/tests/multi_ptr/multi_ptr_common.h
+++ b/tests/multi_ptr/multi_ptr_common.h
@@ -163,6 +163,26 @@ struct check_pointer {
   }
 };
 
+/** @brief Initializes all elements in a container with 0
+ */
+template <typename T, int N, int... Vals>
+struct zero_init_container : zero_init_container<T, N - 1, 0, Vals...> {};
+template <typename T, int... Vals>
+struct zero_init_container<T, 0, Vals...> {
+  using elem_type = typename T::value_type;
+  static inline constexpr T value = {elem_type{Vals}...};
+};
+
+/** @brief Initializes an std::array with either 0's or using the default
+ * constructor
+ */
+template <typename T, int N>
+struct init_array : zero_init_container<std::array<T, N>, N> {};
+template <int N>
+struct init_array<user_def_types::def_cnstr, N> {
+  static inline constexpr std::array<user_def_types::def_cnstr, N> value;
+};
+
 }  // namespace multi_ptr_common
 
 #endif  // SYCL_CTS_TEST_MULTI_PTR_MULTI_PTR_COMMON_H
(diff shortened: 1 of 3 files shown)


tests/multi_ptr/multi_ptr_common.h
```diff
@@ -163,6 +163,26 @@ struct check_pointer {
   }
 };
 
+/** @brief Initializes all elements in a container with 0
```
keryell, line 166: This code is unclear to me. Where does the 0 come?
It is more efficient and clearer nowadays to use https://en.cppreference.com/w/cpp/utility/integer_sequence
Compilation time in O(1) instead of O(N).

tests/multi_ptr/multi_ptr_common.h
```diff
@@ -163,6 +163,26 @@ struct check_pointer {
...
+ */
+template <typename T, int N, int... Vals>
+struct zero_init_container : zero_init_container<T, N - 1, 0, Vals...> {};
+template <typename T, int... Vals>
+struct zero_init_container<T, 0, Vals...> {
+  using elem_type = typename T::value_type;
+  static inline constexpr T value = {elem_type{Vals}...};
+};
+
+/** @brief Initializes an std::array with either 0's or using the default
```
keryell, line 176: I do not understand the relation between the comment and the following.

steffenlarsen: `user_def_types::def_cnstr` is the only class in use that does not have a constructor taking an integer, so we make an exception to the 0-initialization. We could also do it the other way around, i.e. only do explicit initialization of `no_def_cnstr` given it is the only one without a default constructor.

tests/multi_ptr/multi_ptr_common.h
steffenlarsen: The 0's are expanded by recursively repeating 0 in the inheritance. For example, for `N == 2` we would have the inheritance chain

```
zero_init_container<T, 2>
  : zero_init_container<T, 1, 0>
  : zero_init_container<T, 0, 0, 0>
```
where `zero_init_container<T, 0, 0, 0>::value` ends up being `{T::value_type{0},T::value_type{0}}`.

We could use `std::integer_sequence` instead, but I don't see how we could construct with 0 easier using that. Granted, default initialization was the previous behavior so arguably any integer value would do.

ProGTX (approved):

tests/multi_ptr/multi_ptr_common.h
keryell: OK, why not.

keryell (approved):
Thanks for the clarification.

PR #483, opened 2023-01-23
Update tests for queue constructors

AmirIpma:
Add missing tests for queue constructors according to test plan https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/queue_constructors.asciidoc

### DIFF
--- a/tests/queue/queue_constructors.cpp
+++ b/tests/queue/queue_constructors.cpp
@@ -29,6 +29,10 @@ TEST_CASE("Check queue default constructor and destructor", "[queue]") {
   sycl::queue queue;
 }
 TEST_CASE("Check queue (property_list) constructor", "[queue]") {
+  auto device = util::get_cts_object::device();
+  if (!device.has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
+
   sycl::queue queue(
       sycl::property_list{sycl::property::queue::enable_profiling()});
 
@@ -41,6 +45,9 @@ TEST_CASE("Check queue (async_handler) constructor", "[queue]") {
 }
 
 TEST_CASE("Check queue (async_handler, property_list) constructor", "[queue]") {
+  auto device = util::get_cts_object::device();
+  if (!device.has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
   cts_async_handler asyncHandler;
   sycl::queue queue(asyncHandler, {sycl::property::queue::enable_profiling()});
 
@@ -57,6 +64,8 @@ TEST_CASE("Check queue (device_selector) constructor", "[queue]") {
 TEST_CASE("Check queue (device_selector, property_list) constructor",
           "[queue]") {
   cts_selector selector;
+  if (!sycl::device(selector).has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
   sycl::queue queue(selector, {sycl::property::queue::enable_profiling()});
 
   CHECK(queue.get_device() == sycl::device(selector));
@@ -77,6 +86,8 @@ TEST_CASE(
     "Check queue (device_selector, async_handler, property_list) constructor",
     "[queue]") {
   cts_selector selector;
+  if (!sycl::device(selector).has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
   cts_async_handler asyncHandler;
   sycl::queue queue(selector, asyncHandler,
                     {sycl::property::queue::enable_profiling()});
@@ -95,6 +106,8 @@ TEST_CASE("Check queue (device) constructor", "[queue]") {
 
 TEST_CASE("Check queue (device, property_list) constructor", "[queue]") {
   sycl::device device = util::get_cts_object::device();
+  if (!device.has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
   sycl::queue queue(device, {sycl::property::queue::enable_profiling()});
 
   CHECK(queue.get_device() == device);
@@ -113,6 +126,8 @@ TEST_CASE("Check queue (device, async_handler) constructor", "[queue]") {
 TEST_CASE("Check queue (device, async_handler, property_list) constructor",
           "[queue]") {
   sycl::device device = util::get_cts_object::device();
+  if (!device.has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
   cts_async_handler asyncHandler;
   sycl::queue queue(device, asyncHandler,
                     {sycl::property::queue::enable_profiling()});
@@ -133,6 +148,8 @@ TEST_CASE("Check queue (context, device_selector) constructor", "[queue]") {
 TEST_CASE("Check queue (context, device_selector, property_list) constructor",
           "[queue]") {
   cts_selector selector;
+  if (!sycl::device(selector).has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
   auto context = util::get_cts_object::context(selector);
   sycl::queue queue(context, selector,
                     {sycl::property::queue::enable_profiling()});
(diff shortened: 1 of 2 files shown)


tests/queue/queue_constructors.cpp
```diff
@@ -19,425 +19,286 @@
...
+TEST_CASE("Check queue hashing", "[queue]") {
+  cts_selector selector;
+  auto queueA = util::get_cts_object::queue(selector);
+  sycl::queue queueB(queueA);
+  std::hash<sycl::queue> hasher;
+
+  CHECK(hasher(queueA) == hasher(queueB));
+}
+
+// FIXME: re-enable when sycl::errc is implemented in computcpp
```
bader, line 258:
```suggestion
// FIXME: re-enable when sycl::errc is implemented in computecpp
```

tests/queue/queue_properties.cpp
```diff
@@ -20,56 +20,243 @@
...
+  REQUIRE(queue.has_property<sycl::property::queue::in_order>());
+
+  auto prop = queue.get_property<sycl::property::queue::in_order>();
+  check_return_type<sycl::property::queue::in_order>(
+      prop,
+      "sycl::queue::has_property<sycl::property::queue::"
+      "in_order>()");
+}
+// FIXME: re-enable when possibility of a SYCL kernel with an unnamed type is
+// implemented in computcpp
```
bader, line 43:
```suggestion
// FIXME: re-enable when possibility of a SYCL kernel with an unnamed type is
// implemented in computecpp
```

Can we use named kernels here instead? I think unnamed kernels support is not a requirement for queue properties testing. Right?

tests/queue/queue_properties.cpp
```diff
@@ -20,56 +20,243 @@
...
+#endif  // SYCL_CTS_COMPILING_WITH_COMPUTECPP
+void check_in_order(sycl::queue &queue) {
+  check_in_order_prop(queue);
+
+  if (queue.get_device().has(sycl::aspect::usm_device_allocations))
+    check_in_order_functionality(queue);
+  else
+    WARN(
+        "test for in_order functionality is skipped because device doesn't "
+        "support usm_device_allocations");
```
bader, line 91: Can we use SKIP macro instead?

tests/queue/queue_properties.cpp
```diff
@@ -20,56 +20,243 @@
...
+  if (!device.has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
+
+  sycl::queue queue(
+      device, sycl::property_list{sycl::property::queue::enable_profiling()});
+  check_enable_profiling_prop(queue);
+}
+
+// FIXME: re-enable when possibility of a SYCL kernel with an unnamed type is
+// implemented in computcpp
```
bader, line 120:
```suggestion
// implemented in computecpp
```

tests/queue/queue_constructors.cpp
```diff
@@ -19,425 +19,286 @@
...
-
-        if (queue.get_device() != device) {
-          FAIL(log,
-               "queue with device and async_hander was not constructed "
-               "correctly (device equality)");
-        }
+TEST_CASE("Check queue default constructor and destructor", "[queue]") {
+  sycl::queue queue;
+}
+TEST_CASE("Check queue (property_list) constructor", "[queue]") {
```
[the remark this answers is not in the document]
AmirIpma: Done

tests/queue/queue_constructors.cpp
```diff
@@ -19,425 +19,286 @@
...
+  auto device = util::get_cts_object::device();
+  if (!device.has(sycl::aspect::queue_profiling))
+    SKIP("Device does not support queue_profiling");
```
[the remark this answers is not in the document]
AmirIpma: Changed to `in_order` property and removed SKIP since it always supported.

tests/queue/queue_constructors.cpp
AmirIpma: Fixed

tests/queue/queue_properties.cpp
AmirIpma: You are right, changed to named kernel and enabled test case for computecpp

tests/queue/queue_properties.cpp
AmirIpma: Done

tests/queue/queue_properties.cpp
AmirIpma: Removed after enabling test case.

tests/queue/queue_properties.cpp
```diff
@@ -20,56 +20,237 @@
...
+  int loop_array[buffer_size];
+  bool result = false;
+  {
+    sycl::buffer<bool, 1> res_buf(&result, sycl::range<1>(1));
+    sycl::buffer<int, 1> loop_buf(loop_array, sycl::range(buffer_size));
+    // to garantee that data_changed initialized as false before two tested
+    // commands are submitted
+    queue
+        .submit([&](sycl::handler &cgh) {
+          cgh.single_task<class kernel1>([=] { *data_changed = false; });
```
bader, line 55: You need to make forward declarations for kernel typenames to fix the compilation with ComputeCPP.

bader (approved):
LGTM, but we need to fix the compilation with ComputeCPP before the merge.

tests/queue/queue_properties.cpp
AmirIpma: Thank you, done

ProGTX (approved):

bader:
@ProGTX, ComputeCPP build fails with:

```
/sycl/include/SYCL/apis.h:285:39: error: ‘cl::sycl::detail::inspector_t<queue_properties::kernel2> incompatibleLayoutKernel’ has incomplete type
  285 |       detail::inspector_t<kernelName> incompatibleLayoutKernel{};
      |                                       ^~~~~~~~~~~~~~~~~~~~~~~~
/sycl/include/SYCL/apis.h:289:9: error: static assertion failed: The kernel argument memory layout of gcc 9.3.0 is not compatible with the device compiler. Try to use -sycl-driver instead.
  289 |         isLayoutCompatible,
      |         ^~~~~~~~~~~~~~~~~~
```

Any fix suggestions?

fraggamuffin:
issue with compilation model, need to explicitly capture values into the kernel due to different layout between compiler and host
implicit captures have reordered layouts, but spec say = capture is ok
should this be restricted feature set due to different host/device compiler and lambda naming (part of reduced feature set)?
split compilation mean conformance to reduced feature set 
set host compiler to be clang in CI?
do we need split compilation to pass conformance? update configure script, CP to manually build the container

tests/queue/queue_properties.cpp
```diff
@@ -20,56 +20,241 @@
...
+
+  auto prop = queue.get_property<sycl::property::queue::in_order>();
+  check_return_type<sycl::property::queue::in_order>(
+      prop,
+      "sycl::queue::has_property<sycl::property::queue::"
+      "in_order>()");
+}
+
+void check_in_order_functionality(sycl::queue &queue) {
+  bool *data_changed = sycl::malloc_device<bool>(1, queue);
```
keryell, line 48: What happens if USM is not supported by the device?

kopylovanat: If USM is not supported by the device, this part of test is skipped (checked on line 88) and `check_in_order_functionality` is not called. Should we move this check into this function?

fraggamuffin:
docker fix needed

tests/queue/queue_properties.cpp
keryell: Ah good. Perhaps it could be clearer if the test were here.

keryell (approved):
Thanks.

ProGTX:
ComputeCpp Docker image was updated, could you please try running this again?

kopylovanat:
> ComputeCpp Docker image was updated, could you please try running this again?

@ProGTX It seems that it worked, thank you!

PR #487, opened 2023-01-26
Fix conditional inclusion for CTS options

nolmoonen:
Several CTS option macros are defined in the top-level `CMakeLists.txt` using `add_cts_option`. The documentation for the `add_cts_option` function states: 
```cmake
# Important: The preprocessor macro is always set, regardless of its value.
#            Use `#if <OPTION>` instead of `#ifdef <OPTION>`.
```
A quick test indicates that this is indeed required. This PR fixes all violations of this rule.

A related note: the macro `SYCL_CTS_COMPILING_WITH_<IMPLEMENTATION>` is only defined when compiling with `IMPLEMENTATION`. There are many occurrences of `#if SYCL_CTS_COMPILING_WITH_<IMPLEMENTATION>` instead of `#ifdef SYCL_CTS_COMPILING_WITH_<IMPLEMENTATION>` but this is not a problem as the expression evaluates to false if it is not defined. Although it is slightly confusing.

### DIFF
--- a/tests/accessor/accessor_implicit_conversions_fp16.cpp
+++ b/tests/accessor/accessor_implicit_conversions_fp16.cpp
@@ -35,7 +35,7 @@ DISABLED_FOR_TEST_CASE(hipSYCL, ComputeCpp, DPCPP)
     return;
   }
 
-#ifdef SYCL_CTS_ENABLE_FULL_CONFORMANCE
+#if SYCL_CTS_ENABLE_FULL_CONFORMANCE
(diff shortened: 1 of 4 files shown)


bader (approved):
@psalz, is it possible to not define the macro if option is not set? It's very easy to miss the wrong check, so I expect this issue will appear in future again.

ProGTX (approved):

keryell (approved):

PR #492, opened 2023-01-27
Enable reduction tests

aleksmesh:
Fixed a few bugs. Enable tests if possible. The tests used Catch2 now.


tests/reduction/reduction_without_identity_param_common.h
```diff
@@ -274,29 +265,26 @@ void run_test_for_span(FunctorT functor, RangeT range, sycl::queue &queue,
...
+    WARN("Test skipped due to floating point variable cannot be used with " +
+         std::string(typeid(FunctorT).name()) + " functor");
   } else {
     run_test_for_value_ptr<VariableT, UseCombineFlagT, UsePropertyFlag>(
-        functor, range, queue, log, type_name);
-    run_test_for_buffer<VariableT, UseCombineFlagT, UsePropertyFlag>(
-        functor, range, queue, log, type_name);
+        functor, range, queue, type_name);
+    //    run_test_for_buffer<VariableT, UseCombineFlagT, UsePropertyFlag>(
+    //        functor, range, queue, type_name);
```
bader, line 285: Why is this code commented out?

tests/reduction/reduction_without_identity_param_common.h
```diff
@@ -274,29 +265,26 @@ void run_test_for_span(FunctorT functor, RangeT range, sycl::queue &queue,
...
                                        const std::string &type_name) {
   if constexpr (is_sycl_floating_point<VariableT>::value &&
                 (std::is_same<FunctorT, sycl::bit_and<VariableT>>::value ||
                  std::is_same<FunctorT, sycl::bit_or<VariableT>>::value ||
                  std::is_same<FunctorT, sycl::bit_xor<VariableT>>::value)) {
-    log.note(
-        "Test skipped due to floating point variable cannot be used with " +
-        std::string(typeid(FunctorT).name()) + " functor");
+    WARN("Test skipped due to floating point variable cannot be used with " +
+         std::string(typeid(FunctorT).name()) + " functor");
```
bader, line 280: WARN -> SKIP?

tests/reduction/reduction_with_identity_param_fp64.cpp
```diff
@@ -8,38 +8,15 @@
...
+    SKIP("Device does not support double precision floating point operations");
   }
-};
-
-// construction of this proxy will register the above test
-util::test_proxy<TEST_NAME> proxy;
-}  // namespace TEST_NAMESPACE
+  reduction_with_identity::run_test_for_type<double>()(queue, "double");
+}
+}  // namespace reduction_with_identity_param_fp64
```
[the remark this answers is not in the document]
aleksmesh: Done.

tests/reduction/reduction_without_identity_param_common.h
aleksmesh: Done.

tests/reduction/reduction_without_identity_param_common.h
aleksmesh: It's my mistake. Thanks.

ProGTX (approved):

kopylovanat:
CI failed due to new reducer tests from https://github.com/KhronosGroup/SYCL-CTS/pull/478 that were recently merged (they did pass CI because reduction was in filter lists). I'm going to fix these test cases or disable them in current PR.
Since I think fixing these tests might deviate from test plan, to clarify my understanding of Spec - @gmlueck  Do I understand correctly, that because ["To enable compile-time specialization of reduction algorithms, the implementation of the reducer class is unspecified, except for the functions and operators defined in Table 127 and Table 128"](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#reducer-class) and `value_type`, `binary_operation`, `dimensions` are not mentioned in Table 127 and Table 128, it means that this member types are not necessarily should exists in `reducer` implementation?

kopylovanat:
Disabled test cases for 'reducer api core', 'reducer api fp16', 'reducer api fp64'  since it's not fully implemented yet.

@gmlueck I found the answer for my question in the previous comment in this clarification https://github.com/KhronosGroup/SYCL-Docs/pull/345, sorry for bothering you.

tests/reduction/reduction_common.h
```diff
@@ -21,7 +21,7 @@ constexpr bool without_property{false};
 
 constexpr size_t number_iterations{10};
 
-constexpr int identity_value{5};
+constexpr int identity_value{0};
```
keryell, line 24: Could this change mask some implementation feature?
Like imagine that internally the identity value is always ignored and set to 0?

kopylovanat: I changed it to 0 because according to Spec (Table 124 in [4.9.2.1. reduction interface](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#reduction-interface)) "The value of `identity` may be used by the implementation to initialize an unspecified number of temporary accumulation variables." So this value have to be suitable as neutral accumulation value for corresponding operation otherwise result will be unpredictable.
And since only one custom operation without identity (`op_without_identity`) in this tests is basically '+', value is set to zero now.

keryell: That makes sense.

keryell (approved):
Thanks.

PR #493, opened 2023-01-30
Add tests for kernel_bundle working with specialization constants

vladimirkhashev:
Add tests according to [2.3. Tests for working with specialization constants](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/kernel_bundle.asciidoc#23-tests-for-working-with-specialization-constants).


tests/kernel_bundle/kernel_bundle_spec_const.cpp
```diff
@@ -0,0 +1,104 @@
...
+//  limitations under the License.
+//
+*******************************************************************************/
+
+#include "kernel_bundle.h"
+
+namespace kernel_bundle_spec_const {
+
+constexpr sycl::specialization_id<int> SpecName(5);
+constexpr sycl::specialization_id<int> OtherSpecName(5);
```
ProGTX, line 26: I presume the 5 is just a randomly chosen number, but the fact that two different ids use the same number is intentional, because the ids should still be the same? I think that should probably have a comment about that.

tests/kernel_bundle/kernel_bundle_spec_const.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+TEST_CASE(
+    "Check specialization constants functionality with empty kernel_bundle",
+    "[kernel_bundle]") {
+  sycl::context ctx;
+  const auto always_false_selector = [](auto device_image) { return false; };
```
ProGTX, line 32: It's a bit strange that device image selectors work differently from device selectors, but that's a spec issue, the test code here is correct.

ProGTX (approved):

tests/kernel_bundle/kernel_bundle_spec_const.cpp
```diff
@@ -0,0 +1,104 @@
...
+    "[kernel_bundle]") {
+  sycl::device device = sycl_cts::util::get_cts_object::device();
+  sycl::queue queue = sycl_cts::util::get_cts_object::queue();
+  sycl::context ctx = queue.get_context();
+
+  using KernelName = class simple_kernel;
+
+  queue.submit([&](sycl::handler &cgh) {
+    cgh.single_task<KernelName>([=](sycl::kernel_handler h) {
+      int a = h.get_specialization_constant<SpecName>();
```
keryell, line 54: No check about the value read?

tests/kernel_bundle/kernel_bundle_spec_const.cpp
kopylovanat: No, they don't have to be the same, changed it to different value to avoid confusion.

tests/kernel_bundle/kernel_bundle_spec_const.cpp
keryell: At least add a comment so no one will ask again. :-)
Otherwise, there is no use of `a`, which might trigger some compiler warning.

kopylovanat: You are right, there is no use of `a`, removed it and add comment.

keryell (approved):
Thanks.

PR #497, opened 2023-02-01
Add tests for sycl::kernel to cover remaining gaps

vladimirkhashev:
Add tests for sycl::kernel to cover remaining gaps according to test plan https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/kernel.asciidoc.

### DIFF
--- a/tests/kernel/kernel_constructors.cpp
+++ b/tests/kernel/kernel_constructors.cpp
@@ -21,276 +21,215 @@
 
 #include "../common/common.h"
 
-#define TEST_NAME kernel_constructors
-
 template <int a>
 class test_kernel {
  public:
   void operator()() const {}
 };
 
-class kernel0;
-class kernel1;
-
 namespace kernel_constructors__ {
 using namespace sycl_cts;
 
-/** test sycl::kernel
- */
-class TEST_NAME : public sycl_cts::util::test_base {
- public:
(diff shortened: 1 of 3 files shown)


ProGTX (approved):

fraggamuffin:
LGTM

keryell (approved):
Thanks.

tests/kernel/kernel_constructors.cpp
```diff
@@ -21,276 +21,219 @@
...
-          }
-        }
+  if (ctsQueue.get_backend() == sycl::backend::opencl) {
+    auto iopKernelA = sycl::get_native<sycl::backend::opencl>(kernelA);
+    auto iopKernelB = sycl::get_native<sycl::backend::opencl>(kernelB);
+    if (!ctsSelector.is_host() && (iopKernelA != iopKernelB)) {
+      FAIL(
+          "kernel was not constructed correctly. (contains different "
+          "OpenCL kernel object)");
+    }
```
bader, line 56: Should we use `CHECK` instead of `if` + `FAIL`?

vladimirkhashev: Changed it.

bader: Thanks!

PR #499, opened 2023-02-02
Individual functions tests implementation

Melirius:
That is the implementation of mathematical functions tests for CTS.

Even DPCPP cannot compile all the spec prescribed functions (it can do it with `marray` support switched off and commented out function variant `f_select_3` for `sgentype select(sgentype a, sgentype b, bool c)`), therefore the test is excluded from CI.

Type system follows [this scheme](https://github.com/KhronosGroup/SYCL-Docs/issues/353#issuecomment-1406594932), if otherwise will be decided by the WG, it can be changed as tests are written in the way allowing simple regex modification.


tests/kernel_bundle/kernel_bundle.h
```diff
@@ -79,6 +79,36 @@ sycl_cts::util::kernel_restrictions get_restrictions() {
  */
 enum class submit_kernel { yes, no };
 
+/** @brief Call sycl::queue::submit only if dev_is_compat flag is true
```
Melirius, line 82: This is moved to avoid massive number of warnings of forward declaration available only in C++20.

keryell (approved):
Thanks.

Melirius:
Rebased on current upstream.

tests/function_objects/function_objects.h
```diff
@@ -0,0 +1,231 @@
...
+};
+
+/** Checks whether \p T is a sycl::marray type and its data type is integral. */
+template <typename T>
+struct is_integral_marray {
+  static constexpr bool value = false;
+};
+
+template <typename DataT, std::size_t NumElements>
+struct is_integral_marray<sycl::vec<DataT, NumElements>> {
```
bader, line 104: vec -> marray?

tests/function_objects/function_objects_fp16.cpp
```diff
@@ -0,0 +1,72 @@
...
+      sycl_cts::util::extensions::tag::fp16>;
+  if (!availability::check(queue)) {
+    SKIP("Device does not support half precision floating point operations.");
+  }
+
+  const auto types = named_type_pack<TYPES>::generate(TYPE_NAMES);
+  const auto types_only_half = named_type_pack<sycl::half>::generate("half");
+
+  // prevent testing duplicate combinations
+#if 0
```
bader, line 41: Why is this code disabled with preprocessor and not just removed?

tests/function_objects/function_objects_fp16.cpp
```diff
@@ -0,0 +1,72 @@
...
+  for_all_combinations<check_scalar_return_type>(get_op_types(), types,
+                                                 types_only_half);
+  for_all_combinations<check_scalar_return_type>(get_op_types(),
+                                                 types_only_half, types);
+#else
+  WARN(
+      "Specification does not provide rules for the implicit conversion "
+      "surrounding sycl::half. Since most types can be converted to sycl::half"
+      "and vice versa, the type of the conditional operator is ambiguous.");
```
bader, line 50: Do you have a tracker for that issue we can put in TODO comment?

util/math_helper.h
```diff
@@ -56,6 +56,41 @@ template <typename returnT> struct resultRef {
 
 namespace math {
 
+/* return number of elements in a type */
+int numElements(const float &);
```
bader, line 60: Nitpicking: what does it mean: "number of elements in a `const float &` type"?
This looks like a dummy function to simplify the code in templates, which examine type traits of composite types like vector, array, struct, etc. Am I right?
If so, I suggest we update the comment to make it less confusing.

ci/dpcpp.filter
```diff
@@ -1,4 +1,5 @@
 buffer
 exceptions
+math_builtin_api
```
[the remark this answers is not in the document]
nolmoonen: The reason `math_builtin_api` is disabled here is because the Python script that generates the tests has no mechanism for disabling.

`function_objects` is disabled in the CI filter file because hipSYCL cannot compile the header that contains the actual tests as it contains `marray`. Therefore, it cannot be disabled by using `DISABLED_FOR_TEST_CASE` and implementation-specific `#ifdef`s are required, which I understood to be undesirable.

tests/function_objects/function_objects.h
nolmoonen: Good catch, thank you.

tests/function_objects/function_objects_fp16.cpp
nolmoonen: The code is disabled because of "Specification does not provide rules for the implicit conversion surrounding `sycl::half`. Since most types can be converted to `sycl::half` and vice versa, the type of the conditional operator is ambiguous." and not because of that comment, I have moved the comment so it's clear.

The reason it is disabled with the preprocessor is because the "setup" for the test is there, but the test itself cannot handle `half` because the specification is unclear about it. So, the structure of this file is correct but the test cannot be compiled for that data type. Once this specification problem is resolved, the `check_scalar_return_type` can be modified to handle `half` and the statements of this file can be re-enabled.

tests/function_objects/function_objects_fp16.cpp
nolmoonen: Yes, https://github.com/KhronosGroup/SYCL-Docs/issues/350, I have added that.

util/math_helper.h
nolmoonen: Updated the comments.

ci/dpcpp.filter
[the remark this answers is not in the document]
bader: > implementation-specific `#ifdef`s are required, which I understood to be undesirable.

According to my understanding it's the opposite, @psalz added support for `DISABLED_FOR_TEST_CASE` and implementation-specific `#ifdef`s specifically to avoid using CI filters.
@psalz, am I right?

tests/function_objects/function_objects_fp16.cpp
bader: This very minor, but what I'm trying to say is https://github.com/KhronosGroup/SYCL-Docs/issues/350 must be resolved before we can write a test. By the time https://github.com/KhronosGroup/SYCL-Docs/issues/350 is addressed, you might want to write a different test. 
Usually committing commented out code is considered as anti-pattern (e.g. https://kentcdodds.com/blog/please-dont-commit-commented-out-code).

ci/dpcpp.filter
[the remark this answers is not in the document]
nolmoonen: Okay, I trust your understanding of the project better than mine `;)`. I have removed `function_objects` from hipSYCL's CI filter and added the `#ifdef` workaround to deal with `marray`.

For `math_builtin_api` it is not so easy to do. I have not made that part of the PR but I believe this to require quite a big changeset, which may be out of scope for this PR.

tests/function_objects/function_objects_fp16.cpp
nolmoonen: You are right, the test will have to be modified regardless of what is decided in 350. I have removed the half tests for function objects and added a comment in 350 stating that the test should be added once a decision is made.

nolmoonen:
Force-pushed the rebase to resolve the merge conflict.

PR #501, opened 2023-02-06
Fix expected type of reference alias in accessor

steffenlarsen:
The SYCL 2020 specification changed the `reference` member alias in `accessor` to be defined based on `value_type`. This commit changes the expected type in accessor_legacy tests to check for this.

### DIFF
--- a/tests/accessor_legacy/accessor_api_common_all.h
+++ b/tests/accessor_legacy/accessor_api_common_all.h
@@ -61,7 +61,8 @@ void check_accessor_members(sycl_cts::util::logger &log,
                 "value_type is of wrong type");
 
   using reference = typename acc_t::reference;
-  static_assert(std::is_same_v<reference, T &>, "reference is of wrong type");
+  static_assert(std::is_same_v<reference, value_type &>,
+                "reference is of wrong type");
 
   using const_reference = typename acc_t::const_reference;
   static_assert(std::is_same_v<const_reference, const T &>,


steffenlarsen:
Alternatively we remove this check and the check for `value_type` here since these are no longer legacy checks.

steffenlarsen:
Failure in DPCPP should disappear with a version bump.

ProGTX (approved):

tests/accessor_legacy/accessor_api_common_all.h
```diff
@@ -61,7 +61,8 @@ void check_accessor_members(sycl_cts::util::logger &log,
                 "value_type is of wrong type");
 
   using reference = typename acc_t::reference;
-  static_assert(std::is_same_v<reference, T &>, "reference is of wrong type");
+  static_assert(std::is_same_v<reference, value_type &>,
```
keryell, line 64: Just curious: could we have above a simpler
```c++
using value_type = T;
```
?

steffenlarsen: I don't believe so. That's partly my problem with even having these tests in the legacy tests (https://github.com/KhronosGroup/SYCL-CTS/pull/501#issuecomment-1418956716) as `value_type` doesn't follow SYCL 1.2.1 as it is `const` if given read-only access mode. `value_type` above already checks for the SYCL 2020 behavior, so that is what I went with here, but arguably it would be more fitting to just remove those two tests from legacy.

steffenlarsen:
Changes included in DPC++ version bump: https://github.com/KhronosGroup/SYCL-CTS/pull/504
Closing this.

PR #502, opened 2023-02-06
Enable accessor iterator requirement tests

aleksmesh:
Enable accessor iterator requirement tests after few product bugs were fixed.

PR #504, opened 2023-02-06
[CI] Uplift DPCPP version to e47b588

bader:


tests/accessor_legacy/accessor_api_common_all.h
```diff
@@ -61,7 +61,8 @@ void check_accessor_members(sycl_cts::util::logger &log,
                 "value_type is of wrong type");
 
   using reference = typename acc_t::reference;
-  static_assert(std::is_same_v<reference, T &>, "reference is of wrong type");
+  static_assert(std::is_same_v<reference, value_type &>,
```
keryell, line 64: I have already seen this. :-)

bader: Right. I cherry-picked https://github.com/KhronosGroup/SYCL-CTS/pull/504/commits/650660837c346d2edbddec7a4ba9c02109898c2d from https://github.com/KhronosGroup/SYCL-CTS/pull/501. 

As @steffenlarsen said in [this comment](https://github.com/KhronosGroup/SYCL-CTS/pull/501#issuecomment-1418998838), this change should go with or after DPC++ version update. CI says they go together. SYCL implementation version update in CI can validated only with PR created from the local branch. #501 uses private fork, so I created this PR to test DPC++ version uplift. 

NOTE: the PR is a draft and not ready for review yet. I need to make sure there are not regressions added with the version uplift.

bader:
DPC++ implemented a new diagnostic a couple of days ago (https://github.com/intel/llvm/pull/8169), which seems to catch a bug ("bug" according to SYCL 2020) in "legacy" tests.

> ../tests/multi_ptr/multi_ptr_legacy_api_common.h:603:13: note: in instantiation of template class 'sycl::accessor<const float, 1, sycl::access::mode::read_write, sycl::access::target::global_buffer, sycl::access::placeholder::false_t>' requested here

SYCL 2020 says that `const T` (T==float here) can be used with `read` access mode only.
Any objections to remove "invalid" cases from SYCL-2020 branch?

tests/accessor_legacy/accessor_api_common_all.h
steffenlarsen: Thanks @bader ! I have closed the other PR as we cannot go ahead with it without the bump anyway. Open discussions can continue on there, but no point in keeping it open.

keryell:
I do not understand why so many files are removed if it is just some up-lift of DCPP without any other PR description on it.

bader:
> I do not understand why so many files are removed if it is just some up-lift of DCPP without any other PR description on it.

I tried to cover it in [this comment](https://github.com/KhronosGroup/SYCL-CTS/pull/504#issuecomment-1420051174). In nutshell, "legacy" tests are copies of SYCL-1.2.1, which are working almost "as is" for SYCL 2020, but in some cases API has been changed in incompatible way. 
According to my understanding, new multi_ptr tests should cover SYCL 2020 API and we can drop "legacy" tests.
@kopylovanat, could you confirm, please?

kopylovanat:
> In nutshell, "legacy" tests are copies of SYCL-1.2.1, which are working almost "as is" for SYCL 2020, but in some cases API has been changed in incompatible way. According to my understanding, new multi_ptr tests should cover SYCL 2020 API and we can drop "legacy" tests. @kopylovanat, could you confirm, please?

@bader multi_ptr legacy tests are indeed are copies of SYCL-1.2.1 tests with added `access::decorated::legacy` template parameter as stated in test plan [here](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/multi_ptr_test_plan.asciidoc#31-sycl121-compatible-interface). However they test 'legacy interface, inherited from 1.2.1.' that is described in SYCL 2020 in the end of [this paragraph](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:multiptr). As far as I remember the plan initially wasn't to remove these tests unlike it was with accessor legacy tests that were planned to be removed when tests for new accessor interface are ready. Because in SYCL 2020 Specification legacy multi_ptr interface didn't change from SYCL-1.2.1 except for adding `access::decorated::legacy` to distinguish it from new multi_ptr interface. And as I can see right now here the problem with multi_ptr_legacy tests caused by changed accessor not by multi_ptr itself, but I'm not sure if we should try fix them or drop them since it's deprecated feature and tests are now not working because of other changed features.

gmlueck:
> And as I can see right now here the problem with multi_ptr_legacy tests caused by changed accessor not by multi_ptr itself, but I'm not sure if we should try fix them or drop them since it's deprecated feature and tests are now not working because of other changed features.

In general, we should still test features in SYCL 2020 that are deprecated.  A conformant implementation of SYCL 2020 is required to support all features, even the ones marked deprecated.  If we can fix the tests for the legacy multi_ptr, that would be better than dropping them.

bader:
> > And as I can see right now here the problem with multi_ptr_legacy tests caused by changed accessor not by multi_ptr itself, but I'm not sure if we should try fix them or drop them since it's deprecated feature and tests are now not working because of other changed features.
> 
> In general, we should still test features in SYCL 2020 that are deprecated. A conformant implementation of SYCL 2020 is required to support all features, even the ones marked deprecated. If we can fix the tests for the legacy multi_ptr, that would be better than dropping them.

I think the question is about accessor API used in multi_ptr test. Is it deprecated or removed?

gmlueck:
> I think the question is about accessor API used in multi_ptr test. Is it deprecated or removed?

I think most of the SYCL 1.2.1 accessor API's were retained in SYCL 2020, though some are deprecated.  If you have specific questions about whether an API is deprecated vs. removed, let me know and I can answer.

bader:
> > I think the question is about accessor API used in multi_ptr test. Is it deprecated or removed?
> 
> I think most of the SYCL 1.2.1 accessor API's were retained in SYCL 2020, though some are deprecated. If you have specific questions about whether an API is deprecated vs. removed, let me know and I can answer.

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_read_only_accessors says following:

> A const qualified DataT is only allowed for a read-only accessor. Programs which specify a const qualified DataT and any access mode other than access_mode::read are ill formed, and the implementation must issue a diagnostic in this case.

https://github.com/intel/llvm/commit/aa69e4d9b86c9f0e1cb109d7f6870584f3328908 adds diagnostic as said in the spec. It turned out that multi_ptr tests use `const float` type with `read_write` access mode. You can find the full log here: https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4108550601/jobs/7089402839

Is this a test bug or an implementation bug (i.e. should DPC++ issue a deprecation warning instead of an error message)?

This looks like a test bug to me. I agree that we should keep legacy test, but remove a test case for invalid combinations of data types and access modes.

@keryell, I uplifted DPC++ the version prior to the commit with new diagnostic, so we can keep `multi_ptr` legacy tests untouched. I hope this should be enough to unblock #492, #502, and #507. We will move to later DPC++ versions after fixing the issue with `multi_ptr` legacy test. Are you okay with that?

gmlueck:
> Is this a test bug or an implementation bug (i.e. should DPC++ issue a deprecation warning instead of an error message)?

It is a test bug.

> I agree that we should keep legacy test, but remove a test case for invalid combinations of data types and access modes.

+1

bader:
@keryell, do you have any additional comments for this change?

ProGTX (approved):

bader:
> @keryell, do you have any additional comments for this change?

I'm going to merge this change to unblock other pull requests. @keryell, please, let me know if you have any additional comments and I'll address them in follow-up commits.

keryell (commented):
Looks good. Sorry for being late.

PR #515, opened 2023-02-10
Fix wrong #ifdef in kernel_parameters tests

kopylovanat:
SYCL_CTS_COMPILING_WITH_DPCPP was used erroneously


nolmoonen:
`SYCL_CTS_ENABLE_FULL_CONFORMANCE` is always defined, see https://github.com/KhronosGroup/SYCL-CTS/pull/487. You should use `#if` instead of `#ifdef`.

kopylovanat:
> `SYCL_CTS_ENABLE_FULL_CONFORMANCE` is always defined, see #487. You should use `#if` instead of `#ifdef`.

@nolmoonen thank you! Fixed.

ProGTX (approved):

PR #516, opened 2023-02-10
Add comments with links to issues and to PRs

aleksmesh:
Add comments with links to issues and to PRs for tests/accessor

### DIFF
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -223,6 +223,7 @@ inline auto get_all_dimensions() {
   return dimensions;
 }
 // FIXME: re-enable when target::host_task is implemented
+// Issue link https://github.com/intel/llvm/issues/8298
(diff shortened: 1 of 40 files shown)


ProGTX:
Is this something we want to have in the CTS since it's very specific to just one implementation?

kopylovanat:
> Is this something we want to have in the CTS since it's very specific to just one implementation?

@ProGTX 
But since these `#ifndef`'s are also specific to implementations, maybe it would be good to add links when possible, to keep track when it can be enabled?

keryell (approved):
Great idea.
I love that!

bader:
@ProGTX, do you have any objections to add such comments? There are a few other PRs with similar comments, and I'd like to understand your position before merging them.

fraggamuffin:
adding a link is good
also add a line to coding guideline

fraggamuffin:
wait for Peter

PR #519, opened 2023-02-13
Enable kernel bundle api tests

aleksmesh:
Enable kernel bundle api tests for `kernel_bundle::has_kernel` member functions.


aleksmesh:
Failure in DPCPP should disappear with a version bump.

keryell (approved):

PR #525, opened 2023-02-14
Fix and enable tests for optional kernel features

kopylovanat:
Fix tests that are using macros to call kernel defined as a lambda.
Added kernel names to speculative compilation test.
Enable tests for DPCPP.

### DIFF
--- a/tests/optional_kernel_features/kernel_features_common.h
+++ b/tests/optional_kernel_features/kernel_features_common.h
@@ -473,6 +473,12 @@ void run_functor(const bool is_exception_expected,
 template <typename KernelName, call_type CallType>
 class kernel_submission_call;
 
+#define ANOTHER_ASPECT(current_aspect)                                    \
+  ((current_aspect == sycl::aspect::fp16)                                 \
+       ? sycl::aspect::fp64                                               \
+       : ((current_aspect == sycl::aspect::fp64) ? sycl::aspect::atomic64 \
+                                                 : sycl::aspect::fp16))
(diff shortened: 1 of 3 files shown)


tests/optional_kernel_features/kernel_features_common.h
```diff
@@ -473,6 +473,12 @@ void run_functor(const bool is_exception_expected,
 template <typename KernelName, call_type CallType>
 class kernel_submission_call;
 
+#define ANOTHER_ASPECT(current_aspect)                                    \
+  ((current_aspect == sycl::aspect::fp16)                                 \
+       ? sycl::aspect::fp64                                               \
+       : ((current_aspect == sycl::aspect::fp64) ? sycl::aspect::atomic64 \
+                                                 : sycl::aspect::fp16))
+
```
keryell, line 481: Add a comment.
Why a macro?
I have the feeling that a `constexpr` function would be cleaner.

kopylovanat: I used macro because there is a problem of using constexpr variable in attribute for nested lambda, similar to example here https://reviews.llvm.org/D122494 and similar to example here  https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2036r3.html#odr-used-when-not-odr-usable
However, I realized that static can be used instead of macro, changed to it.

keryell (approved):
Great!
I missed the fact it was used in an attribute...

PR #528, opened 2023-02-14
Fix group_functions tests. Enable group_shift and group_pemute tests.

aleksmesh:
Removed `group_permute` tests for `sycl::group` according to SYCL2020 specification [4.17.4.3. permute](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_permute)
`1. Constraints: Available only if std::is_same_v<std::decay_t<Group>, sub_group> is true and T is a trivially copyable type.`

Removed `group_shift` tests for `sycl::group` according to SYCL2020 specification [4.17.4.2. shift_left and shift_right](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_shift_left_and_shift_right)
`1. Constraints: Available only if std::is_same_v<std::decay_t<Group>, sub_group> is true and T is a trivially copyable type.`

### DIFF
--- a/tests/group_functions/group_barrier.cpp
+++ b/tests/group_functions/group_barrier.cpp
@@ -32,6 +32,7 @@ TEMPLATE_TEST_CASE_SIG("Group barriers", "[group_func][dim]", ((int D), D), 1,
 
   // FIXME: hipSYCL and DPCPP have no implemented
   //  atomic_fence_scope_capabilities query
+  //  Issue to link https://github.com/intel/llvm/issues/8323
 #if !(defined(SYCL_CTS_COMPILING_WITH_HIPSYCL) || \
       defined(SYCL_CTS_COMPILING_WITH_DPCPP))
   std::vector<sycl::memory_scope> supported_barriers =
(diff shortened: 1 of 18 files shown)


keryell:
It is unclear to me why it has been removed from your PR description. What was the problem?

aleksmesh:
> It is unclear to me why it has been removed from your PR description. What was the problem?

The problem is that `std::is_same_v<std::decay_t<sycl::group<N>>, sycl::sub_group> == false`, that is against specification requirement: 
`1. Constraints: Available only if std::is_same_v<std::decay_t<Group>, sub_group> is true and T is a trivially copyable type.`

keryell:
> The problem is that `std::is_same_v<std::decay_t<sycl::group<N>>, sycl::sub_group> == false`, that is against specification requirement: `1. Constraints: Available only if std::is_same_v<std::decay_t<Group>, sub_group> is true and T is a trivially copyable type.`

This is the kind of important information I would expect in the commit messages! :-)

keryell (approved):
Thanks.

PR #534, opened 2023-02-21
Fix scalars tests

vladimirkhashev:
Added device check for fp16 and fp64 support.


tests/scalars/scalars_interopability_types.cpp
```diff
@@ -115,19 +124,54 @@ class TEST_NAME : public util::test_base {
...
+            // Floating Point Interop Data Type
+            // sign
             accSignResult[9] = check_type_sign<sycl::cl_float>(true);
-            accSignResult[10] = check_type_sign<sycl::cl_double>(true);
 
-            // sizes
-            accSizeResult[9] = check_type_min_size<sycl::cl_half>(2);
+            // size
             accSizeResult[10] = check_type_min_size<sycl::cl_float>(4);
-            accSizeResult[11] = check_type_min_size<sycl::cl_double>(8);
```
keryell, line 127: All these magical numbers are a nightmare. :-(
As such, now you need to update lines 90, 91, 94, 96.
Probably all this could be metaprogrammed by using a list of types from a util library to be written (inspired by a lot of existing copy-pasted code all over vect/marray :-/ ), create buffers with the size of this list and use a util function like
```
for_type(type_list, [&](auto& t) {
  using type = decltype(type)::type;
  accSignResult[i++] = check_type_sign<type>(std::is_signed_v(type);
}
```
@bader any motivated good C++ programmer on your side? :-)
That said, it might not worth it...

tests/scalars/scalars_interopability_types.cpp
```diff
@@ -115,19 +124,54 @@ class TEST_NAME : public util::test_base {
...
+          myQueue
+              .submit([&](sycl::handler &cgh) {
+                auto accSignResult =
+                    bufSignResult.get_access<sycl::access_mode::read_write>(
+                        cgh);
+                auto accSizeResult =
+                    bufSizeResult.get_access<sycl::access_mode::read_write>(
+                        cgh);
+
+                cgh.single_task<scalars_interopability_fp64>([=]() {
```
keryell, line 164: Useless
```suggestion
                cgh.single_task<scalars_interopability_fp64>([=] {
```

tests/scalars/scalars_interopability_types.cpp
keryell: Ah I have found it. This is `tests/common/type_coverage.h` with `unnamed_type_pack` and `for_all_types`.

vladimirkhashev: Could you please clarify what do you mean about updating lines 90, 91, 94, 96? Should I refactor the code for this test using `unnamed_type_pack` and `for_all_types` in this PR?

tests/scalars/scalars_interopability_types.cpp
vladimirkhashev: Fixed.

tests/scalars/scalars_interopability_types.cpp
keryell: I have the feeling that in removing some tests that used to have a writing slot in the arrays you can now use smaller arrays and buffers, and as such you should adapt accordingly the new size on lines  90, 91, 94, 96.
For the refactoring of the code for this test using `unnamed_type_pack` and `for_all_types` in this PR, I do not know. The code would be more compact, but at the end perhaps more difficult to understand compared to how the code is right now.
Let's keep it as is for now, except if @bader is pushing for it.

bader: > I have the feeling that in removing some tests that used to have a writing slot in the arrays you can now use smaller arrays and buffers, and as such you should adapt accordingly the new size on lines 90, 91, 94, 96. For the refactoring of the code for this test using `unnamed_type_pack` and `for_all_types` in this PR, I do not know. The code would be more compact, but at the end perhaps more difficult to understand compared to how the code is right now. Let's keep it as is for now, except if @bader is pushing for it.

These tests were moved under `if` statement i.e. not removed, so technically we can't use smaller arrays and buffers. 

Having said that, I agree that it's better to refactor the code to make it less error-prone, but I don't mind doing this refactoring separately.
@vladimirkhashev, it's up to you.

keryell: Oh I see. The same buffer is completed by some other kernels conditionally.
Sorry for not having looked at the full file. :-(
Then it becomes too complex to refactor the code with these 3 kernels.
Leave it as it is. :-)
But while you are here, you could fix the typo in the file name: 
`scalars_interopERability_types.cpp`.

keryell (approved):
Thanks!

tests/scalars/scalars_interopability_types.cpp
vladimirkhashev: So, I'm leaving it as it is for now. :-)
The file name has been fixed.

PR #535, opened 2023-02-21
Removed std containers from type_coverage, kernel_parameters and accessor

aleksmesh:
Due to discussion in this PR [#357](https://github.com/KhronosGroup/SYCL-Docs/pull/357,) the statement in specification section [3.13.1. Device copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable) has changed from:
`Any type that is trivially copyable (as defined by the C++ core language) is implicitly device copyable. In addition, the following types from the C++ core language are implicitly device copyable`

to:

`Although implementations are not required to support device code that calls library functions from the C++ core language, some implementations may provide device support for some of these functions. If the implementation provides device support for one of the following classes, that type is also implicitly device copyable`

usage of std containers types has been removed from accessor and kernel_parameters tests, because there is no guarantee that std containers are supported in device code. Test plan for kernel_parameters is corrected for the same reason.


keryell (approved):
Sad but reasonable.

PR #537, opened 2023-02-21
Fix and improve reduction testing coverage

steffenlarsen:
This commit makes the following fixes and adjustments to the reduction tests:
 * Adds missing test case combinations of range type and `initialize_to_property` usage for `half` and `double` types.
 * Avoid using `initialize_to_property` for identityless test cases.
 * Use the correct initial value for expected reduction results with booleans.

### DIFF
--- a/tests/reduction/reduction_common.h
+++ b/tests/reduction/reduction_common.h
@@ -118,16 +118,15 @@ VariableT get_init_value_for_reduction() {
 template <typename VariableT, typename FunctorT, bool UsePropertyFlagT = false>
 VariableT get_init_value_for_expected_value() {
   VariableT init_value{};
-  if constexpr (std::is_same_v<VariableT, bool>)
-    get_init_value_bool<FunctorT, UsePropertyFlagT>(init_value);
-  else if constexpr (UsePropertyFlagT) {
+  if constexpr (UsePropertyFlagT) {
     // case when using reduction with initialize_to_identity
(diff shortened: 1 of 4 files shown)


tests/reduction/reduction_common.h
```diff
@@ -118,16 +118,15 @@ VariableT get_init_value_for_reduction() {
...
-  else if constexpr (UsePropertyFlagT) {
+  if constexpr (UsePropertyFlagT) {
     // case when using reduction with initialize_to_identity
     if constexpr (sycl::has_known_identity<FunctorT, VariableT>::value)
       init_value = sycl::known_identity<FunctorT, VariableT>::value;
     else
       init_value = identity_value;
   } else {
-    init_value = init_value_without_property_case;
+    // otherwise it should use the same intial value as the reduction.
```
keryell, line 128: Configure your editor/IDE to have a spell checker in the comments
```suggestion
    // Otherwise it should use the same initial value as the reduction.
```

keryell (changes requested):

tests/reduction/reduction_common.h
keryell: Just in case, I use Emacs with
```elisp
;; Add spell-checking in program modes for the comments
(add-hook 'prog-mode-hook 'flyspell-prog-mode)
```

keryell (approved):
Thanks.

PR #538, opened 2023-02-22
Add missed cases for buffer constructor with shared_ptr[]

kopylovanat:
Added test cases for buffer constructor with shared_ptr[] and fixed existing cases with shared_ptr.
Fixed some comments.
Removed unnecessary warnings with testing types info.
Also removed test case buffer(shared pointer, range, mutex, std allocator) because there is no such constructor for buffer in SYCL 2020 spec, and in reality it just duplicates previous test case


tests/buffer/buffer_constructors_common.h
```diff
@@ -221,6 +226,19 @@ class buffer_ctors {
...
+    {
+      std::shared_ptr<T[]> data(new T[size]);
+      std::fill(data.get(), (data.get() + size), 0);
+      sycl::buffer<T, dims> buf(data, r, propList);
+      sycl::buffer<T, dims> buf1(data, r);
+      constexpr bool data_verify = true;
+      if (!check_buffer_constructor(buf, r, data_verify) ||
+          !check_buffer_constructor(buf1, r, data_verify)) {
+        FAIL(log, "(shared pointer[], range) constructor fail.");
+      }
```
bader, line 239:
```
if (<condition>) {
  FAIL(...);
}
```
should be replaced by an assertion
```
CHECK(condition);
```

tests/buffer/buffer_constructors_common.h
```diff
@@ -303,11 +345,13 @@ class buffer_ctors {
...
           !check_buffer_constructor(buf_iter1, r_exp, data_verify)) {
-        FAIL(log, "buffer iterator constructor constructor fail.");
+        FAIL(log,
+             "buffer iterator constructor with allocator template param fail.");
       }
     }
 
-    /* Check subBuffer (buffer, id, range) constructor*/
+    /* Check subBuffer (buffer, id, range) constructor with allocator template
+     * param*/
```
bader, line 354:
```suggestion
     * param */
```

tests/buffer/buffer_constructors_common.h
```diff
@@ -200,278 +198,266 @@ class buffer_ctors {
...
+      CHECK(std::equal(ref.cbegin(), ref.cend(), cont2.cbegin(),
+                       value_operations::are_equal<T, T>));
+      CHECK(alloc_log.str().empty());
     }
 
-    /* check (shared pointer, range) constructor*/
     {
-      std::shared_ptr<T> data(new T[size]);
+      INFO("check (shared pointer, range) constructor");
+      std::shared_ptr<T> data(new T[size], deleter<T>());
```
keryell, line 210: By looking at https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr I do not think you need to provide a specific deleter, even in old C++17. It should use the right deleter by default.
This would be necessary for `std::make_shared`, which has been fixed only in C++20.

tests/buffer/buffer_constructors_common.h
```diff
@@ -61,6 +61,11 @@ class logging_alloc {
   }
 };
 
+template <typename T>
+struct deleter {
+  void operator()(T const *p) { delete[] p; }
+};
```
keryell, line 67: Useless?

tests/buffer/buffer_constructors_common.h
```diff
@@ -200,278 +198,266 @@ class buffer_ctors {
...
-      }
+      CHECK(check_buffer_constructor(buf, r, data_verify));
+      CHECK(check_buffer_constructor(buf1, r, data_verify));
+    }
+
+    {
+      INFO("check (shared pointer[], range, allocator) constructor");
+      sycl::buffer_allocator<T> buf_alloc;
+      std::shared_ptr<T[]> data(new T[size]);
+      std::fill(data.get(), (data.get() + size), 0);
```
keryell, line 379:
```suggestion
      std::fill(data.get(), data.get() + size, 0);
```
Overkill?

kopylovanat: Removed.

tests/buffer/buffer_constructors_common.h
```diff
@@ -200,278 +193,266 @@ class buffer_ctors {
...
+                       value_operations::are_equal<T, T>));
+      CHECK(alloc_log.str().empty());
     }
 
-    /* check (shared pointer, range) constructor*/
     {
-      std::shared_ptr<T> data(new T[size]);
-      std::fill(data.get(), (data.get() + size), 0);
+      INFO("check (shared pointer, range) constructor");
+      std::shared_ptr<T> data(new T[size], std::default_delete<T[]>());
```
keryell, line 205: @kopylovanat By rereading your comment

> But on cppreference for constructor `template< class Y > shared_ptr( Y* ptr )` of `std::shared_ptr<T>` it is stated that "3) Uses the delete-expression `delete ptr` if `T` is not an array type; delete[] ptr if T is an array type (since C++17) as the deleter."
And in tested case T is not an array type, so it seems delete ptr will be used, not delete[] ptr. Am I missing something here?

and looking again at the C++17 draft
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4660.pdf
I think I missed the very important point of your comment: *in tested case T is not an array type*.
Yes, you are right, `new T[size]` does return a pointer `T*` and not an array `T[]`. :-(
OK, I think I understand now. I was thinking you were implementing the test from line 215 with `std::shared_ptr<T[]>`.
But actually this is the test really testing with a `std::shared_ptr<T>`!
So yes, you need to provide the real deleter corresponding to the new.

keryell (approved):
Great!

PR #539, opened 2023-02-22
Add test plan for any_device_has and all_devices_have

kopylovanat:
Test plan for traits `any_device_has` and `all_devices_have` that the application can use to query aspects at compilation time
that are described in [4.6.4.3. Device aspects](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-aspects) .

### DIFF
--- a/test_plans/any_device_has_all_devices_have.asciidoc
+++ b/test_plans/any_device_has_all_devices_have.asciidoc
@@ -34,9 +34,8 @@ Tests described below are performed for following aspects:
 
 == Tests
 
-* Check that `any_device_has<aspect>` inherits from `std::true_type` if there is any device available that supports tested aspect,
-and inherits from `std::false_type` otherwise
-* Check that `any_device_has_v<aspect>` = `any_device_has<aspect>::value`
-* Check that `all_devices_have<aspect>` inherits from `std::true_type` if all available device that supports tested aspect,
-and inherits from `std::false_type` otherwise
-* Check that `all_devices_have_v<aspect>` = `all_devices_have<aspect>::value`
+* Check each device to see if it has aspect `A`.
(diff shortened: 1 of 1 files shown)


test_plans/any_device_has_all_devices_have.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+* `sycl::aspect::usm_host_allocations`
+* `sycl::aspect::usm_atomic_host_allocations`
+* `sycl::aspect::usm_shared_allocations`
+* `sycl::aspect::usm_atomic_shared_allocations`
+* `sycl::aspect::usm_system_allocations`
+
+== Tests
+
+* Check that `any_device_has<aspect>` inherits from `std::true_type` if there is any device available that supports tested aspect,
+and inherits from `std::false_type` otherwise
```
gmlueck, line 38: I think this test is not valid.  What exactly do you mean by:

> if there is any device available that supports tested aspect

I assume that you plan to run the program and test whether any of the devices that are available on the test system have this aspect?

This is not a valid test.  The spec says:

> The trait `any_device_has<aspect>` inherits from `std::true_type` if the **compilation environment** supports any device which has the specified aspect

(Emphasis mine.)

It is possible that the compilation environment could support some devices that are not present on the system where you run the test program.

Therefore, I think you need to weaken the test to this:

* Check each device to see if it has aspect _A_.
* If some device has aspect _A_ and `any_device_has_v<A>` is `false`, then the test fails.
* If some device does not have aspect _A_ and `all_devices_have_v<A>` is `true`, then the test fails.

kopylovanat: Thank you! Fixed test plan.

PR #541, opened 2023-02-22
Fix check has_kernel_bundle with empty vector

kopylovanat:
Because `has_kernel_bundle(ctx, {dev}, {k_ids})` should return `true` if `{k_ids}` is empty

### DIFF
--- a/tests/kernel_bundle/has_kernel_bundle_zero_kernels.h
+++ b/tests/kernel_bundle/has_kernel_bundle_zero_kernels.h
@@ -20,10 +20,10 @@
 namespace sycl_cts::tests::has_kernel_bundle::check {
 
 static const std::string error_message_ex_false{
-    "sycl::has_kernel_bundle return true for zero kernels"};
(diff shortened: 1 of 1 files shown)


tests/kernel_bundle/has_kernel_bundle_zero_kernels.h
```diff
@@ -19,9 +19,12 @@
 
 namespace sycl_cts::tests::has_kernel_bundle::check {
 
-static const std::string error_message{
+static const std::string error_message_ex_false{
     "sycl::has_kernel_bundle return true for zero kernels"};
 
+static const std::string error_message_ex_true{
+    "sycl::has_kernel_bundle return false for empty kernel_id vector"};
```
keryell, line 26:
```suggestion
static const std::string error_message_ex_false{
    "sycl::has_kernel_bundle returned true for zero kernel"};

static const std::string error_message_ex_true{
    "sycl::has_kernel_bundle returned false for empty kernel_id vector"};
```

keryell (approved):
Thanks.

PR #542, opened 2023-02-23
Fixed allocation of too large buffer on the device for the group_barrier test

vladimirkhashev:
The work group size is limited for group_barrier test because of out of memory exceptions was being thrown at runtime due to large buffer size which depends on the size of the work group.

### DIFF
--- a/tests/group_functions/group_barrier.cpp
+++ b/tests/group_functions/group_barrier.cpp
@@ -23,6 +23,17 @@
 
 #include "group_functions_common.h"
 
+/** Returns maximum int32_t elements which can be allocated in device global or
+ * local memory. */
+template <typename DeviceMemoryInfoAspect>
+inline uint64_t max_elements_in_device_memory() {
+  using el_type = int32_t;
+  sycl::device device = sycl_cts::util::get_cts_object::device();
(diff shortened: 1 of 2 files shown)


Melirius:
Good catch, but you need to introduce guards and warnings, and then separate handling for hipSYCL and ComputeCPP, they have not yet implemented device information descriptors as classes: see [1](https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4254879278/jobs/7401747758#step:8:85), [2](https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4254879278/jobs/7401747297#step:8:338).

tests/group_functions/group_barrier.cpp
```diff
@@ -23,6 +23,17 @@
 
 #include "group_functions_common.h"
 
+/** Returns maximum int32_t elements which can be allocated in device global or
+ * local memory. */
```
keryell, line 27:
```suggestion
/** Returns maximum int32_t elements which can be allocated in device global or
    local memory. */
```

ProGTX (approved):

keryell (approved):
Thanks.

PR #544, opened 2023-02-27
Remove sycl::device_selector from tests

kopylovanat:
Removed sycl::device_selector usage and tests for sycl::device_selector since it's not part of SYCL 2020.


tests/common/cts_selector.h
```diff
@@ -44,14 +44,14 @@ class cts_selector : public sycl::device_selector {
    *  return <  0  : device will never be selected
    *  return >= 0  : positive device rating
    */
-  virtual int operator()(const sycl::device& dev) const {
+  int operator()(const sycl::device& dev) const {
```
keryell, line 47: In SYCL 2020 the device selector is just a simple callable so it could be just a simple function too.
So you could simplify the code everywhere by replacing the classes by just functions.
I do not know the impact on the use cases. If it does imply changing the use cases everywhere, just keep it with the classes.

ProGTX (approved):

tests/common/cts_selector.h
kopylovanat: Turned cts_selector into function.
Only use case in kernel constructors test had to be updated. For other tests just code simplification.

tests/cuda_interop/cuda_backend_avail.cpp
```diff
@@ -10,7 +10,7 @@
 
 #include "cuda_helper.hpp"
 
-class cuda_selector : public sycl::device_selector {
+class cuda_selector {
```
keryell, line 13: This one could be replaced too.

tests/kernel/kernel_constructors.cpp
```diff
@@ -54,16 +53,18 @@ TEST_CASE("Test copy constructor", "[kernel]") {
     INFO(
         "kernel was not constructed correctly. (contains different "
         "OpenCL kernel object)");
-    CHECK((ctsSelector.is_host() || (iopKernelA == iopKernelB)));
+    CHECK((sycl::device(cts_selector)
+                   .get_info<sycl::info::device::device_type>() ==
+               sycl::info::device_type::host ||
+           (iopKernelA == iopKernelB)));
```
keryell, line 59: This test has to be updated since the concept of `host` does not exist anymore in SYCL 2020.

tests/kernel/kernel_constructors.cpp
```diff
@@ -85,16 +86,18 @@ TEST_CASE("Test assignment operator", "[kernel]") {
...
```
keryell, line 92: Same everywhere.

keryell (commented):
Thanks for the changes.

tests/cuda_interop/cuda_backend_avail.cpp
kopylovanat: Done.

tests/kernel/kernel_constructors.cpp
kopylovanat: Removed safeguards for host in kernel tests.

keryell (approved):
Thanks.

PR #545, opened 2023-02-27
Fix device_image_semantics and kernel_bundle_semantics tests

kopylovanat:
To make sure that test are getting correct device_image and kernel_bundle for semantics checks.


keryell (approved):
Thanks.

PR #546, opened 2023-02-27
Fix accessor legacy bool

AmirIpma:
Fixed compfail with `std::vector<bool>` with FULL_CONFORMANCE mod by replacing it with `std::unique_ptr<T[]>`. Compfail was caused because vector has no member `data()` for `bool` type.
Changed expected type for `accessor::get_pointer()` in accordance to SYCL 2020 spec.


bader:
@AmirIpma, please revert https://github.com/KhronosGroup/SYCL-CTS/pull/546/commits/8b649c399ea75833f9e0d5ac27094080b190c5ed and apply one-line patch.

```diff
--- tests/accessor_legacy/accessor_api_buffer_common.h	(before formatting)
+++ tests/accessor_legacy/accessor_api_buffer_common.h	(after formatting)
@@ -738,8 +738,7 @@
     std::unique_ptr<T[]> dataMultiDimSyntax;
     get_buffer_input_data<T>(count, dims, dataMultiDimSyntax, useIndexesWrite);
 
-    static constexpr bool isHostBuffer =
-        (target == sycl::target::host_buffer);
+    static constexpr bool isHostBuffer = (target == sycl::target::host_buffer);
 
     auto errors = get_error_data(isHostBuffer ? 2 : 4);
     {
```
This should fix code formatting change and make it easier to review the PR. Right now, it has a lot of formatting changes not related to your fix.

tests/accessor_legacy/accessor_api_common_buffer_local.h
```diff
@@ -64,7 +64,8 @@ struct explicit_pointer<T, sycl::target::constant_buffer> {
 */
 template <typename T>
 struct explicit_pointer<T, sycl::target::host_buffer> {
-  using type = T *;
+  using type = std::add_pointer_t<T>;
+  ;
```
bader, line 68:
```suggestion
  using type = std::add_pointer_t<T>;
```

bader:
> Fixed compfail with std::vector< bool > with FULL_CONFORMANCE mod.

Coud you expand on that problem, please?

AmirIpma:
Had to use force push because of wrong changes reverting

AmirIpma:
> > Fixed compfail with std::vector< bool > with FULL_CONFORMANCE mod.
> 
> Coud you expand on that problem, please?

Changed PR description.

tests/accessor_legacy/accessor_api_common_buffer_local.h
```diff
@@ -50,7 +50,7 @@ struct explicit_pointer<T, sycl::target::local> {
 */
 template <typename T>
 struct explicit_pointer<T, sycl::target::device> {
```
bader, line 52: Don't we need to modify specializations for `sycl::target::local` and `sycl::target::constant_buffer` too?

tests/accessor_legacy/accessor_api_buffer_common.h
```diff
@@ -409,15 +409,16 @@ class check_buffer_accessor_api {
         "check_buffer_accessor_api::reads", typeName, log);
 #endif  // SYCL_CTS_ENABLE_VERBOSE_LOG
 
-    auto dataIdSyntax = get_buffer_input_data<T>(count, dims);
-    auto dataMultiDimSyntax = get_buffer_input_data<T>(count, dims);
-    auto errors = get_error_data(2);
+    std::unique_ptr<T[]> dataIdSyntax;
+    get_buffer_input_data<T>(count, dims, dataIdSyntax);
```
keryell, line 413: Why `auto` did not work here as before?

tests/accessor_legacy/accessor_api_utility.h
```diff
@@ -745,21 +746,24 @@ std::vector<T> get_buffer_input_data(size_t count, int dims,
       data[0] = get_zero_dim_buffer_value<T>();
     }
   } else {
-    std::fill(std::begin(data), std::end(data), 0);
+    for (size_t i = 0; i < count; ++i) {
```
keryell, line 749: This change requires a comment!

tests/accessor_legacy/accessor_api_utility.h
```diff
@@ -745,21 +746,24 @@ std::vector<T> get_buffer_input_data(size_t count, int dims,
...
+ * @return Pointer to array
  */
-std::vector<int> get_error_data(size_t count) {
+std::unique_ptr<int[]> get_error_data(size_t count) {
   static constexpr int dims = 1;
   static constexpr bool useIndexes = false;
-  return get_buffer_input_data<int>(count, dims, useIndexes);
+  std::unique_ptr<int[]> errors;
+  get_buffer_input_data<int>(count, dims, errors, useIndexes);
+  return errors;
```
keryell, line 766: Should not change.

tests/accessor_legacy/accessor_api_utility.h
```diff
@@ -727,16 +728,16 @@ T get_zero_dim_buffer_value() {
...
  *        If false, data will be zero initialized.
- * @return Initialized data container
  */
 template <typename T>
-std::vector<T> get_buffer_input_data(size_t count, int dims,
-                                                bool useIndexes = true) {
-  auto data = std::vector<T>(count);
+void get_buffer_input_data(size_t count, int dims, std::unique_ptr<T[]>& data,
+                           bool useIndexes = true) {
+  data = std::make_unique<T[]>(count);
```
keryell, line 740: It is possible to return a `unique_ptr` instead of this complex code.

keryell (commented):
The infamous `std::vector<bool>` hit again.
Since this is a subtle problem, I guess there should be some comments in the code explaining why the code is now quite more complex.
Otherwise an alternative would be to create the buffer from 2 iterators on the `std::vector` `instead` of using a non existing `.data()` some times, perhaps with some tweaking at that place with constexpr conditional code instead of making the code more complex every where.

tests/accessor_legacy/accessor_api_common_buffer_local.h
AmirIpma: No, these targets uses correct types due to spec.

ProGTX (approved):

tests/accessor_legacy/accessor_api_utility.h
```diff
@@ -731,12 +732,12 @@ T get_zero_dim_buffer_value() {
...
- * @return Initialized data container
+ * @return unique_ptr to initialized array
  */
 template <typename T>
-std::vector<T> get_buffer_input_data(size_t count, int dims,
-                                                bool useIndexes = true) {
-  auto data = std::vector<T>(count);
+std::unique_ptr<T[]> get_buffer_input_data(size_t count, int dims,
+                                           bool useIndexes = true) {
+  auto data = std::make_unique<T[]>(count);
```
keryell, line 740: Some alternative design could have been:
```c++
auto get_buffer_input_data(size_t count, int dims,
                                                bool useIndexes = true) {
  using return_type = std::vector<std::conditional_t<std::is_same_v<T, bool>, std::int8_t, T>>;
  auto data = return_type(count);
```
and then I have the feeling that there is no need to change anything else in the code.

Or even restructure the code for not using `get_buffer_input_data` at all but create an uninitialized buffer and initialize the data with an `host_accessor` on it, check the results with an `host_accessor` on it.
So no more `std::vector` at all. :-)
But this is probably too much for this PR.

AmirIpma: I think alternative design will make the code more complex because we have to cast raw pointer of vector from `int8_t` to `bool` , so there are two possible implementations: add type checks or use `reinterpret_cast`  on each buffer constructor, whereas I use just `data.get()`.

Restructuring will make the code even more complex. `get_buffer_input_data` is used many times, so we will replace this function with another using `host_accessor` and change every buffer initialization.

bader (approved):

tests/accessor_legacy/accessor_api_buffer_common.h
```diff
@@ -732,18 +729,16 @@ class check_buffer_accessor_api {
...
-    auto dataMultiDimSyntax =
-        get_buffer_input_data<T>(count, dims, useIndexesWrite);
+    auto dataMultiDimSyntax = get_buffer_input_data<T>(count, dims);
 
-    static constexpr bool isHostBuffer =
-        (target == sycl::target::host_buffer);
-    auto errors = get_error_data(isHostBuffer ? 2 : 4);
+    static constexpr bool isHostBuffer = (target == sycl::target::host_buffer);
 
+    auto errors = get_error_data(isHostBuffer ? 2 : 4);
```
keryell, line 736: `isHostBuffer` ? 2 : 4` is a magical incantation used twice.
Put in in a variable with a clear good name which will be used twice.

tests/accessor_legacy/accessor_api_utility.h
keryell: OK. Thanks for the argumentation.

keryell (approved):
Thanks!
Just a minor not.

tests/accessor_legacy/accessor_api_buffer_common.h
AmirIpma: Done

PR #553, opened 2023-03-07
Add tests for any_device_has and all_devices_have

kopylovanat:
Add test for any_device_has and all_devices_have according to [test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/any_device_has_all_devices_have.asciidoc).

Only devices that support the trivial kernel is considered to make sure they are supported by the code that the compiler generates, to align with that clarification https://github.com/KhronosGroup/SYCL-Docs/pull/370

### DIFF
--- a/test_plans/any_device_has_all_devices_have.asciidoc
+++ b/test_plans/any_device_has_all_devices_have.asciidoc
@@ -34,8 +34,14 @@ Tests described below are performed for following aspects:
 
 == Tests
 
-* Check each device to see if it has aspect `A`.
-* If some device has aspect `A`, check that `any_device_has_v<A>` is `true`
-and `any_device_has<A>` inherits from `std::true_type`.
-* If some device does not have aspect `A`, check that `all_devices_have_v<A>` is `false`
-and `all_devices_have<A>` inherits from `std::false_type`.
+* The test defines an empty kernel K as a mechanism to determine which devices are supported by the compilation environment.
+* At runtime, iterate over all devices returned by device::get_devices. For each device `D`:
+** Check if `is_compatible<K>(D)` is `true`. Since kernel `K` is empty, it should be compatible with any device that is supported by the compilation environment.
+** For each aspect `A` check if `D.has_aspect(A)` is `true`.
+** If `(is_compatible<K>(D) && D.has(A))` is `true`, then `any_device_has_v<A>` must be `true` and `any_device_has<A>` must inherit from `std::true_type`.
+** If `any_device_has_v<A>` is `false` then `(is_compatible<K>(D) && D.has(A))` must be `false`.
+** If `(is_compatible<K>(D) && !D.has(A))` is `true`, then `all_devices_have_v<A>` must be `false`
+and `all_devices_have<A>` must inherit from `std::false_type`.
+** If `all_devices_have_v<A>` is `true` then `(is_compatible<K>(D) && D.has(A))` or `(!is_compatible<K>(D) && !D.has(A))` must be `true`.
+
+
--- a/test_plans/any_device_has_all_devices_have.asciidoc
+++ b/test_plans/any_device_has_all_devices_have.asciidoc
@@ -42,6 +42,6 @@ Tests described below are performed for following aspects:
 ** If `any_device_has_v<A>` is `false` then `(is_compatible<K>(D) && D.has(A))` must be `false`.
 ** If `(is_compatible<K>(D) && !D.has(A))` is `true`, then `all_devices_have_v<A>` must be `false`
 and `all_devices_have<A>` must inherit from `std::false_type`.
-** If `all_devices_have_v<A>` is `true` then `(is_compatible<K>(D) && D.has(A))` or `(!is_compatible<K>(D) && !D.has(A))` must be `true`.
+** If `all_devices_have_v<A>` is `true` then `(D.has(A) || !is_compatible<K>(D))` must be `true`.
 
 
--- a/tests/device_selector/device_selector_aspect.cpp
+++ b/tests/device_selector/device_selector_aspect.cpp
@@ -408,6 +408,7 @@ class check_any_device_has_all_devices_have {
  public:
   void operator()(const std::string& aspect_name) {
     using k_name = kernel_aspect<AspectT>;
+    INFO("Check for aspect: " + aspect_name);
 
     sycl::queue queue = util::get_cts_object::queue();
     queue.submit(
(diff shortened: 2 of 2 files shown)


kopylovanat:
DPCPP version need a bump for CI to pass.

tests/device_selector/device_selector_aspect.cpp
```diff
@@ -395,6 +396,49 @@ constexpr void check_for_random_set(const named_type_pack<AspectsT...>&) {
...
+        any_of(devices.begin(), devices.end(), exec_device_has_aspect);
+    bool all_devices =
+        all_of(devices.begin(), devices.end(), exec_device_has_aspect);
+    if (any_device) {
+      INFO(
+          "Check that if some device has aspect A, "
+          "any_device_has_v<A> is true.");
+      CHECK(std::is_base_of_v<std::true_type, sycl::any_device_has<aspect>>);
+      CHECK(sycl::any_device_has_v<aspect>);
+    }
```
bader, line 430: `else` branch is missing, but I think we can simplify the code by checking that `sycl::any_device_has_v<aspect>` value matches `any_device`.

```c++
CHECK(sycl::any_device_has_v<aspect> == any_device);
```

The same applies to `all_devices`.

kopylovanat: The `else` case is not there because according to Spec "The trait any_device_has<aspect> inherits from std::true_type if the compilation environment supports any device which has the specified aspect" but it is possible that the compilation environment could support some devices that are not present on the system where test is run. So even when any_device is `false` we can't be sure that there are no such devices supported by the compilation environment and that any_device_has_v is supposed to be `false`. As discussed here https://github.com/KhronosGroup/SYCL-CTS/pull/539#discussion_r1114877847.
Same with `all_devices`.

Added comments in test to clarify that.

bader: I need to confess; I hadn't read the spec before I reviewed this patch the first time. Now I think this test doesn't work as intended.
1. **Coverage**. The spec says "The trait any_device_has<aspect> inherits from std::true_type if the compilation environment supports any device which has the specified aspect, and it inherits from std::false_type if no device has the aspect.". The test attempts to check the first part of this statement, but not the second one and I still believe that we can check the second part too.
2. **Compile-time traits vs runtime**. Line 419 returns **_all_** devices the test finds in the system at runtime, but it can be a superset of device we compiled the test for. Let's say we compiled for a x86 CPU device but run the test on a system with x86 CPU and NVIDIA GPU device. Even if we run the test on CPU device, the runtime query `any_device` will be `true` for the aspects supported by GPU device, but not supported by CPU device. This is a bug in the test. I think we can fix it by quiring the devices associated with `kernel_bundle` rather than all available devices.
3. **JIT compilation**. How is this supposed to work for "runtime" compilation environments? I think by "compilation environment" specification means compile-time when we build the test and at the same time specification allows implementations to use "online compilation". When implementation supports online compilation, the device might be unknown in "compilation environment". For example, DPC++ uses SPIR-V (i.e. intermediate language) target with all possible aspects enabled by default, so I would expect all compile-time traits will return `true` (tagging @AlexeySachkov and @steffenlarsen to confirm). I see no way for runtime query to return a "virtual" device supporting all aspects.

steffenlarsen: 1. It may be possible to check the second part, but it would be the other way around from the current tests. I imagine something like "if `any_device_has<aspect>` inherits from `std::false_type`, check that all devices return `false` for `has(aspect)`". Is that what you had in mind?
2. I believe that `sycl::is_compatible` was intended to address this, but reading the specification it does seem like "compatibility" only considers the aspects and `device_has` attributes (as well as builtin kernel ownership). Should this also consider the compilation environment (tag @gmlueck)? Otherwise, I agree we would need to use `kernel_bundle`, but I am not even convinced it necessarily makes any stronger guarantees than `is_compatible`.
3. These traits are arguably not very useful for more open compilation environments, such as the SPIR-V JIT case. Depending on the environment, you may be able to make some guarantees, but in the most general case you would have `any_device_has` being trivially true and `all_devices_have` being trivially false.

test_plans/any_device_has_all_devices_have.asciidoc
```diff
@@ -34,8 +34,14 @@ Tests described below are performed for following aspects:
...
-and `all_devices_have<A>` inherits from `std::false_type`.
+* The test defines an empty kernel K as a mechanism to determine which devices are supported by the compilation environment.
+* At runtime, iterate over all devices returned by device::get_devices. For each device `D`:
+** Check if `is_compatible<K>(D)` is `true`. Since kernel `K` is empty, it should be compatible with any device that is supported by the compilation environment.
+** For each aspect `A` check if `D.has_aspect(A)` is `true`.
+** If `(is_compatible<K>(D) && D.has(A))` is `true`, then `any_device_has_v<A>` must be `true` and `any_device_has<A>` must inherit from `std::true_type`.
+** If `any_device_has_v<A>` is `false` then `(is_compatible<K>(D) && D.has(A))` must be `false`.
+** If `(is_compatible<K>(D) && !D.has(A))` is `true`, then `all_devices_have_v<A>` must be `false`
+and `all_devices_have<A>` must inherit from `std::false_type`.
+** If `all_devices_have_v<A>` is `true` then `(is_compatible<K>(D) && D.has(A))` or `(!is_compatible<K>(D) && !D.has(A))` must be `true`.
```
gmlueck, line 45:
```suggestion
** If `all_devices_have_v<A>` is `true` then `(D.has(A) || !is_compatible<K>(D))` must be `true`.
```

I think this is more correct, and it is a clearer way of thinking about it.

kopylovanat: Thank you! Fixed.

gmlueck (approved):
Test plan looks good.

bader (approved):
Thanks!
That make sense to me now assuming that behavior of `sycl::is_compatible` is changed according to [the comment from Greg](https://github.com/KhronosGroup/SYCL-CTS/pull/553#discussion_r1134001556).

tests/device_selector/device_selector_aspect.cpp
gmlueck: > I agree that we should make a spec clarification saying that `is_compatible` returns `false` when the translation unit containing the kernel does not support the given device. I can make this clarification.

I created this PR: KhronosGroup/SYCL-Docs#381

ProGTX (approved):

PR #554, opened 2023-03-07
Fix fill_init_values for spec_constants tests

steffenlarsen:
The specializations constant tests use fill_init_values to initialize reference values that are used as expected values when comparing with the default values of specialization constants. However, the current version of fill_init_values fills the individual elements of vectors and marrays with incrementing values, which do not correspond to the default values used for the specialization constants of these sizes. This commit changes them to use the same initial values.


bader (approved):

keryell (approved):
Thanks.

PR #557, opened 2023-03-09
Re-worked scan group tests to compare to the results of std::exclusive_scan and std::inclusive_scan

kopylovanat:
Re-worked tests to use results of std::exclusive_scan and std::inclusive_scan as reference values.
Refactor code to reduce repeating of similar lines.

### DIFF
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -134,15 +134,8 @@ void check_scan(sycl::queue& queue, size_t size,
   std::vector<U> reference_e(size, U(-1));
   std::vector<U> reference_i(size, U(-1));
 
-// FIXME: known_identity is not implemented in hipSYCL yet
-#if SYCL_CTS_COMPILING_WITH_HIPSYCL
-  I init_value = (with_init) ? I(init)
-                 : (std::is_same_v<OpT, sycl::plus<U>>)
-                     ? I{}
-                     : std::numeric_limits<I>::lowest();
-#else
   I init_value = (with_init) ? I(init) : sycl::known_identity<OpT, I>::value;
-#endif
+
   std::exclusive_scan(v.begin(), v.end(), reference_e.begin(), init_value, op);
   std::inclusive_scan(v.begin(), v.end(), reference_i.begin(), op, init_value);
   for (int i = 0; i < size; i++) {
@@ -378,15 +371,8 @@ void check_scan_over_group(sycl::queue& queue, sycl::range<D> range, OpT op) {
   CHECK(ret_type_e);
   CHECK(ret_type_i);
 
-// FIXME: known_identity is not implemented in hipSYCL yet
-#if SYCL_CTS_COMPILING_WITH_HIPSYCL
-  T init_value = (with_init) ? T(init)
-                 : (std::is_same_v<OpT, sycl::plus<T>>)
-                     ? T{}
-                     : std::numeric_limits<T>::lowest();
-#else
   T init_value = (with_init) ? T(init) : sycl::known_identity<OpT, T>::value;
-#endif
+
   for (int i = 0; i < range_size; i++) {
     int shift = i - local_id[i];
     ;
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -37,20 +37,22 @@ Group get_group(const sycl::nd_item<D>& item) {
     return item.get_group();
 }
 
-template <typename I, typename T, typename U, typename Group, typename OpT>
+template <bool with_init, typename I, typename T, typename U, typename Group,
+          typename OpT>
 auto joint_inclusive_scan_helper(Group group, T* v_begin, T* v_end,
-                                 U* r_i_begin, OpT op, bool with_init) {
-  if (with_init) {
+                                 U* r_i_begin, OpT op) {
+  if constexpr (with_init) {
     return sycl::joint_inclusive_scan(group, v_begin, v_end, r_i_begin, op,
                                       I(init));
   } else
     return sycl::joint_inclusive_scan(group, v_begin, v_end, r_i_begin, op);
 }
 
-template <typename I, typename T, typename U, typename Group, typename OpT>
+template <bool with_init, typename I, typename T, typename U, typename Group,
+          typename OpT>
 auto joint_exclusive_scan_helper(Group group, T* v_begin, T* v_end,
-                                 U* r_e_begin, OpT op, bool with_init) {
-  if (with_init) {
+                                 U* r_e_begin, OpT op) {
+  if constexpr (with_init) {
     return sycl::joint_exclusive_scan(group, v_begin, v_end, r_e_begin, I(init),
                                       op);
   } else
@@ -111,12 +113,12 @@ void check_scan(sycl::queue& queue, size_t size,
                 U* r_e_begin = res_e_acc.get_pointer();
                 U* r_i_begin = res_i_acc.get_pointer();
 
-                auto r_e_end = joint_exclusive_scan_helper<I>(
-                    group, v_begin, v_end, r_e_begin, op, with_init);
+                auto r_e_end = joint_exclusive_scan_helper<with_init, I>(
+                    group, v_begin, v_end, r_e_begin, op);
                 ret_type_e_acc[0] = std::is_same_v<U*, decltype(r_e_end)>;
 
-                auto r_i_end = joint_inclusive_scan_helper<I>(
-                    group, v_begin, v_end, r_i_begin, op, with_init);
+                auto r_i_end = joint_inclusive_scan_helper<with_init, I>(
+                    group, v_begin, v_end, r_i_begin, op);
                 ret_type_i_acc[0] = std::is_same_v<U*, decltype(r_i_end)>;
 
                 end_e_acc[0] = (r_e_begin + res_e_acc.size() == r_e_end);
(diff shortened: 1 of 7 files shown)


util/accuracy.h
```diff
@@ -49,4 +49,22 @@ inline sycl::half get_ulp_sycl<sycl::half>(sycl::half x) {
   return static_cast<sycl::half>(ulp * multiplier);
 }
 
+template <typename T>
+bool compare_with_ulp(T actual, T expected, unsigned int ulpsExpected) {
+  if (actual == expected)
+    return true;
+  else if constexpr (!std::is_integral_v<T>) {
```
bader, line 56:
```suggestion
    return true;
  if constexpr (!std::is_integral_v<T>) {
```

util/accuracy.h
```diff
@@ -49,4 +49,22 @@ inline sycl::half get_ulp_sycl<sycl::half>(sycl::half x) {
...
+    const T difference = static_cast<T>(std::fabs(actual - expected));
+    const T differenceExpected = ulpsExpected * get_ulp_sycl(expected);
```
bader, line 58: Could you add a comment about what `compare_with_ulp` is supposed to do, please?

tests/group_functions/group_scan.h
```diff
@@ -513,163 +290,173 @@ class invoke_init_joint_scan_group_same_type {
...
+                                    T init) {
+  return sycl::inclusive_scan_over_group(g, x, init, binary_op);
+}
+#else
+template <typename Group, typename V, typename T, typename BinaryOperation>
+auto inclusive_scan_over_group_impl(Group g, V x, BinaryOperation binary_op,
+                                    T init) {
+  return sycl::inclusive_scan_over_group(g, x, binary_op, init);
+}
+#endif
```
bader, line 310: CTS is supposed to validate that implementation conforms to the specification. We can't adjust to implementation bugs here. Please, keep only standard version and remove all non-standard code.

bader (changes requested):

tests/group_functions/group_scan.h
kopylovanat: Removed this workaround and disable scan_over_group tests with init for hipSYCL.
Kept workaround for known_identity for hipSYCL, because it's not what is tested here and it's not implemented yet.

tests/group_functions/group_scan.h
```diff
@@ -20,11 +20,149 @@
...
+
+// FIXME: known_identity is not implemented in hipSYCL yet
+#if SYCL_CTS_COMPILING_WITH_HIPSYCL
+  I init_value = (with_init) ? I(init)
+                 : (std::is_same_v<OpT, sycl::plus<U>>)
+                     ? I{}
+                     : std::numeric_limits<I>::lowest();
+#else
+  I init_value = (with_init) ? I(init) : sycl::known_identity<OpT, I>::value;
+#endif
```
bader, line 145: This is another place where I think we should remove `#if SYCL_CTS_COMPILING_WITH_HIPSYCL`.
It's reasonable to require `sycl::known_identity` support for all implementations.

tests/group_functions/group_scan.h
```diff
@@ -20,11 +20,149 @@
...
+  } else
+    return sycl::joint_exclusive_scan(group, v_begin, v_end, r_e_begin, op);
+}
+
+template <int D, typename T, typename U, typename Group, bool with_init,
+          typename I = U, typename OpT>
+void check_scan(sycl::queue& queue, size_t size,
+                sycl::nd_range<D> executionRange, OpT op) {
+  std::vector<T> v(size);
+  std::iota(v.begin(), v.end(), T(1));
```
bader, line 67: I suggest we fill input buffer with random numbers and print a seed to reproduce the input.

tests/group_functions/group_scan.h
```diff
@@ -20,11 +20,149 @@
...
+#endif
+  std::exclusive_scan(v.begin(), v.end(), reference_e.begin(), init_value, op);
+  std::inclusive_scan(v.begin(), v.end(), reference_i.begin(), op, init_value);
+  for (int i = 0; i < size; i++) {
+    {
+      INFO("Check joint_exclusive_scan for element " + std::to_string(i));
+      INFO("Result: " + std::to_string(res_e[i]));
+      INFO("Expected: " + std::to_string(reference_e[i]));
+      // ulp is i because there can be different rounding mods
+      CHECK(is_equal_with_ulp(res_e[i], reference_e[i], i));
```
bader, line 154: This check is not correct. I suppose you add `is_equal_with_ulp` to compare floating point values, but floating point addition is not associative and result is non-deterministic.
Consider following input: {10^9, -10^9, 10^{-9}}. The last element of output can be either 0 or 10^{-9}. I can't provide an input to create arbitrary ULP difference.

Probably we can add verification only for pre-defined input and my proposal with using random input doesn't fly.

What is the size of buffer here? Up to some limit the expected ULP must be 0 as there should be no rounding errors to represent numbers 1, 2, 3, ... Using `size` for allowed error in ULP seems too loose.

util/accuracy.h
```diff
@@ -49,4 +51,27 @@ inline sycl::half get_ulp_sycl<sycl::half>(sycl::half x) {
...
+/**
+ * @brief Uses provided allowed difference in ULPs to calculate max allowed
+ *        difference between actual and expected values to consider them equal.
+ *        Returns whether the values are equal by such definition.
+ *        If type T is integral, result of exact comparing is returned.
+ */
+template <typename T>
+bool is_equal_with_ulp(T actual, T expected, unsigned int ulpsExpected) {
+  if (actual == expected) return true;
+  if constexpr (!std::is_integral_v<T>) {
```
bader, line 63: Non-negative conditions are easier to read.
```suggestion
  if constexpr (std::is_integral_v<T>) return false;
```

tests/group_functions/group_scan.h
bader: > Kept workaround for known_identity for hipSYCL, because it's not what is tested here and it's not implemented yet.

I suggest we either use `known_identity` for all implementations or don't use it at all.

tests/group_functions/group_scan.h
kopylovanat: In this test with size of following values: {5, work_group_size / 2, 3 * work_group_size}, where work_group_size is max size of work-group supported by device. For scan_over_group tests size is work_group_size.
Should we use some fixed number instead and remove comparing with ulp?

tests/group_functions/group_scan.h
kopylovanat: Removed workaround for known_identity and disable tests for hipSYCL.

tests/group_functions/group_scan.h
kopylovanat: Removed workaround for known_identity and disable tests for hipSYCL.

tests/group_functions/group_scan.h
bader: AFAIK, some devices support work-group size > 1000. I don't see much difference between 10 (27 == 3x3x3) and 1000 elements if we to test scan functionality. I would use minimal work-group size required for functional testing and check that tested device supports that minimal size.

With regard to validation of floating point variant, I think we have to use pre-defined input to keep the verification code simple. According to my understanding, using `is_equal_with_ulp` for pre-defined input is unnecessary, but it can be used. We just need to be more accurate with reference values. In opinion, current values are too loose.

kopylovanat: Set max size for work-group as 12 using functionality of helper function util::work_group_range. With restricted size and pre-defined values 1, 2, 3, ... there are no rounding errors and comparing with ulp is unnecessary - it is removed.

util/accuracy.h
```diff
@@ -49,4 +51,28 @@ inline sycl::half get_ulp_sycl<sycl::half>(sycl::half x) {
...
+ *        difference between actual and expected values to consider them equal.
+ *        Returns whether the values are equal by such definition.
+ *        If type T is integral, result of exact comparing is returned.
+ */
+template <typename T>
+bool is_equal_with_ulp(T actual, T expected, unsigned int ulpsExpected) {
+  if (actual == expected) return true;
+  if constexpr (std::is_integral_v<T>)
+    return false;
+  else {
```
bader, line 65: Do we need `else`-branch or we can drop it?

util/accuracy.h
```diff
@@ -49,4 +51,28 @@ inline sycl::half get_ulp_sycl<sycl::half>(sycl::half x) {
...
 
-#endif // __SYCLCTS_UTIL_ACCURACY_H
+/**
+ * @brief Uses provided allowed difference in ULPs to calculate max allowed
+ *        difference between actual and expected values to consider them equal.
+ *        Returns whether the values are equal by such definition.
+ *        If type T is integral, result of exact comparing is returned.
+ */
+template <typename T>
+bool is_equal_with_ulp(T actual, T expected, unsigned int ulpsExpected) {
```
bader, line 61: Do we still need this template?

tests/group_functions/group_scan.h
```diff
@@ -515,164 +285,151 @@ class invoke_init_joint_scan_group_same_type {
...
+        .wait_and_throw();
+  }
+  CHECK(ret_type_e);
+  CHECK(ret_type_i);
+
+  T init_value = (with_init) ? T(init) : sycl::known_identity<OpT, T>::value;
+
+  for (int i = 0; i < range_size; i++) {
+    int shift = i - local_id[i];
+    ;
```
bader, line 379:
```suggestion
    int shift = i - local_id[i];
```

bader (approved):
LGTM, just a few minor comments.

util/accuracy.h
kopylovanat: No, removed.

ProGTX (approved):

keryell (approved):

PR #558, opened 2023-03-09
Separate test cases for pointers types in atomic_ref tests

vladimirkhashev:
Made separate test cases for pointers types similar to `atomic64` types because size of pointers types can be 64 bits wide and could be not supported by device and to make sure that tests for pointers to `atomic64` types will be tested even if `atomic64` aspect is not supported by device and size of pointers types is less than 64 bits. Tests for pointers types will be performed if size of pointers types is less than 64 bits or if device supports `atomic64` aspect.


bader (approved):

keryell (approved):
Thanks.

PR #560, opened 2023-03-13
Fix math_builtin_api tests

aleksmesh:
Fixed bugs in python module and math functions for marray in util.

### DIFF
--- a/tests/math_builtin_api/math_builtin.h
+++ b/tests/math_builtin_api/math_builtin.h
@@ -108,7 +108,7 @@ verify(sycl_cts::util::logger &log, T value, sycl_cts::resultRef<T> r,
 }
 
 template <typename T>
-typename std::enable_if<std::is_integral<T>::value, bool>::type verify(
+typename std::enable_if_t<std::is_integral_v<T>, bool> verify(
     sycl_cts::util::logger &log, T value, sycl_cts::resultRef<T> r, int,
     const std::string &) {
   bool result = value == r.res || !r.undefined.empty();
--- a/tests/math_builtin_api/math_builtin.h
+++ b/tests/math_builtin_api/math_builtin.h
@@ -66,14 +66,14 @@ template <> inline sycl::half min_t<sycl::half>() {
 
 template <typename T>
 bool verify(sycl_cts::util::logger &log, T a, T b, int accuracy,
-            std::string comment);
+            const std::string &comment);
 
 template <typename T>
 typename std::enable_if<std::is_floating_point<T>::value ||
                             std::is_same<sycl::half, T>::value,
                         bool>::type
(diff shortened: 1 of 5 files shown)


tests/math_builtin_api/math_builtin.h
```diff
@@ -130,16 +131,28 @@ bool verify(sycl_cts::util::logger &log, sycl::vec<T, N> a,
...
-            std::string comment) {
+            const std::string &comment) {
   return verify(log, a, sycl_cts::resultRef<T>(b), accuracy, comment);
 }
 
 template <int N, typename returnT, typename funT>
 void check_function(sycl_cts::util::logger &log, funT fun,
                     sycl_cts::resultRef<returnT> ref, int accuracy = 0,
-                    std::string comment = "") {
+                    const std::string &comment = "") {
```
keryell, line 155: While you are here :-)
```suggestion
                    const std::string &comment = {}) {
```
You do not want every time to measure this empty C string and then construct a `std::string` from it...

tests/math_builtin_api/math_builtin.h
```diff
@@ -166,9 +179,8 @@ void check_function(sycl_cts::util::logger &log, funT fun,
 template <int N, typename returnT, typename funT, typename argT>
 void check_function_multi_ptr_private(sycl_cts::util::logger &log, funT fun,
                                       sycl_cts::resultRef<returnT> ref,
-                                      argT ptrRef,
-                                      int accuracy = 0,
-                                      std::string comment = "") {
+                                      argT ptrRef, int accuracy = 0,
+                                      const std::string &comment = "") {
```
keryell, line 183:
```suggestion
                                      const std::string &comment = {) {
```
everywhere.

aleksmesh: Fixed it. Thanks!

tests/math_builtin_api/math_builtin.h
aleksmesh: Done.

tests/math_builtin_api/math_builtin.h
```diff
@@ -102,15 +102,15 @@ verify(sycl_cts::util::logger &log, T value, sycl_cts::resultRef<T> r, int accur
...
-    msg += ", " + comment;
+  if (!comment.empty()) msg += ", " + comment;
   log.note(msg);
   return false;
 }
 
 template <typename T>
-typename std::enable_if<std::is_integral<T>::value, bool>::type
-verify(sycl_cts::util::logger &log, T value, sycl_cts::resultRef<T> r, int, std::string) {
+typename std::enable_if<std::is_integral<T>::value, bool>::type verify(
```
keryell, line 111: Since you are here :-)
```suggestion
typename std::enable_if_t<std::is_integral_v<T>, bool> verify(
```

keryell (approved):
Thanks.

PR #569, opened 2023-03-22
Extend sub_group_mask extension test plan

aleksmesh:
Cover gaps in sub_group_mask extension test plan according to updated version of specification [sycl_ext_oneapi_sub_group_mask](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_sub_group_mask.asciidoc).

### DIFF
--- a/test_plans/sub_group_mask.asciidoc
+++ b/test_plans/sub_group_mask.asciidoc
@@ -4,7 +4,7 @@
 = Test plan for sub_group_mask
 
 This is a test plan for the APIs described in
-https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/GroupMask/GroupMask.asciidoc[GroupMask.asciidoc]
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_sub_group_mask.asciidoc[sycl_ext_oneapi_sub_group_mask]
 
 
 == Testing scope
--- a/test_plans/sub_group_mask.asciidoc
+++ b/test_plans/sub_group_mask.asciidoc
@@ -40,7 +40,6 @@ All of the following tests have these initial steps performed in defferent .cpp 
   `device.get_info<sycl::info::device::sub_group_sizes>()` should be used
   to check supported `sub_group` sizes.
 * In lambda function use `nd_item::get_sub_group()` to get `sycl::sub_group` instance.
-* Use `sub_group_ballot` to get `sub_group_mask` instance with predicate suitable for test.
 * Use accessor to buffer for type `bool` to save results.
 * For regular test run check is performed only for leader work-item
 * For full conformance test run check is performed for all work-items.
@@ -85,9 +84,9 @@ and `dim` is `2`, `5`, and `10`.
 
 For `size_t value = [item.get_local_linear_id(), 0b1010...]` and `marray<T, dim> val`:
 * If `CHAR_BIT * sizeof(val) < CHAR_BIT * sizeof(value)`
-  for `N = CHAR_BIT * sizeof(val)` set `N` bits of `val` to `N` bits of `value`.
+  for `N = CHAR_BIT * sizeof(val)` set first `N` bits of `val` to first `N` bits of `value`.
 * If `CHAR_BIT * sizeof(val) >= CHAR_BIT * sizeof(value)`
-  for `N = CHAR_BIT * sizeof(value)` set `N` bits of `val` to `N` bits of `value`.
+  for `N = CHAR_BIT * sizeof(value)` set first `N` bits of `val` to first `N` bits of `value`.
 * Construct `sub_group_mask` instance `s` with `sub_group_mask(val)` ctor.
 * Check that for `N = 0...sub_group_mask.size() - 1` `s[id(N)]` == `((val >> N) & 1)`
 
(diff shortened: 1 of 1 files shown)


test_plans/sub_group_mask.asciidoc
```diff
@@ -36,13 +48,73 @@ All of the following tests have these initial steps performed in defferent .cpp
...
+==== sub_group_mask() [[default_ctor]]
+
+* Construct `sub_group_mask` instance `s` `with sub_group_mask()` ctor.
+* Check that `s.none() == true`.
+
+==== sub_group_mask(unsigned long long val) [[ull_ctor]]
+
+* Construct `sub_group_mask` instance `s` with `sub_group_mask(unsigned long long val)` ctor
+  and `size_t val` == `[item.get_local_linear_id(), 0b1010...]`.
+* Check that for `N = 0...sub_group_mask.size() - 1` `s[id(N)]` == `((val >> N) & 1)`
```
gmlueck, line 65: I'm sorry, but I don't understand this at all.  What does the notation with the square brackets mean:

> `size_t val` == `[item.get_local_linear_id(), 0b1010...]`

`val` is a scalar value, so I don't understand what the square brackets are supposed to signify.

aleksmesh: Corrected algorithm description.

test_plans/sub_group_mask.asciidoc
```diff
@@ -16,33 +16,106 @@ is selected on the CTS command line.
...
+
+==== sub_group_mask() [[default_ctor]]
+
+* Construct `sub_group_mask` instance `s` `with sub_group_mask()` ctor.
+* Check that `s.none() == true`.
+
+==== sub_group_mask(unsigned long long val) [[ull_ctor]]
+
+* Construct `sub_group_mask` instance `s` with `sub_group_mask(unsigned long long val)` ctor
+  and `size_t val`. Where `val` is `item.get_local_linear_id()` and  `0b1010...`.
```
gmlueck, line 64: Are you saying that each work-item constructs two `sub_group_mask` instances:

```
sb1 = sub_group_mask(item.get_local_linear_id());
sb2 = sub_group_mask(0b1010...);
```

?

test_plans/sub_group_mask.asciidoc
```diff
@@ -16,33 +16,106 @@ is selected on the CTS command line.
```
gmlueck, line 16: I noticed that this link above is broken;

> This is a test plan for the APIs described in
https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/GroupMask/GroupMask.asciidoc[GroupMask.asciidoc]

Can you fix the link and also change the text of the link to be "sycl_ext_oneapi_sub_group_mask" (which is the name of the extension)?

test_plans/sub_group_mask.asciidoc
gmlueck: I think this test is OK, but it would be clearer to explain it this way:

> Run the following test with `VALUE = item.get_local_linear_id()` and with `VALUE = 0b101010...`:
> ```
> size_t val = VALUE;
> sub_group_mask mask{val};
> for (size_t i = 0;  i < mask.size();  i++) {
>   if (mask[id(i)] != ((val >> i) & 1)) {
>     /* test fails */
>   }
> }
> ```

test_plans/sub_group_mask.asciidoc
```diff
@@ -16,33 +16,106 @@ is selected on the CTS command line.
...
+and `dim` is `2`, `5`, and `10`.
+
+* Construct `marray<T, dim>` instance `val` with default ctor.
+* Use initial variable `size_t value` with values `item.get_local_linear_id()` and `0b1010...`.
+* If `CHAR_BIT * sizeof(val) < CHAR_BIT * sizeof(value)`
+  for `N = CHAR_BIT * sizeof(val)` set first `N` bits of `val` to first `N` bits of `value`.
+* If `CHAR_BIT * sizeof(val) >= CHAR_BIT * sizeof(value)`
+  for `N = CHAR_BIT * sizeof(value)` set first `N` bits of `val` to first `N` bits of `value`.
+* Construct `sub_group_mask` instance `s` with `sub_group_mask(val)` ctor.
+* Check that for `N = 0...sub_group_mask.size() - 1` `s[id(N)]` == `((val >> N) & 1)`.
```
gmlueck, line 92: This is OK too, but I think it would be clearer like this:

> Run the following test with all combinations of:
> * TYPE: `char`, `signed char`, ....
> * DIM: `2`, `5`, `10`
> * VALUE: `item.get_local_linear_id()`, `0b101010...`
> ```
> marray<TYPE, DIM> ma;
> size_t val = VALUE;
> if (sizeof(ma) < sizeof(val)) {
>   int N = CHAR_BIT * sizeof(ma);
>   /* set first N bits of "ma" to first N bits of "val" */
> } else {
>   int N = CHAR_BIT * sizeof(val);
>   /* set first N  bits of "ma" to first N bits of "val" */
> }
> sub_group_mask mask{ma};
> for (size_t i = 0;  i < mask.size();  i++) {
>   if (mask[id(i)] != ((val >> i) & 1)) {
>     /* test fails */
>   }
> }
> ```

test_plans/sub_group_mask.asciidoc
aleksmesh: Thanks for the clarification! Fixed it.

test_plans/sub_group_mask.asciidoc
aleksmesh: Fixed it. Thanks!

gmlueck (approved):

PR #571, opened 2023-03-22
Cover gaps in accessor constructor tests

aleksmesh:
Tests for sycl::host_accessor and sycl::accessor constructors with zero-length buffer added.
sycl::host_accessor constructors tests enabled.
Split file with sycl::accessor constructors tests into separate files for different constructor
types due to the fact that compiling a single file requires a huge amount of RAM.


tests/accessor/accessor_common.h
```diff
@@ -1150,7 +1160,7 @@ void check_linearization() {
 
             if constexpr (Target == sycl::target::device) {
               sycl::accessor res_acc(res_buf, cgh);
-              cgh.single_task<kname>([=]() {
+              cgh.single_task([=]() {
```
keryell, line 1163: While you are here
```suggestion
              cgh.single_task([=] {
```

keryell (approved):
Thanks.

PR #572, opened 2023-03-22
Add checks for exception for placeholder accessor for different memory operations

kopylovanat:
Cover gaps that only parallel_for_work_group was checked on exception when placeholder accessor is passed as an argument to a command.


keryell (approved):
Thanks.

Issue #532, opened 2023-02-17
Test that CGF is invoked synchronously within same thread as call to `sycl::queue`


psalz:
This was just clarified in https://github.com/KhronosGroup/SYCL-Docs/pull/354.

Something along the lines of

```c++
std::thread::id cgf_tid;
queue.submit([&](sycl::handler& cgh) {
	cgf_tid = std::this_thread::get_id();
	cgh.host_task([] {});
});
CHECK(cgf_tid == std::this_thread::get_id());
```

vladimirkhashev:
Created PR: #575

PR #575, opened 2023-03-27
Add checks that sycl::queue::submit() is invoked in the calling thread

vladimirkhashev:
This PR is to implement suggestions from this issue: #532


keryell (approved):
Thanks.

PR #577, opened 2023-03-29
Fix type deduction of reducer class instance in "reducer class" test

vladimirkhashev:
Fix wrong type deduction of reducer class instance by adding std::remove_reference_t


keryell (approved):
Good catch.

PR #578, opened 2023-03-29
Enable tests with sycl::aspect::emulated

kopylovanat:
Enable tests with sycl::aspect::emulated for DPCPP because issue https://github.com/intel/llvm/issues/8324 were resolved.

Also enable usage of target::host_task  for DPCPP in the same file.

### DIFF
--- a/tests/common/string_makers.h
+++ b/tests/common/string_makers.h
@@ -62,9 +62,7 @@ struct StringMaker<sycl::target> {
       case type::device:
         return "target::device";
 // FIXME: re-enable when target::host_task is implemented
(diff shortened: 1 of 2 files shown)


keryell (approved):

keryell:
I guess the title of the PR is a bit misleading since there is not only `aspect::emulated` which is enabled.
There is also `target::host_task` which is enabled.

kopylovanat:
> I guess the title of the PR is a bit misleading since there is not only `aspect::emulated` which is enabled. There is also `target::host_task` which is enabled.

You are right, fixed title.

PR #580, opened 2023-03-29
Fix host_task tests

kopylovanat:
Add expected template parameter for interop functions.
Replace `target::host_buffer` with `sycl::target::host_task`.


keryell (approved):
Thanks.

PR #581, opened 2023-03-29
Enable tests for sycl::joint_reduce

kopylovanat:
Enable tests for sycl::joint_reduce for DPCPP because issue https://github.com/intel/llvm/issues/8348 is resolved.


keryell (approved):
Thanks.

PR #586, opened 2023-03-30
Add large range tests

TApplencourt:
This PR adds 2 more tests, which verify that we can run "arbitrary" large size `parallel_for` with range.

In theory, we should try to submit a `std::numeric_limits<size_t>::max()`, but this will take some time. So using something relatively large.


tests/range/range_api.cpp
```diff
@@ -212,6 +212,14 @@ class test_range {
...
 };
 
+void test_empty_kernel_1d_range(size_t N, sycl::queue q) {
+  int *a = sycl::malloc_shared<int>(q, 1);
+  a[0] = 0;
+  q.parallel_for(sycl::range<1>(N), [=](auto i) { a[0] = 1; }).wait_and_throw();
+  CHECK_VALUE_SCALAR(log, a[0], 1);
+  sycl::free(a, q);
+}
+
```
bader, line 222:
```suggestion
void test_empty_kernel_1d_range(size_t N, sycl::queue q) {
  q.parallel_for(sycl::range<1>(N), [=](auto i) { }).wait_and_throw();
}
```
The name of the function is misleading. The kernel is not empty. 
What the intent here? I don't think current test gives valuable information.

tests/range/range_api.cpp
```diff
@@ -254,6 +262,14 @@ class TEST_NAME : public util::test_base {
         test2d(log, range_2d_g, range_2d_l, my_queue);
         test_range<3> test3d;
         test3d(log, range_3d_g, range_3d_l, my_queue);
+
+        // 32 bits, it's trivial. Sanity test.
```
bader, line 266: sizeof(unsigned int) is not guaranteed to be 32 bits. I suggest using unit32_t.

tests/range/range_api.cpp
```diff
@@ -254,6 +262,14 @@ class TEST_NAME : public util::test_base {
...
+        test_empty_kernel_1d_range(std::numeric_limits<unsigned int>::max(),
+                                   my_queue);
+        // Most GPU hardware have limitation to 32bits * 1024 for their native
+        // API so let's try more than that
+        test_empty_kernel_1d_range(
```
bader, line 271: I suggest we put these checks under `#if SYCL_CTS_ENABLE_FULL_CONFORMANCE`.

bader (commented):
Although it's not required for this patch, I think we should add `-fno-sycl-id-queries-fit-in-int` flag for DPC++ just to avoid potential issues in the future.

tests/range/range_api.cpp
TApplencourt: Good catch. Before, It was an empty kernel, and then I think more and found this trick... Will fix the name, sorry!

bader: NOTE: "this trick" tells you that at least one work-item was executed. Don't you expect all N to be executed? It's not checked by this kernel.

tests/range/range_api.cpp
TApplencourt: Good call. The doc C++ seems to imply that `std::numeric_limits` didn't took `unit32_t` but I just checked and it worked. So will switch to that.

tests/range/range_api.cpp
bader: Why do you prefer `q.parallel_for(sycl::range<1>(N), [=](auto i) { a[0] = 1; }).wait_and_throw();` over an empty kernel?

TApplencourt: I was afraid that a smart compiler could optimize away an empty kernel submission...  Where I have more hope that t no the compiler will be able to optimize away `N-1` work-item submission.

tests/range/range_api.cpp
```diff
@@ -212,6 +212,20 @@ class test_range {
...
+// This tests is only here to check if one can submit a large kernel.
+// We don't use an `empty kernel` to avoid smart compiler
+// optimizing the submision away
+// This test:
+//    - Doesn't check that all the work-item have be submited
+//    - It's technicaly UB (concurent write)
+void test_launch_kernel_1d_range(size_t N, sycl::queue q) {
+  int *a = sycl::malloc_shared<int>(q, 1);
+  a[0] = 0;
+  q.parallel_for(sycl::range<1>(N), [=](auto i) { a[0] = 1; }).wait_and_throw();
```
bader, line 224: Please, "name" the kernel. ComputeCPP doesn't support unnamed kernels.

bader (approved):

tests/range/range_api.cpp
TApplencourt: Is this not now part of the spec now? :( Will do! I guess I need to use template to avoid some double-definition errors in case of inlining.

tests/range/range_api.cpp
```diff
@@ -212,6 +212,20 @@ class test_range {
...
 };
 
+// This tests is only here to check if one can submit a large kernel.
+// We don't use an `empty kernel` to avoid smart compiler
+// optimizing the submision away
+// This test:
+//    - Doesn't check that all the work-item have be submited
+//    - It's technicaly UB (concurent write)
+void test_launch_kernel_1d_range(size_t N, sycl::queue q) {
+  int *a = sycl::malloc_shared<int>(q, 1);
```
bader, line 222:
```suggestion
  int *a = sycl::malloc_shared<int>(1, q);
```

tests/range/range_api.cpp
bader: I think we should also check that device supports that number of work-items (i.e. N <= info::device::max_work_item_sizes<1>).

bader: > Is this not now part of the spec now? :( Will do! I guess I need to use template to avoid some double-definition errors in case of inlining.

You are right. Technically, it's required unless the implementation supports only reduced feature-set. You can just disable the testing with ComputeCPP then.

TApplencourt: This is the trick `max_work_item_sizes` is only for the work items inside the work group of the nd_range.  It's unrelated to the absolute number of work items supported inside `range` used in `parralel_for`. Those are only bounded by `size_t` AFAIK.

tests/range/range_api.cpp
```diff
@@ -212,6 +212,20 @@ class test_range {
   }
 };
 
+// This tests is only here to check if one can submit a large kernel.
+// We don't use an `empty kernel` to avoid smart compiler
+// optimizing the submision away
+// This test:
+//    - Doesn't check that all the work-item have be submited
+//    - It's technicaly UB (concurent write)
```
keryell, line 220:
```suggestion
//    - It's technically UB (concurrent write)
```

tests/range/range_api.cpp
```diff
@@ -212,6 +212,20 @@ class test_range {
   }
 };
 
+// This tests is only here to check if one can submit a large kernel.
+// We don't use an `empty kernel` to avoid smart compiler
+// optimizing the submision away
```
keryell, line 217:
```suggestion
// optimizing the submission away
```

tests/range/range_api.cpp
```diff
@@ -212,6 +212,20 @@ class test_range {
...
+// This test:
+//    - Doesn't check that all the work-item have be submited
```
keryell, line 219:
```suggestion
//    - Doesn't check that all the work-items have be submitted
```

tests/range/range_api.cpp
```diff
@@ -254,6 +268,27 @@ class TEST_NAME : public util::test_base {
...
+        test_launch_kernel_1d_range(std::numeric_limits<uint32_t>::max(),
+                                    my_queue);
+
+        // Prime number bigger than UINT32_MAX
+        test_launch_kernel_1d_range(4294967311L, my_queue);
+
+        // Most GPU hardware have limitation to 32bits * 1024 for their native
+        // API so let's try more than that
+        test_launch_kernel_1d_range(
+            std::numeric_limits<uint32_t>::max() * 2024L, my_queue);
```
keryell, line 290: What is `std::numeric_limits<uint32_t>::max() * 2024L` on my Vax11/785 with its FP164 accelerator?

TApplencourt: :( This is a valid point. Those magic numbers are always ugly. One should use interopts, and for each backend, compute the maximum number of work items and then add `+1`. 

In the beginning, I was planning to use `std::numeric_limits<size_t>` but the kernel will be quite long to submit. As a tradeoff, I can replace `2024` with `max_work_item_sizes*2`? On a few hardware I tests `max_work_item_sizes` return 1024 :)

tests/range/range_api.cpp
```diff
@@ -254,6 +268,27 @@ class TEST_NAME : public util::test_base {
...
+
+// The tests are using unnamed lambda
+#if SYCL_CTS_ENABLE_FULL_CONFORMANCE && \
+    !defined(SYCL_CTS_COMPILING_WITH_COMPUTECPP)
+        // 32 bits, it's trivial. Sanity test.
+        test_launch_kernel_1d_range(std::numeric_limits<uint32_t>::max(),
+                                    my_queue);
+
+        // Prime number bigger than UINT32_MAX
+        test_launch_kernel_1d_range(4294967311L, my_queue);
```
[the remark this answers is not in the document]
TApplencourt: Correct. Will use `ull`

tests/range/range_api.cpp
```diff
@@ -254,6 +268,27 @@ class TEST_NAME : public util::test_base {
...
+
+// The tests are using unnamed lambda
+#if SYCL_CTS_ENABLE_FULL_CONFORMANCE && \
+    !defined(SYCL_CTS_COMPILING_WITH_COMPUTECPP)
+        // 32 bits, it's trivial. Sanity test.
+        test_launch_kernel_1d_range(std::numeric_limits<uint32_t>::max(),
+                                    my_queue);
+
+        // Prime number bigger than UINT32_MAX
+        test_launch_kernel_1d_range(4294967311ull, my_queue);
```
keryell, line 285: Help our eyes if you really love big numbers
```suggestion
        test_launch_kernel_1d_range(4'294'967'311ull, my_queue);
```

tests/range/range_api.cpp
TApplencourt: If I understand correctly we have two problem:
- uint32_t is optional so i cannot use it. Should I use `2^32-1` and call it a day or use `uint_least32_t` / `char32_t`?
- uint32_t * 20204L may overflow.  should I use `ull`?

Did I miss something else?

keryell: `std::uint32_t` is probably here in the SYCL world (and I prefer the `std::` version to the C one, by the way :-) ), so just use it.

So the problem is the second point. Thus yes we need `ULL` (I prefer the uppercase version to avoid code obfuscation like `11ull` vs `11ULL`).

keryell:
There is another issue, the case where `std::size_t` is 32 bits.

TApplencourt:
FPGA people and their weird system...  :) 
Fixed. I can more be paranoid and compute the minimum number of bits required for my tests to pass  (36 ?) if you prefer and do not assume either 32 or 64 bits size_t.

tests/range/range_api.cpp
```diff
@@ -254,6 +273,31 @@ class TEST_NAME : public util::test_base {
...
+        if (sizeof(size_t) > 4) {
+          // Prime number bigger than UINT32_MAX
+          test_launch_kernel_1d_range(4'294'967'311ULL, my_queue);
+
+          // Most GPU hardware have limitation to 32bits * 1024 for their native
+          // API so lets try more
+          test_launch_kernel_1d_range(
+              std::numeric_limits<uint32_t>::max() * 2024ULL, my_queue);
+        } else {
+          WARM("Some parameters not tested because size_t is 32bits");
```
keryell, line 298: I love the concept of WARM WARNing. :-)

tests/range/range_api.cpp
```diff
@@ -254,6 +273,31 @@ class TEST_NAME : public util::test_base {
...
+#endif
+
+// The tests are using unnamed lambda
+#if SYCL_CTS_ENABLE_FULL_CONFORMANCE && \
+    !defined(SYCL_CTS_COMPILING_WITH_COMPUTECPP)
+        // 32 bits, it's trivial. Sanity test.
+        test_launch_kernel_1d_range(std::numeric_limits<uint32_t>::max(),
+                                    my_queue);
+
+        if (sizeof(size_t) > 4) {
```
keryell, line 289:
```suggestion
        if constexpr (sizeof(size_t) > 4) {
```
and then no longer use `long long`.

tests/range/range_api.cpp
```diff
@@ -254,6 +273,31 @@ class TEST_NAME : public util::test_base {
...
+        if constexpr (sizeof(size_t) > 4) {
+          // Prime number bigger than UINT32_MAX
+          test_launch_kernel_1d_range(4'294'967'311ULL, my_queue);
+
+          // Most GPU hardware have limitation to 32bits * 1024 for their native
+          // API so lets try more
+          test_launch_kernel_1d_range(
+              std::numeric_limits<uint32_t>::max() * 2024ULL, my_queue);
+        } else {
+          WARN("Some parameters not tested because size_t is 32bits");
```
keryell, line 298: Would not be possible to test for example with `std::numeric_limits<int32_t>::max()` in that case?

keryell: Oh actually it is done already on line 286. :-) Never mind.

keryell (approved):

bader:
@TApplencourt, could you fix clang-format check, please?
See https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4691255728/attempts/1#summary-12727430805

tomdeakin:
@TApplencourt to run Clang Format, THEN ready to merge

TApplencourt:
Clang-format fixed! 
PS: Running `./clangformat.py` show a lot a difference. Maybe we need to do a big formatting PR at some point.

PR #590, opened 2023-04-04
Expand atomic_ref test plan to check atomicity

kopylovanat:
Atomic behavior was out of scope for initial atomic_ref test plan. In this PR checks for that are added.

### DIFF
--- a/test_plans/atomic_ref.asciidoc
+++ b/test_plans/atomic_ref.asciidoc
@@ -565,10 +565,13 @@ vals.fill(0);
 sycl::buffer buf{vals.data(), {GROUP_RANGE}};
 q.submit([&](sycl::handler &cgh) {
   sycl::accessor acc{buf, cgh};
+  sycl::local_accessor<TYPE> lacc{{1}, cgh}
   cgh.parallel_for(sycl::nd_range(GROUP_RANGE * LOCAL_SIZE, LOCAL_SIZE),
                     [=](auto item) {
-     sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{acc[item.group_id()]};
-     a_r.fetch_sub(2);
+     sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{lacc[0]};
+     if (a_r.fetch_sub(2) == -(LOCAL_SIZE*2)) {
+        acc[item.group_id()] = a_r.load();
+     }
   });
 });
 ----
@@ -599,39 +602,42 @@ These `ORDER` values which are supported by the device (according to `info::devi
 
 These `SPACE` values:
 
-* `access::address_space::global_space`
+* `access::address_space::local_space`
 * `access::address_space::generic_space`
 
 ==== Aquire and release
 
-This test uses a `parallel_for` with `nd_range` parameter to atomically update scalar value and release other changes that were made in one thread,
-and atomically load this scalar value and aquire these other changes in another thread within local group.
+This test uses a `parallel_for` with `nd_range` parameter to atomically update shared scalars in two threads within local group.
 
 [source,c++]
 ----
 const TYPE val chaged_val = 42;
-sycl::nd_range nd_range(GLOBAL_RANGE, 2);
+sycl::range group_range(GLOBAL_RANGE/ 2);
 std::array<bool, GLOBAL_RANGE / 2> res;
 vals.fill(false);
 sycl::buffer buf{res.data(), {GLOBAL_RANGE/2}};
 q.submit([&](sycl::handler &cgh) {
   sycl::accessor res_acc{buf, cgh};
-  sycl::local_accessor<TYPE> local_acc{cgh};
-  sycl::local_accessor<bool> flag{cgh};
-  cgh.parallel_for(nd_range, [=](auto item) {
-     sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{flag[0]};
-     if (item.global_id() % 2 == 0) {
-        /* long loop */
-        local_acc[0] = changed_val;
-        a_r.store(true);
-     } else {
-        while(flag.load()) {
-          res_acc[item.group_id()] = (local_acc[0] == changed_val);
-          break;
+  sycl::local_accessor<TYPE> x{cgh};
+  sycl::local_accessor<TYPE> y{cgh};
+  cgh.parallel_for_work_group(group_range, sycl::range(2),  [=](auto group) {
+    TYPE A = 0;
+    TYPE B = 0;
+    group.parallel_for_work_item([&](auto item) {
+      sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> refA{A};
+      sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> refB{B};
+      if (item.get_local_id() == 0) {
+          x[0] = refA.load();
+          refB.store(1);
+        } else {
+          y[0] = refB.load();
+          refA.store(1);
         }
-     }
+     });
+     res_acc[group.get_id()] = !(x[0] == 1 && y[0] == 1);
+    });
   });
-});
+
 ----
 
 Verify that all elements of res is `true` after the loop completes.
@@ -680,8 +686,8 @@ q.submit([&](sycl::handler &cgh) {
   cgh.parallel_for(nd_range, [=](auto item) {
      sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{local_acc[0]};
      arr_acc[item.local_id()] = true;
-     a_r.store(item.local_id(), memory_order::relaxed);
-     res_acc[item.global_id()] = (arr_acc[a_r.load()] == true);
+     a_r.store(item.local_id(), ORDER1);
+     res_acc[item.global_id()] = (arr_acc[a_r.load(ORDER2)] == true);
   });
 });
 ----
@@ -705,7 +711,14 @@ These `SCOPE` values which are supported by the device (according to `info::devi
 
 These `ORDER` values which are supported by the device (according to `info::device::atomic_memory_order_capabilities`):
 
+* `memory_order::relaxed`
+* `memory_order::acq_rel`
 * `memory_order::seq_cst`
+
+These `ORDER1` and `ORDER2` values which are supported by the device (according to `info::device::atomic_memory_order_capabilities`):
+
+* `ORDER1` is `memory_order::release` and `ORDER2` is `memory_order::acquire`
+* `ORDER1` and `ORDER2` are both `memory_order::seq_cst`
 
 These `SPACE` values:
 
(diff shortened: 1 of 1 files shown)


test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +507,137 @@ Check if it returns the original value of the referenced object. Check returned
...
+
+* `access::address_space::global_space`
+* `access::address_space::generic_space`
+
+Create `int` variable `val` and initialize it with `0`.
+Create `sycl::accessor acc` for `val` with `target` = `device`, `access_mode` = `read_write`.
+Call `parallel_for` with range<1> parameter
+Create `sycl::atomic_ref<int, order, scope, space> a_r` for `acc[0]`.
+Call `a_r.fetch_add(2)`.
+After kernel is executed check that `val` = `range.size() * 2`.
```
gmlueck, line 537: Can you rewrite these descriptions to show a small code sample that illustrates the core part of the test?  For example, this first one might be something like:

> This test uses a simple-form `parallel_for` to atomically update a single scalar value from all work-items:
>
> ```
> TYPE val{};
> sycl::buffer buf{&val, {1}};
> q.submit([&](sycl::handler &cgh) {
>   sycl::accessor acc{buf};
>   cgh.parallel_for({SIZE}, [=](int i) {
>      sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{acc[0]};
>      a_r.fetch_add(2);
>   });
> });
> ```
>
> Verify that `val == SIZE*2` after the loop completes.  (The result must be accurate within some small error range when `TYPE` is a floating point type.)
>
> This test is repeated for all combinations of supported values listed below:
>
> These `TYPE` types:
>
> * `int`
> * `long long` (if `sizeof(long long) == 8` then only if the device has `aspect::atomic64`)
> * `double` (only if the device has `aspect::fp64` and if `sizeof(double) == 8` only if the device has `aspect::atomic64`)
>
> These `SCOPE` values which are supported by the device (according to `info::device::atomic_memory_scope_capabilities`):
>
> * `memory_scope::device`
> * `memory_scope::system`
>
> These `ORDER` values which are supported by the device (according to `info::device::atomic_memory_order_capabilities`):
>
> * `memory_order::relaxed`
> * `memory_order::acq_rel`
> * `memory_order::seq_cst`
>
> These `SPACE` values:
>
> * `access::address_space::global_space`
> * `access::address_space::generic_space`

We used this type of test description in the [atomic_fence](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/atomic_fence.asciidoc) test plan, and I think it makes the plan much easier to understand.

Note that I also suggest some additional types for this test.  I think it's important to test a 64-bit type and a floating point type also.

kopylovanat: Rewrote test cases description to this format, added also type `float` to test floating point type even if fp64 is not supported.
Also changed how atomicity with respect to atomic operations in host code should be tested to correctly deal with pointers.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+size_t local_range = q.get_info<sycl::info::device::max_work_group_size>();
+sycl::nd_range nd_range(GLOBAL_RANGE, local_range);
+std::array<bool, GLOBAL_RANGE> res;
+vals.fill(false);
+sycl::buffer buf{res.data(), {GLOBAL_RANGE}};
+q.submit([&](sycl::handler &cgh) {
+  sycl::accessor res_acc{buf, cgh};
+  sycl::local_accessor<TYPE> local_acc{cgh};
+  sycl::local_accessor<bool> arr_acc{{local_range}, cgh};
+  cgh.parallel_for(nd_range, [=](auto item) {
```
gmlueck, line 686: The values in `arr_acc` are not guaranteed to be initialized to false when the work-group starts.  It might be good to ensure this by adding the following to the start of the kernel:

```
arr_acc[item.get_local_id())] = false;
```

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+     sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{local_acc[0]};
+     arr_acc[item.local_id()] = true;
+     a_r.store(item.local_id(), ORDER1);
+     res_acc[item.global_id()] = (arr_acc[a_r.load(ORDER2)] == true);
```
gmlueck, line 690:
```suggestion
     arr_acc[item.get_local_id()] = true;
     a_r.store(item.get_local_id(), ORDER1);
     res_acc[item.get_global_id()] = (arr_acc[a_r.load(ORDER2)] == true);
```

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+* `ORDER1` and `ORDER2` are both `memory_order::seq_cst`
+
+These `SPACE` values:
+
+* `access::address_space::local_space`
+* `access::address_space::generic_space`
+
+==== Atomicity with respect to atomic operations in host code
+
+This check is skipped if device doesn't support `aspect::usm_atomic_shared_allocations`.
```
gmlueck, line 730: The device also needs to support `memory_scope::system` according to `info::device::atomic_memory_scope_capabilities`.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+
+This check is skipped if sizeof(pointer) == 8 and device doesn't support `aspect::atomic64`.
```
gmlueck, line 732: This is not necessary.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+----
+const TYPE val chaged_val = 42;
+sycl::range group_range(GLOBAL_RANGE/ 2);
+std::array<bool, GLOBAL_RANGE / 2> res;
+vals.fill(false);
+sycl::buffer buf{res.data(), {GLOBAL_RANGE/2}};
+q.submit([&](sycl::handler &cgh) {
+  sycl::accessor res_acc{buf, cgh};
+  sycl::local_accessor<TYPE> x{cgh};
+  sycl::local_accessor<TYPE> y{cgh};
```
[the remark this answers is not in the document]
kopylovanat: Done.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+  cgh.parallel_for_work_group(group_range, sycl::range(2),  [=](auto group) {
```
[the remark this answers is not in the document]
kopylovanat: Re-worked test to use parallel_for.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+
+This test uses a `parallel_for` with `nd_range` parameter to atomically read scalar in one thread,
+and then aquire all changes from it in another thread within local group.
+
+[source,c++]
+----
+size_t local_range = q.get_info<sycl::info::device::max_work_group_size>();
+sycl::nd_range nd_range(GLOBAL_RANGE, local_range);
+std::array<bool, GLOBAL_RANGE> res;
+vals.fill(false);
```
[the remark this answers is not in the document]
kopylovanat: Fixed.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+sycl::buffer buf{res.data(), {GLOBAL_RANGE}};
+q.submit([&](sycl::handler &cgh) {
+  sycl::accessor res_acc{buf, cgh};
+  sycl::local_accessor<TYPE> local_acc{cgh};
```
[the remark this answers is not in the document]
kopylovanat: Done.

test_plans/atomic_ref.asciidoc
kopylovanat: Done.

test_plans/atomic_ref.asciidoc
kopylovanat: Done.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,264 @@ Check if it returns the original value of the referenced object. Check returned
...
+
+* `memory_scope::device`
+* `memory_scope::system`
+* `memory_scope::work_group`
+
+These `ORDER` values which are supported by the device (according to `info::device::atomic_memory_order_capabilities`):
+
+* `memory_order::relaxed`
+* `memory_order::acq_rel`
+* `memory_order::seq_cst`
```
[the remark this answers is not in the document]
kopylovanat: Changed to only use `memory_order::relaxed`.

test_plans/atomic_ref.asciidoc
kopylovanat: Added note about skipping check if `memory_scope::system` is not supported.

test_plans/atomic_ref.asciidoc
kopylovanat: Removed.

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,262 @@ Check if it returns the original value of the referenced object. Check returned
...
+    refA.store(0);
+    refB.store(0);
+    sycl::group_barrier(item.get_group());
+    if (item.get_local_id() == 0) {
+        x = refA.load();
+        refB.store(1);
+      } else {
+        y = refB.load();
+        refA.store(1);
+      }
```
gmlueck, line 637: I think you need a barrier after this line:

```
```

Otherwise, the code below could run before all work-items finish running the test.

kopylovanat: You are right, added.

gmlueck (approved):
Looks good!

test_plans/atomic_ref.asciidoc
```diff
@@ -505,4 +505,252 @@ Check if it returns the original value of the referenced object. Check returned
...
+std::atomic a{val_p};
+auto event = q.submit([&](sycl::handler &cgh) {
+  cgh.parallel_for({SIZE}, [=](auto i) {
+     sycl::atomic_ref<TYPE, ORDER, SCOPE, SPACE> a_r{val_p};
+     a_r++;
+  });
+});
+a++;
+event.wait();
+----
```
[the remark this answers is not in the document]
gmlueck: I think it is OK to use `std::atomic_ref` in the CTS, but the tests that rely on it should be protected with `#ifdef __cpp_lib_atomic_ref`.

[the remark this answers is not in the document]
kopylovanat: Thank you! Updated test plan here https://github.com/KhronosGroup/SYCL-CTS/pull/647

PR #591, opened 2023-04-04
Fix read_write_zero_dim_acc

jzc:
Based on `read_write_acc`, it looks like `read_write_zero_dim_acc` should be assigning `changed_val` to element that is accessed by the given accessor `testing_acc`, and not assigning `changed_val` to a local variable. (Or, just using `testing_acc` as the LHS of the assignment.)

### DIFF
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -485,7 +485,7 @@ void read_write_zero_dim_acc(AccT testing_acc, ResultAccT res_acc) {
     res_acc[0] = value_operations::are_equal(acc_ref, other_data);
   }
   if constexpr (AccessMode != sycl::access_mode::read) {
-    DataT acc_ref(testing_acc);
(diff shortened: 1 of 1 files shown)


tests/accessor/accessor_common.h
```diff
@@ -485,7 +485,7 @@ void read_write_zero_dim_acc(AccT testing_acc, ResultAccT res_acc) {
     res_acc[0] = value_operations::are_equal(acc_ref, other_data);
   }
   if constexpr (AccessMode != sycl::access_mode::read) {
```
bader, line 487: Shouldn't we apply the fix to the code in `AccessMode != sycl::access_mode::write` branch?

jzc: I didn't think it mattered in that branch because the initialization `acc_ref` would still read from the accessor and no reference is necessary for testing of equality.

bader (approved):

PR #593, opened 2023-04-05
Enabled accessor tests partially and fix bugs

aleksmesh:
Enabled accessor tests partially and fix bugs in accessor_api_common.h and accessor_implicit_conversions.h.


bader:
> The hosted runner: GitHub Actions 14 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

I think the root cause of this error is "CPU/Memory starvation" initiated by enabling compilation of accessor test cases. @aleksmesh, is it possible to mitigate this issue by reducing the amount of code in a single compilation unit (might not help if the problematic step is linking)?
Building in one thread might be helpful, but on the other hand it increases the build time and there is a risk to hit the time limit for a GitHub Action job.

It's worth raising this topic with DPC++ developers - https://github.com/intel/llvm/discussions.

aleksmesh:
> @aleksmesh, is it possible to mitigate this issue by reducing the amount of code in a single compilation unit (might not help if the problematic step is linking)? Building in one thread might be helpful, but on the other hand it increases the build time

I thought so, the problem is in the big source code files. Thanks for the advice!

aleksmesh:
@bader It seems that CI is now passing without a crash.

keryell (approved):
Thanks for the refactoring.

mmoadeli:
@bader any reason not to have this merged? It will help evaluating the status of accessor 2020 features.

PR #594, opened 2023-04-06
Test plan for sycl_ext_oneapi_enqueue_barrier extension 

vladimirkhashev:
Add test plan for `sycl_ext_oneapi_enqueue_barrier` extension according to the extension specification [link](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_enqueue_barrier.asciidoc)

### DIFF
--- a/test_plans/enqueue_barrier.asciidoc
+++ b/test_plans/enqueue_barrier.asciidoc
@@ -99,7 +99,7 @@ be equal to `sycl::event` for both member functions.
 Check that waiting on returned `event` waits for the previously submitted
 kernels to complete.
 
-Checking with one queue:
+==== Checking with one queue
 
 * Create a queue
 * Allocate two USM buffers in host memory BufA and BufB
@@ -110,7 +110,7 @@ Checking with one queue:
 * Wait on E from host code
 * Check the BufA and BufB contain the results of kernels A and B
 
-Checking with three queues:
+==== Checking with three queues
 
 * Create 3 queues
 * Allocate two USM buffers in host memory BufA and BufB


test_plans/enqueue_barrier.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+The `sycl_ext_oneapi_enqueue_barrier` extension defines the following member
+functions for the `sycl::queue` class:
+
+* `event ext_oneapi_submit_barrier()`
+* `event ext_oneapi_submit_barrier( const std::vector<event> &waitList )`
+
+These member functions should be checked the same as the corresponding member
+functions of the `sycl::handler` class according to the algorithm description
+in the previous section. In addition return type should be checked which should
+be equal to `sycl::event` for both member functions.
```
gmlueck, line 96: Can you add one more test to check the event returned from `ext_oneapi_submit_barrier`?  We should make sure that host code which waits on this event waits for the previously submitted kernels to complete.  Something like:

* Define kernel A which writes a result in USM buffer BufA.
* Define kernel B which writes a result in USM buffer BufB.
* Submit kernels A and B to a queue.
* Call `queue::ext_oneapi_submit_barrier`, which returns event E.
* Wait on E from host code.
* Make sure the BufA and BufB contain the results of kernels A and B.

It would be good to also have a version of this test that uses 3 queues, like the one you define above.

vladimirkhashev: Added.

keryell (approved):

tomdeakin:
One review OK for extensions. Ready to merge if @gmlueck 's comments are resolved.
Will re-review if promoted to main spec.

bader:
@gmlueck, could you check if your comment is addressed, please?

gmlueck (approved):

gmlueck:
> @gmlueck, could you check if your comment is addressed, please?

Looks good!

PR #599, opened 2023-04-17
Enable common reference semantics tests for local_accessor and host_accessor

kopylovanat:
Enable tests with std::hash for local_accessor and host_accessor for DPCPP because issue https://github.com/intel/llvm/issues/8332 was resolved.


kopylovanat:
DPCPP version needs a bump for CI to pass.

PR #602, opened 2023-04-18
Remove extra checks for common math functions

kopylovanat:
Remove checks for following functions:
- clamp(genfloath, sycl::half, sycl::half), 
- max(genfloath, sycl::half)
- min(genfloath, sycl::half)
- mix(genfloath, genfloath, sycl::half)
- step(sycl::half, genfloath)
- smoothstep(sycl::half, sycl::half, genfloath)

since there are no such functions in [4.17.7. Common functions](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_common_functions).

### DIFF
--- a/tests/math_builtin_api/modules/sycl_functions.py
+++ b/tests/math_builtin_api/modules/sycl_functions.py
@@ -131,9 +131,6 @@ def create_common_signatures():
     f_clamp_3 = funsig("sycl", "genfloatd", "clamp", ["genfloatd", "double", "double"])
     sig_list.append(f_clamp_3)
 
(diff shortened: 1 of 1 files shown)


gmlueck:
I think this is an oversight in the spec.  In fact, we have an open issue that these functions should be defined for the `sycl::half` data type (KhronosGroup/SYCL-Docs#321).

keryell:
Looks good but converting this to a draft while we solve the spec problem around it.

fraggamuffin:
Close

PR #605, opened 2023-04-18
Add waiting on queue for memcpy2d queue shortcut exceptions tests

kopylovanat:
To make sure that all submitted commands finished before releasing SECTION resources.


keryell (approved):
Good catch.

TApplencourt (approved):

PR #606, opened 2023-04-20
Remove sycl::context(platform) constructors

AmirIpma:
SYCL 2020 spec does not contain `context(platform) and context(platform, async_handler) ` constructors, so they were removed.


keryell (approved):
You are right.

TApplencourt (approved):

PR #608, opened 2023-04-21
Cover gaps for marray

vladimirkhashev:
Add checking of ==, !=, <, >, <=, >= operators and member types of marray


vladimirkhashev:
Disabled checking of `>=` and `<=` operators for computecpp because `<=` and `>=` are not being compiled in computecpp for type `double` (error: static_assert failed due to requirement `num_elements<double>() == 64UL` "Invalid number of elements provided in the template parameter")  and due to the operators are ambiguous for `const double` and `const sycl::marray<double, 1>`

TApplencourt (approved):

PR #610, opened 2023-04-21
sycl::vec operator[] const and operator vector_t() tests 

AmirIpma:
Added tests for const version of subscript operator type check and operator `vector_t()`

### DIFF
--- a/tests/vector_operators/generate_vector_operators.py
+++ b/tests/vector_operators/generate_vector_operators.py
@@ -1257,7 +1257,7 @@ subscript_operator_test_template = Template("""
       if (subscriptVec1[i] != data[i] || subscriptVec2[i] != data[i]
 // FIXME: re-enable when subscript operator for swizzle vec is implemented
 // link to issue: https://github.com/intel/llvm/issues/8880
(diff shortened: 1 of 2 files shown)


tests/vector_operators/generate_vector_operators.py
```diff
@@ -1243,6 +1242,17 @@
       subscriptVec2[i] = data[i];
     }
 
+    // operator[] type check
+    if(!std::is_same_v<decltype(subscriptVec2[0]), ${type}&>) {
+      resAcc[0] = false;
+    }
+
+    // const operator[] type check
+    if(!std::is_same_v<decltype(subscriptVec1[0]), const ${type}&>) {
```
keryell, line 1251:
```suggestion
    if constexpr (!std::is_same_v<decltype(subscriptVec2[0]), ${type}&>) {
      resAcc[0] = false;
    }

    // const operator[] type check
    if constexpr (!std::is_same_v<decltype(subscriptVec1[0]), const ${type}&>) {
```

keryell (changes requested):

keryell (approved):
Thanks!

PR #613, opened 2023-04-24
Fix mul_hi in math_reference

aleksmesh:
Add check if we need to carry on overflow with signed types.


util/math_reference.cpp
```diff
@@ -346,7 +346,16 @@ T mul_hi_signed(T a, T b) {
...
 
-  return (a >> msb) ^ (b >> msb) ? static_cast<T>(~hi) : hi;
+  T result = hi;
+  if ((a >> msb) ^ (b >> msb)) {
+    result = static_cast<T>(~hi);
+    // Check overflow to see if we need to carry
+    if (0 == static_cast<T>(a * b)) {
+      result += 1;
+    }
+  }
```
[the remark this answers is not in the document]
aleksmesh: Comments added.

And I was wrong - it's not overflow checking, just checking that `U(r*s + ((cross2+cross1)<<shft)) == 0` (i.e. low half equal to zero) and we need to carry.
And I replaced it by `static_cast<T>(a*b) == 0` just for code simplification.

util/math_reference.cpp
```diff
@@ -310,6 +310,15 @@ T mul_hi_unsigned(T x, T y) {
   return hi;
 }
 
+/**
+ * @brief Function to get high sizeof(T)*8 bits of the product of two signed T
```
keryell, line 314: I guess the fact that `T` is an integer is missing. Can `T` be signed or unsigned?

keryell: Add some `static_assert(std::is_integral_v...)`?

util/math_reference.cpp
```diff
@@ -346,7 +356,17 @@ T mul_hi_signed(T a, T b) {
   lo >>= shft;
   hi += lo + (cross1 >> shft) + (cross2 >> shft);
 
-  return (a >> msb) ^ (b >> msb) ? static_cast<T>(~hi) : hi;
+  T result = hi;
+  // if result is negative
+  if ((a >> msb) ^ (b >> msb)) {
```
keryell, line 361: All this obfuscated code does not make sense.
```suggestion
  if ((a < 0) ^ (b < 0)) {
```
Feel free to simplify all the code above too. :-)

util/math_reference.cpp
```diff
@@ -336,6 +345,7 @@ T mul_hi_signed(T a, T b) {
   r = static_cast<U>(a_pos) & mask;
   s = static_cast<U>(b_pos) & mask;
 
+  // calc half products
```
keryell, line 348: Comments are expected to be read by human beings
```suggestion
  // Compute half products
```

util/math_reference.cpp
```diff
@@ -336,6 +345,7 @@ T mul_hi_signed(T a, T b) {
...
   lo = r * s;
   hi = p * q;
   cross1 = (p * s);
```
keryell, line 351: You can remove the parenthesis while you are here. :-)

keryell: Why not using `std::abs` above, replacing `typedef` by `using`, make `a_pos` and `b_pos` just a `U` and remove the `static_cast` later...

util/math_reference.cpp
```diff
@@ -346,7 +356,17 @@ T mul_hi_signed(T a, T b) {
...
   hi += lo + (cross1 >> shft) + (cross2 >> shft);
 
-  return (a >> msb) ^ (b >> msb) ? static_cast<T>(~hi) : hi;
+  T result = hi;
+  // if result is negative
+  if ((a >> msb) ^ (b >> msb)) {
+    result = static_cast<T>(~hi);
+    // check that the low half is zero to see if we need to carry
+    if (0 == static_cast<T>(a * b)) {
+      result += 1;
```
keryell, line 365: I need more time to understand this...

util/math_reference.cpp
aleksmesh: `mul_hi_signed` is only available inside math_reference.cpp and is only called below in `mul_hi` with signed input parameters, so it seems to me that we can do without static type checking.

util/math_reference.cpp
aleksmesh: Fixed. Thanks.

util/math_reference.cpp
aleksmesh: Fixed it. It certainly looks better. Thanks!

util/math_reference.cpp
aleksmesh: Done.

util/math_reference.cpp
```diff
@@ -310,11 +310,20 @@ T mul_hi_unsigned(T x, T y) {
...
 }
 
+/**
+ * @brief Function to get high sizeof(T)*8 bits of the product of two signed T
+ *
+ * @tparam T signed type T
+ * @param a The first number of multiply
+ * @param b The second number of multiply
+ * @return T High sizeof(T)*8 bits of the sizeof(T*2)*8 bits result of the
+ * multiplication
```
keryell, line 320:
```suggestion

   @tparam T signed type of operand
   @param a The first operand of multiply
   @param b The second operand of multiply
   @return T with high sizeof(T)*8 bits of the sizeof(T)*2*8 bits result of the
   multiplication
```

util/math_reference.cpp
keryell: OK, if the result of the multiplication of 2 non null operands is 0 there is a carry to propagate. If only one operand is 0, result is -1 and then adding 1 set it back to 0.
This looks like it works. :-)

keryell (approved):
Thanks. Just the comments to clarify.

util/math_reference.cpp
aleksmesh: Done. Thanks!

PR #614, opened 2023-04-24
Add missed check for get_kernel_bundle

kopylovanat:
Add checks for:
`get_kernel_bundle<KernelName, BundleState>(const context& ctxt) `
`get_kernel_bundle<KernelName, BundleState>(const context& ctxt, const std::vector<device>& devs)`
with incompatible kernels.

Fix checks for get_kernel_bundle with builtin kernels.


keryell (approved):
Thanks.

PR #615, opened 2023-04-25
Fix info::kernel::num_args usage in kernel tests

kopylovanat:
Moved check_get_info_param top section that uses built_in kernel because otherwise exception is thrown as expected.
Changed semantics tests to use get_context() instead of info::kernel::num_args because there are not built_in kernels.


keryell (approved):
Thanks.

PR #625, opened 2023-05-03
math_builtin host checks

AmirIpma:
Added checks for math_builtin functions on host in `check_function()` for functions without pointers and `check_function_multi_ptr_private()` for functions with pointers because there is no need to pass accessor into lambda function.

### DIFF
--- a/tests/math_builtin_api/math_builtin.h
+++ b/tests/math_builtin_api/math_builtin.h
@@ -173,6 +173,11 @@ void check_function(sycl_cts::util::logger &log, funT fun,
   if (!verify(log, kernelResult, ref, accuracy, comment))
     FAIL(log,
          "tests case: " + std::to_string(N) + ". Correctness check failed.");
+
+  // host check
+  auto hostRes = fun();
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks.

keryell:
I have created https://github.com/KhronosGroup/SYCL-CTS/pull/635 to address the formatting issue.

bader:
> I have created #635 to address the formatting issue.

@keryell, you mean, we should merge #635 first, then fix the formatting issue and after that we can merge this change. Am I right?

keryell:
> > I have created #635 to address the formatting issue.
> 
> @keryell, you mean, we should merge #635 first, then fix the formatting issue and after that we can merge this change. Am I right?

It would be the best but I did not even expect that. Just an investment for the next PR is already good.

bader:
@AmirIpma, #635 is merged. Could you re-apply clang-format for your patch with updated config file, please?

AmirIpma:
> @AmirIpma, #635 is merged. Could you re-apply clang-format for your patch with updated config file, please?

Done.

PR #626, opened 2023-05-03
ext_oneapi_bfloat16 test plan

AmirIpma:
Test plan for oneAPI [bfloat16](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_bfloat16.asciidoc) extension


test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+* `std::is_default_constructible_v<bfloat16> == true`.
+* `std::is_copy_constructible_v<bfloat16> == true`.
+
+`bfloat16(const float &a)` +
+`bfloat16(const sycl::half &a)`
+
+Since `bfloat16` has 7 bits significand, epsilon value for this type is 2^-7, or 0.0078125. Therefore, objects created using these constructors must be in the range `(a - 0.0078125, a + 0.0078125)`. Create `bfloat16` object named `bf` and verify:
+
+* `sizeof(bf) == 2`
+* `(bf > a - epsilon) && (bf < a + epsilon)`
```
keryell, line 56: All this seems over-simplistic.
To give you an idea, if `a = 1f30`, `a - epsilon` or `a + epsilon` will be actually `a`, so the interval comparison will not hold.
You need to define differently your interval to take into account the exponent too and also define the type to be used to compute the interval checking (probably `float`).

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+
+* `neg_bf == -bf`
+* `bf == -neg_bf`
+
+(Prefix) +
+`bfloat16 &operator++(bfloat16 &bf)` +
+`bfloat16 &operator--(bfloat16 &bf)`
+
+* Check if it adds/substracts 1 to the value of the object referenced by this `bf`.
+* Check that new value of the referenced object is equal to `(previous value +/- 1)`.
```
keryell, line 83: Should we check this in both intervals where `++` does something and where `++` does nothing (because the precision in that interval is not enough tore represent the ±1)?

test_plans/bfloat16.asciidoc
AmirIpma: I will rewrite this part, but the spec says that dynamic range of `float` and `bfloat` is almost the same. I calculated a static error cause by fewer bits. So, do I have to describe it in details and do we have to test dynamic range?

keryell: You might test with epsilon like that but that means that you need to test with values around 1, which was not said anywhere.

test_plans/bfloat16.asciidoc
keryell: If there is nobody chiming on this, you can leave it as is.

test_plans/bfloat16.asciidoc
gmlueck, line 56:
```suggestion
* `(bf > bf - epsilon) && (bf < bf + epsilon)`
```

I assume this is what you meant?  Otherwise, I don't understand what `a` is.

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+`bfloat16()` +
+`bfloat16(const bfloat16 &)`
+
+Check that:
+
+* `std::is_default_constructible_v<bfloat16> == true`.
+* `std::is_copy_constructible_v<bfloat16> == true`.
+
+`bfloat16(const float &a)` +
+`bfloat16(const sycl::half &a)`
```
gmlueck, line 51: How are you testing that it is possible to construct a `bfloat16` from `float` or `sycl::half`?  The description below doesn't explain this.  There should also be tests to verify the assignment operators taking `float` and `sycl::half`.

test_plans/bfloat16.asciidoc
keryell: > I assume this is what you meant? Otherwise, I don't understand what `a` is.

I am unsure.

AmirIpma: `a` is the argument of `bfloat16` constructors listed above (`bfloat16(const float &a)`and 
`bfloat16(const sycl::half &a)`)

gmlueck: Oh, I see.  Can you clarify this by saying:

```suggestion
* `(bf > a - epsilon) && (bf < a + epsilon)`, where `a` is the value passed to the constructor.
```

bader:
@AmirIpma, please, address the comments from @keryell and @gmlueck.

test_plans/bfloat16.asciidoc
AmirIpma: Applied.

test_plans/bfloat16.asciidoc
AmirIpma: I will clarify this section

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+* Check results of arithmetic operations returned to initial `bfloat16` object.
+* Check returned value type.
+
+OP is `+`, `-`, `*`, `/` +
+`bfloat16 operatorOP(const bfloat16 &lhs, const bfloat16 &rhs)`
+
+* Check results of arithmetic operations.
+* Check returned value type.
+
+OP is `==`, `!=`, `<`, `>`, `<=`, `>=` +
```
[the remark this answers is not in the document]
AmirIpma: Applied.

test_plans/bfloat16.asciidoc
AmirIpma: I am going to remove that precision measurement part, because it is too complicated and was not mentioned in spec and test dynamic range comparison with some error calculating between bfloat16 and float instead.

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,149 @@
...
+    ** Convert it to `bfloat16` variable `BFLT16_MAX`
+
+    ** Create `std::bitset` with 16 bits to store minimum value;
+    ** Set all bits to 0;
+    ** Set bit 8 to 0 to get minimum normal value;
+    ** Convert it to `bfloat16` variable `BFLT16_MIN`
+
+* Calculate `float` dynamic range equal to `log10(FLT_MAX / FLT_MIN)` or `log10(FLT_MAX) - log10(FLT_MIN)`;
+* Calculate `bfloat16` dynamic range equal to `log10(BFLT16_MIN) - log10(BFLT16_MIN)`;
+* Compare integer parts of computed values.
```
gmlueck, line 87: I don't understand the point of this test.  It seems like it is just testing mathematical properties of the bfloat16 floating point representation, which doesn't really test the implementation.

The only part of the implementation that you are testing is that you can convert the special values "BFLT16_MIN" and "BFLT16_MAX" to `float`.  This does seem valuable, but it is not testing the "dynamic range".

I'd suggest changing this test and calling it "Test special values".  For example, the test for BFLT16_MIN could look like:

```
uint16_t min_bits = 0b0000000010000000;
bfloat16 bfmin = sycl::bit_cast<bfloat16>(min_bits);
uint32_t float_bits = 0b00000000100000000000000000000000;
float expect = sycl::bit_cast<float>(float_bits);
if (bfmin != expect)
  /* test fails */
```

You could expand this test to other special values like positive / negative infinity and NaN.

test_plans/bfloat16.asciidoc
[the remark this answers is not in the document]
gmlueck: You missed the other occurrence of `<=`.

test_plans/bfloat16.asciidoc
AmirIpma: Do I understand correctly that dynamic range is ratio of maximum to minimum value? I took a logarithm to reduce very large values of this ratio, as audio engineers do. Spec says that dynamic range of `bfloat16` is the same as `float`, so I wanted to test it this way.

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,149 @@
...
+* Calculate the maximum and minimum positive values of `bfloat16`:
+    ** Create `std::bitset` with 16 bits to store maximum value;
+    ** Set all bits to 1;
+    ** Set bit 15 (sign bit) and bit 7 (special values) to 0;
+    ** Convert it to `bfloat16` variable `BFLT16_MAX`
+
+    ** Create `std::bitset` with 16 bits to store minimum value;
+    ** Set all bits to 0;
+    ** Set bit 8 to 0 to get minimum normal value;
+    ** Convert it to `bfloat16` variable `BFLT16_MIN`
```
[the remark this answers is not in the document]
AmirIpma: Applied.

test_plans/bfloat16.asciidoc
gmlueck: I think this is the relevant part of the spec:

> However, bfloat16 allots 8 bits to the exponent instead of the 5 bits used by sycl::half and 7 bits to the significand versus 10 bits used by sycl::half. Thus, bfloat16 has the same dynamic range as a 32-bit float but with reduced precision.

The statement about the dynamic range is simply a result of the format, which has an 8 bit exponent and a 7 bit significand.  If you want to test this sentence in the spec, you should somehow test that the format has an 8 bit exponent and a 7 bit significand.

The test I propose above would do this because it constructs the value of a `bfloat16` number by setting specific bits in the exponent and significand and then comparing the `bfloat16` number with the expected value of a `float`.  This comparison will fail if the implementation does not use an 8 bit exponent and 7 bit significand.

AmirIpma: Done.

test_plans/bfloat16.asciidoc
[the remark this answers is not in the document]
AmirIpma: Done.

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+    `0b1111111111000001,`         +
+    `0b0111111110000001, // sNan` +
+    `0b1111111110000001`
+* `float`:                                        +
+    `0b01111111110000000000000000000001, // qNan` +
+    `0b11111111110000000000000000000001,`         +
+    `0b01111111100000000000000000000001, // sNan` +
+    `0b11111111100000000000000000000001`
+
+* Verify that `std::isnan()` is `true`` for these values.
```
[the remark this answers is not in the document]
AmirIpma: Just wanted to show where these values for `bfloat16` came from using `float` special values from IEEE 754 spec

test_plans/bfloat16.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+[source,c++]
+----
+uint16_t bfloat16_bits = 0b0000000010000000;
+uint32_t float_bits    = 0b00000000100000000000000000000000;
+
+bfloat16 bf_min = sycl::bit_cast<bfloat16>(bfloat16_bits);
+float float_min = sycl::bit_cast<float>(float_bits);
+----
+
+Verify that the minimum values of `bfloat16` type is equal to the minimum value of `float`. This comparison will fail if the implementation does not use an 8 bit exponent and 7 bit significand for `bfloat16`.
```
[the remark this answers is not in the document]
AmirIpma: Applied.

bader:
@keryell, could you approve, please?

keryell (approved):
Thanks.
@bader I am at a conference this week, so lower review bandwidth unfortunately...

PR #628, opened 2023-05-05
Update local accessor usage in legacy tests

kopylovanat:
In SYCL 2020 local accessor must not be used in a SYCL kernel function that is invoked via single_task or via the simple form of parallel_for that takes a range parameter.


keryell (approved):
Thanks.

PR #632, opened 2023-05-09
Add test plan for execution and memory model

kopylovanat:
Test plan for checks for [3.7.1.2. SYCL command groups and execution order](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:command-groups-exec-order) and [3.8.1. SYCL application memory model](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sub.section.memmodel.app).


test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+buffer<int, 1> sub_buf1(buffer, 0, 5);
+buffer<int, 1> sub_buf2(buffer, 10, 5);
+int *pflag = sycl::malloc_shared<int>(1, queue);
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc1(sub_buf1, cgh);
+  cgh.single_task([=] {
+    // long loop
+    for (int i = SIZE; i >= 0; i--)
+      acc1[i] = i;
+    *pflag = 1;
```
keryell, line 80: I guess you need an `atomic_ref` here?

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+  });
+});
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc2(sub_buf2, cgh);
+  cgh.single_task([=] {
+    for (int i = SIZE; i >= 0; i--)
+      acc2[i] = i;
+    *pflag = 2;
```
keryell, line 88: Here too?

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+  });
+});
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc(buffer, cgh);
+  sycl::accessor<int, 1> res_acc(res_buffer, cgh);
+  copy(acc1, res_acc);
+});
+----
+
+Check that *pflag is equal to 1, that means first two commands executes independently, and check that all elements in result buffer are equal to expected values which means that requisites for third command were satisfied before it started to execute.
```
keryell, line 98: For me we can have either `*pflag` equals to 1 or 2 when the copy happens, but how to test it with this `cgh.copy()`. I think you need a normal kernel instead to read it and write it for example in another buffer.

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+    sycl::atomic_ref<int> rflag(*pflag);
+    for (int i = SIZE; i >= 0; i--) acc2[i] = i;
+    rflag = 2;
+  });
+});
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc(buffer, cgh);
+  sycl::accessor<int, 1> res_acc(res_buffer, cgh);
+  sycl::accessor<int, 1> flag_res_acc(flag_res_buffer, cgh);
+  copy(acc1, res_acc);
```
gmlueck, line 92: Did you intend to remove this `copy`?  It seems like the kernel below is also copying.

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+  cgh.parallel_for(SIZE, [=](auto item) {
+    res_acc[item.get_id()] = acc[item.get_id()];
+    flag_res_acc[item.get_id()] = *pflag;
+  });
+});
+----
+
+Check that *pflag was equal to 1 for all work items, that means first two commands executes independently, and check that all elements in result buffer are equal to expected values which means that requisites for third command were satisfied before it started to execute.
```
gmlueck, line 100: I don't understand this test.  It seems like `*pflag` could be either `1` or `2` because the order of the first two kernels is indeterminate.  (In fact, they could even run in parallel.)

What is this test trying to validate?

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc(buffer, cgh);
+  cgh.single_task([=] {
+    // long loop
+    for (int i = SIZE; i >= 0; i--) acc[i] = i;
+    *pflag = 42;
+  });
+});
+sycl::host_accessor<int, 1> host_acc(buffer);
+-- --
```
gmlueck, line 118:
```suggestion
----
```

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+
+   Check that immediately after host_acc creation *pflag is equal to expected
+       value,
+    that means that host_accessor acts as barrier.
```
gmlueck, line 122: Why the weird formatting here?

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+  sycl::accessor<int, 1> res_acc1(res_buffer1, cgh);
+  sycl::accessor<int, 1> res_acc2(res_buffer2, cgh);
+  cgh.parallel_for(sycl::range<1>(SIZE), [=](id<1> index) {
+    res_acc1[index] = acc1[index];
+    res_acc2[index] = acc2[index];
+  });
+});
+----
+
+Check that all elements in result buffer are equal to expected values which means that requisites for third command were satisfied before it started to execute.
```
gmlueck, line 54: I think this test looks good, but I'm surprised we don't already have a test like this from the SYCL 1.2.1 test suite.

test_plans/execution_model.asciidoc
kopylovanat: Yes, removed

test_plans/execution_model.asciidoc
kopylovanat: I couldn't find such checks in CTS unfortunately.

test_plans/execution_model.asciidoc
kopylovanat: The idea was that since they executes independently the one without long loop is expected to finish first, but now I think we still can't guarantee that. Do you think there is other way to check that kernels with assess to different parts of buffer execute independently or we should discard this check completely?

test_plans/execution_model.asciidoc
kopylovanat: Fixed.

test_plans/execution_model.asciidoc
kopylovanat: Fixed.

test_plans/execution_model.asciidoc
gmlueck: The spec does not require the first two kernels to run concurrently, it only says that it is legal to run them concurrently if the implementation chooses.  Therefore, I think it's hard to write a test for that.

However, we could write a related test.  The spec says that kernels with overlapping sub-buffers can **not** be run concurrently.  We could create a test that has overlapping sub-buffers, and then verify that the kernels run in a specific order.

kopylovanat: Changed test case to check order of kernels with overlapping sub-buffers in https://github.com/KhronosGroup/SYCL-CTS/pull/632/commits/04f949f0231db3d479e7fe828581afcb33ca08b3

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,112 @@
...
+    // long loop
+    for (int i = acc1.size(); i >= 0; i--) acc1[i] = i;
+    rflag = 1;
+  });
+});
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc2(sub_buf2, cgh);
+  sycl::accessor<bool, 1> res_acc(res_buffer, cgh);
+  cgh.single_task([=] {
+    res_acc[0] = *pflag == 1;
```
gmlueck, line 85: I think the point of this test is to validate that the runtime waits for the first kernel to complete before executing the second kernel.  The logic for checking that seems broken.  It seems like the second kernel should use `atomic_ref` to set `*pflag` to `2`.  Then, the validation logic should verify that `*pflag == 2` after both kernels complete.

The current validation logic checks if `*pflag == 1`, but that will of course be true because it is only the first kernel that writes to `*pflag` and it sets the value to `1`.

gmlueck: I see.  I think this description needs to be updated then:

> Check that *pflag was equal to 1, that means that requisites for second command with overlapping sub-buffer were satisfied before it started to execute.

To say something like:

> Check that `res_buffer` is equal to `true`, which means that requisites for second command with overlapping sub-buffer were satisfied before it started to execute.

gmlueck: You should also use `atomic_ref` consistently.  If you want to use `atomic_ref` to set the value of `*pflag` in the first kernel, you should also use `atomic_ref` to read the value of `*pflag` in the second kernel.

kopylovanat: Thank you! Fixed.

gmlueck (approved):

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+    // <another long loop>
+    rflag = 1;
+  });
+});
+queue.submit([&](handler& cgh) {
+  sycl::accessor<int, 1> acc2(sub_buf2, cgh);
+  sycl::accessor<bool, 1> res_acc(res_buffer, cgh);
+  cgh.single_task([=] {
+    sycl::atomic_ref<int> rflag(*pflag);
+    res_acc[0] = (rflag.load() == 1);
```
keryell, line 91: Not really useful
```suggestion
    res_acc[0] = (rflag == 1);
```

test_plans/execution_model.asciidoc
```diff
@@ -0,0 +1,118 @@
...
+
+Check that all elements in result buffer are equal to expected values which means that requisites for third command were satisfied before it started to execute.
+
+=== Execution order of three command groups submitted to the different queues
+
+Same as previous tests but each command is run on different queue with diffrent context.
+
+=== Requirements on overlapping sub-buffers
+
+This check is skipped if device doesn't support `aspect::usm_device_allocations`.
```
gmlueck, line 62: You changed this from `aspect::usm_atomic_device_allocations` to `aspect::usm_device_allocations` in your last commit.  However, the code below still uses `sycl::atomic_ref` on the `*pflag`.  If you want to use `atomic_ref`, then you need to check for `aspect::usm_atomic_device_allocations`.

kopylovanat: I realized that there are no mentions of `aspect::usm_atomic_device_allocations` in current version of SYCL 2020, and after re-reading [4.8.2. Kinds of unified shared memory](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_kinds_of_unified_shared_memory) I got the impression that such atomic operations is expected to be supported if `aspect::usm_device_allocations` is supported.

gmlueck: Oh, you are right!  I was thinking of `usm_atomic_shared_allocations`.  Yes, `usm_device_allocations` is sufficient to ensure that atomic operations are supported on USM "device" allocations".

keryell (approved):
Thanks.

PR #633, opened 2023-05-09
Add test plan for SYCL_EXTERNAL

kopylovanat:
This is a test plan for SYCL_EXTERNAL macro as described in [5.10.1. SYCL functions and member functions linkage](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:syclexternal).

### DIFF
--- a/test_plans/sycl_external.asciidoc
+++ b/test_plans/sycl_external.asciidoc
@@ -1,0 +1,64 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for device image
+
+This is a test plan for SYCL_EXTERNAL macro as described in Section 5.10. of the SYCL 2020 specification.
+
+== Testing scope
+
+=== Backend coverage
+
+All the tests described below are not backend-specific and are performed for any SYCL backend.
+
+=== Device coverage
+
+All tests construct a test device for which conformance is assessed. All the tests described below are performed once for that test device.
+
+== Tests
+
+All test cases a function is declared with SYCL_EXTERNAL in translation unit with test case and defined in some translation unit, where the function is also declared with SYCL_EXTERNAL.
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks.

test_plans/sycl_external.asciidoc
```diff
@@ -0,0 +1,64 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for device image
```
gmlueck, line 4: Update title

test_plans/sycl_external.asciidoc
```diff
@@ -0,0 +1,64 @@
...
+
+=== Device coverage
+
+All tests construct a test device for which conformance is assessed. All the tests described below are performed once for that test device.
+
+== Tests
+
+All test cases a function is declared with SYCL_EXTERNAL in translation unit with test case and defined in some translation unit, where the function is also declared with SYCL_EXTERNAL.
+This function is called in test case from both host and device code. From device code it's called from both single_task and parallel_for (for 1 item).
+Function has accessor as a parameter and is used to assign value to it. For call from host host_accessor is used.
```
gmlueck, line 22: Note that `SYCL_EXTERNAL` is an optional feature, so implementations are not required to support it.  I think this test needs to use `#ifdef SYCL_EXTERNAL`, and the test should be skipped if the implementation doesn't define that macro.

gmlueck (approved):

test_plans/sycl_external.asciidoc
kopylovanat: Fixed.

test_plans/sycl_external.asciidoc
kopylovanat: Added condition for `#ifdef SYCL_EXTERNAL`.

gmlueck (approved):

PR #634, opened 2023-05-09
Prevent re-creating context for every testing type

vladimirkhashev:
Changed queue creation (`get_cts_object::queue()` -> `once_per_unit::get_queue()`) in several tests to use one individual queue per translation unit. Changes are done for the tests where combinations for different types are performed. In these tests queue object was created many times. In case when a compiler implementation isn't using a default context for queue creation such behavior leads to a new context creation for every kernel, then it may require re-creation of all SYCL entities which are context-dependent, i.e. re-compilation of all kernel bundles involved. Eventually it can lead to long execution time for tests that re-create queue object many times because SYCL 2020 spec does not require any implicit optimizations to be present for this case. Such problem was happened for DPCPP compiler implementation on windows.


ProGTX (approved):

tests/buffer/buffer_constructors_common.h
```diff
@@ -149,7 +150,7 @@ class buffer_ctors {
         sycl::buffer<T, dims> buf1(cont1, propList);
         sycl::buffer<T, dims> buf2(cont2);
         {
-          util::get_cts_object::queue()
+          once_per_unit::get_queue()
               .submit([&](sycl::handler &cgh) {
```
bader, line 154:
```suggestion
              .submit([&](sycl::handler& cgh) {
```

tests/buffer/buffer_constructors_common.h
```diff
@@ -179,7 +180,7 @@ class buffer_ctors {
         sycl::buffer<T, dims, decltype(logging_alloc)> buf2(cont2,
                                                             logging_alloc);
         {
-          util::get_cts_object::queue()
+          once_per_unit::get_queue()
               .submit([&](sycl::handler &cgh) {
```
bader, line 184:
```suggestion
              .submit([&](sycl::handler& cgh) {
```

tests/math_builtin_api/math_builtin.h
```diff
@@ -37,8 +37,7 @@ inline sycl::queue makeQueueOnce() {
   return q;
 }
 
-template <typename returnT, typename ArgT>
-struct privatePtrCheck {
+template <typename returnT, typename ArgT> struct privatePtrCheck {
```
bader, line 40:
```suggestion
template <typename returnT, typename ArgT>
struct privatePtrCheck {
```

bader (approved):

PR #639, opened 2023-05-10
sycl::kernel_bundle begin() end() tests 

AmirIpma:
Implemented new tests for `sycl::kernel_bundle` `begin()` and `end()` member functions return type and LegacyForwardIterator requirement.


AmirIpma:
[check-clang-format](https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4937167520/jobs/8825591118)
GitHub Actions has encountered an internal error when running your job.

PR #642, opened 2023-05-11
[CI] Uplift DPC++ version to d1f67ca.

bader:


bader:
@psalz, `check-clang-format` complains about someone's else code.
See this message: https://github.com/KhronosGroup/SYCL-CTS/actions/runs/4983629657/attempts/1#summary-13494404248. This PR doesn't change the file `check-clang-format` complains about. Any ideas what might be causing this? What command to build the diff?

psalz:
Sorry for the late response, was on vacation. I assume this is due to the back-merge of the SYCL-2020 branch which included a change to the file in question, however it is strange that this formatting issue was not reported in the original PR #631 🤔 .

The diff is computed [here](https://github.com/KhronosGroup/SYCL-CTS/blob/0479c2edadb7491c8782298f12e2429eccefd4b0/.github/workflows/cts_ci.yml#LL99C75-L99C75), where we use `event.pull_request.base.sha`, which afaik refers to the current HEAD of the base branch (in this case `SYCL-2020`) at the time the workflow is triggered (not when it was opened). So this should already have included the changes you back-merged - I'm really not sure what happened here.

PR #644, opened 2023-05-11
Fix placeholder accessor tests

kopylovanat:
To make sure that placeholder accessor usage in kernel is not optimized out.


keryell (approved):
Thanks.

PR #645, opened 2023-05-12
Improve group reduce function object coverage and add result verification

jiezzhang:
1. Test group reduce functions with all function objects in FULL CONFORMANCE
2. Verify results with std::accumulate
3. Update incorrect usage of `SYCL_CTS_ENABLE_FULL_CONFORMANCE`


keryell:
Can you resolve the recent conflicts?

keryell (approved):

PR #646, opened 2023-05-15
sycl::stream enable semantics tests

AmirIpma:
Enabled `bool check(const sycl::stream& stream)` in  `stream_semantics.cpp` due to solved [issue](https://github.com/intel/llvm/issues/8326)


keryell (approved):
Nice to see progress.

PR #648, opened 2023-05-16
Fix more image-related tests

steffenlarsen:
This commit fixes the following in the image-related tests:
 - Adds the data type template argument to `get_host_access` calls.
 - Fixes storage classes for sampled and unsampled image checks.
 - Replace calls to non-existant `range` memeber function to existing `get_range`.
 - Replaces subscripting on host image accessors to approriate `read` calls.
 - Removes a test case trying to call non-const `get_host_access` on a constant copy of an image.
 - Adds optional requirements to `common_reference_semantics::check_kernel` to enforce skips for images when unsupported on the selected device.


keryell (approved):
Thanks.

PR #650, opened 2023-05-17
sycl::usm_allocator tests for constructors with property_list

AmirIpma:
Implemented tests for `usm_allocator` constructors with empty property list.


keryell (approved):
Thanks.

PR #652, opened 2023-05-17
Test for fixed width types sizes on host and device

AmirIpma:
Due to [spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:scalartypes) built-in scalar data types with fixed width must have same size on host and device. Added new tests case to check it.


keryell (approved):
Thanks.

PR #659, opened 2023-05-23
sycl::is_device_copyable test plan

AmirIpma:
Created test plan for `sycl::is_device_copyable` type trait


test_plans/is_device_copyable.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+    `std::is_base_of_v<std::integral_constant<bool, true>, is_device_copyable<T>> == true` or
+    `std::is_base_of_v<std::integral_constant<bool, false>, is_device_copyable<T>> == true`.
+
+    * The member names of the base characteristic (`std::integral_constant`) are not hidden and are unambiguously available. Verify that:
+
+        ** `std::integral_constant<bool, true>::value_type` and `is_device_copyable<T>::value_type` are the same type
+
+        ** `std::integral_constant<bool, true>::type` and `is_device_copyable<T>::type` are the same type
+        ** Operators `std::integral_constant<bool, true>::value_type() and `is_device_copyable<T>::value_type()` have the same return type
+        ** `std::integral_constant<bool, true>::operator()` returns true
```
gmlueck, line 69: Isn't this last bullet item just testing something from the C++ headers?  It doesn't seem like it's testing anything from SYCL.

test_plans/is_device_copyable.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+
+=== Specialization for user defined class
+
+Add a specialization for `T`:
+
+`template<> +
+struct is_device_copyable<T> : std::true_type {}` +
+
+* Verify that `is_device_copyable<T>::value == true`
+* Create an instance of class `T` on the host and pass an accessor to it in device code;
```
gmlueck, line 79: Should we add a test here:

* Verify that the object of type `T` in the kernel has the same value that was initialized on the host.

test_plans/is_device_copyable.asciidoc
AmirIpma: It tests UnaryTrait requirement for `is_device_copyable`,  "The member names of the base characteristic are not hidden and are unambiguously available", therefore `is_device_copyable` must has access to `std::integral_contant` members. This is what this part is about, but I don't know how to test an access to members in other way. If it's not needed I can remove this part.

test_plans/is_device_copyable.asciidoc
AmirIpma: I planned to initialize new object on device with another data and pass this data to the host object, so main verification would be the last point of this section.

test_plans/is_device_copyable.asciidoc
gmlueck: If you want to test that `std::integral_constant::operator()` is unambiguously available to `is_device_copyable`, then it seems like you should test this instead:

* `is_device_copyable<T>::operator()` return true.

test_plans/is_device_copyable.asciidoc
gmlueck: Yes, I see that.  The last bullet point verifies that the object of type `T` can be copied from device-to-host.  The bullet I proposed checks the opposite case: that the object can be copied from host-to-device.

test_plans/is_device_copyable.asciidoc
AmirIpma: Instead of all members of `std::integral_constant` or `operator()` only?

gmlueck: It seems like your last bullet point:

> ** `std::integral_constant<bool, true>::operator()` returns true

is trying to validate whether `std::integral_constant::operator()` is unambiguously available to `is_device_copyable`.  If this is your intent, it seems like that bullet should be this instead:

> ** `is_device_copyable<T>::operator()` return true.

AmirIpma: Updated, but with comparison between them because it haven't always return true

test_plans/is_device_copyable.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+
+    /* -- common interface members -- */
+
+    int a;
+    float b;
+    char c;
+};
+```
+
+Given that `is_device_copyable<T>::value` can be either `true` or `false`, `integral_constant_t` will be used instead of `std::integral_constant<bool, true>` and `std::integral_constant<bool, false>`.
```
gmlueck, line 41: I don't understand this statement.  Doesn't your test program set `is_device_copyable<t>` to true?  Therefore, the value can only be `true`.

test_plans/is_device_copyable.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+
+* Verify that `is_device_copyable<T>::value == true`;
+
+* Check data transmission:
+
+    ** Initialize a constant instance of class `T` as a reference;
+    ** Create an instance of class `T` on the host by assigning values from the reference object;
+    ** Create an instance of class `T` on the device by assigning values from the reference object;
+    ** Get access to the device object using `accessor` to check data-to-host copying;
+    ** Verify that the object of type `T` in the kernel has the same value that was initialized on the host.
```
gmlueck, line 81: This new description seems a lot less clear to me.  It seems like you are only testing that `T` is correctly copied from host-to-device, but you are not testing the other way around (device-to-host).  It might be clearer if you wrote some small code snippet showing the basic structure of the test.

AmirIpma: "Get access to the device object using `accessor` to check data-to-host copying;" isn't this point testing device to host copying? By the way, there is one more typo, device-to-host and instead of data-...

AmirIpma: I have already written the sample code that is described in this section
```  
  T host_object(reference_instance);
  T device_object_access;
  {
    sycl::buffer<user_def_type_derived, 1> buf(&device_object_access, {1});
    queue.submit([&](sycl::handler& cgh) {
      sycl::accessor<user_def_type_derived, 0, sycl::access_mode::read_write> acc(buf, cgh);
      cgh.single_task<kernel_name>([=] {
        T device_object(reference_instance);
        acc = device_object;
      });
    });
    queue.wait_and_throw();
  }
```

test_plans/is_device_copyable.asciidoc
AmirIpma: I tried to add specialization after requirements tests using inheritance, but You are right, I will rewrite it.

test_plans/is_device_copyable.asciidoc
```diff
@@ -0,0 +1,81 @@
...
+* The destructor has no effect when executed on the device.
+
+Define a new class of type `T` with public default constructor, destructor, copy constructor, copy assignment operator and several public members:
+```
+class T {
+  public:
+    T() = default;
+    ~T() = default;
+
+    /* -- common interface members -- */
```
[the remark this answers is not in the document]
AmirIpma: Returned and extended.

test_plans/is_device_copyable.asciidoc
AmirIpma: Applied.

gmlueck (approved):

PR #660, opened 2023-05-23
Fix atomic_ref stress tests

kopylovanat:
Add checking with possible error range for floating point type as per test plan.
Limit size for range to prevent possible overflow in calculations.


bader (approved):
Do we run atomic_ref stress tests only in the full mode or always?

keryell (approved):
Thanks.

kopylovanat:
> Do we run atomic_ref stress tests only in the full mode or always?

They are always run.

PR #662, opened 2023-05-24
Cover gap for handler::require exception

aleksmesh:
Add check for exception with handler::require and empty accessor.

### DIFF
--- a/tests/handler/handler_exceptions.cpp
+++ b/tests/handler/handler_exceptions.cpp
@@ -1,0 +1,42 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2017-2022 Codeplay Software LTD. All Rights Reserved.
+//  Copyright (c) 2022-2023 The Khronos Group Inc.
+//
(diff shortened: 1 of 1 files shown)


tests/handler/handler_exceptions.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace handler_exceptions {
+
+using AccT =
+    sycl::accessor<int, 1, sycl::access_mode::write, sycl::target::device>;
+
+// FIXME: re-enable when sycl::errc is implemented in computecpp
+DISABLED_FOR_TEST_CASE(ComputeCpp)
+("handler require() exception", "[handler]")({
+  auto queue = sycl_cts::util::get_cts_object::queue();
+  auto action = [&]() {
```
keryell, line 38: Useless
```suggestion
  auto action = [&] {
```

keryell (approved):
Thanks.

tests/handler/handler_exceptions.cpp
aleksmesh: Fixed. Thanks!

PR #666, opened 2023-05-26
Fix type pack creation for atomic_ref tests

vladimirkhashev:
Because of type long can have size greater than 32 bits added checking of its size. Now this type is included in `atomic64` type pack if its size greater than 4 bytes.

### DIFF
--- a/tests/atomic_ref/atomic_ref_common.h
+++ b/tests/atomic_ref/atomic_ref_common.h
@@ -97,25 +97,41 @@ inline std::string get_section_name(const std::string& type_name,
  * @brief Factory function for getting type_pack with fp64 type
  */
 inline auto get_atomic64_types() {
-  static const auto types =
 #if SYCL_CTS_ENABLE_FULL_CONFORMANCE
-      named_type_pack<long long, unsigned long long>::generate(
-          "long long", "unsigned long long");
+  if constexpr (sizeof(long) == 32) {
+    static const auto types =
+        named_type_pack<long long, unsigned long long>::generate(
+            "long long", "unsigned long long");
+    return types;
+  } else {
+    static const auto types =
+        named_type_pack<long, unsigned long, long long,
+                        unsigned long long>::generate("long", "unsigned long",
+                                                      "long long",
+                                                      "unsigned long long");
+    return types;
+  }
 #else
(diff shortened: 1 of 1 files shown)


tests/atomic_ref/atomic_ref_common.h
```diff
@@ -97,25 +97,41 @@ inline std::string get_section_name(const std::string& type_name,
...
 /**
  * @brief Factory function for getting type_pack with all generic types
  */
 inline auto get_full_conformance_type_pack() {
-  static const auto types =
-      named_type_pack<int, unsigned int, long int, unsigned long int,
-                      float>::generate("int", "unsigned int", "long int",
-                                       "unsigned long int", "float");
-  return types;
+  if constexpr (sizeof(long) > 32) {
```
keryell, line 123: It might be possible on an FPGA... ;-)
Comments.

keryell (changes requested):

tests/atomic_ref/atomic_ref_common.h
keryell: Typically the information from the PR descriptions should appear somehow as comment in the code to be clearer to read.

tests/atomic_ref/atomic_ref_common.h
```diff
@@ -97,25 +97,41 @@ inline std::string get_section_name(const std::string& type_name,
  * @brief Factory function for getting type_pack with fp64 type
  */
 inline auto get_atomic64_types() {
-  static const auto types =
 #if SYCL_CTS_ENABLE_FULL_CONFORMANCE
-      named_type_pack<long long, unsigned long long>::generate(
-          "long long", "unsigned long long");
+  if constexpr (sizeof(long) == 32) {
```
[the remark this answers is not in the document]
vladimirkhashev: Fixed. I'm talking about `long` here, according to c++ standard `long` might to be at least 32 bits or wider, so if this type is grater than 4 bytes we should perform `atomic_ref` tests with this type only if device supports `atomic64` aspect. Added comments.

bader (approved):

ProGTX (approved):

keryell (approved):
Great clarification!

PR #668, opened 2023-05-29
Update tests usage of accessor providing multi_ptr

kopylovanat:
Updated tests that are use accessor and multi_ptr - get_pointer returns T* for target::device and get_multi_ptr should be used instead.
Changed accessors that are used in multi_ptr to read_write instead of read to accommodate this Spec change https://github.com/KhronosGroup/SYCL-Docs/pull/405  
Changed async_work_group_copy to take decorated ptr as per Spec.

### DIFF
--- a/tests/multi_ptr/multi_ptr_subscript_op_common.h
+++ b/tests/multi_ptr/multi_ptr_subscript_op_common.h
@@ -98,7 +98,7 @@ class run_subscript_op_tests {
         if constexpr (space == sycl::access::address_space::local_space) {
           sycl::local_accessor<T> exp_arr_acc{{array_size}, cgh};
           sycl::local_accessor<T> exp_arr_neg_acc{{array_size}, cgh};
-          cgh.single_task([=] {
+          cgh.parallel_for(sycl::nd_range<1>({1}, {1}), [=](auto item) {
             value_operations::assign(exp_arr_acc, exp_arr);
             value_operations::assign(exp_arr_neg_acc, exp_arr);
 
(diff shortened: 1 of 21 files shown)


tests/multi_ptr/multi_ptr_subscript_op_common.h
```diff
@@ -98,7 +98,7 @@ class run_subscript_op_tests {
         if constexpr (space == sycl::access::address_space::local_space) {
           sycl::local_accessor<T> exp_arr_acc{{array_size}, cgh};
           sycl::local_accessor<T> exp_arr_neg_acc{{array_size}, cgh};
-          cgh.single_task([=] {
+          cgh.parallel_for(sycl::nd_range<1>({1}, {1}), [=](auto item) {
```
bader, line 101: Why do we need this change?

bader (approved):

tests/multi_ptr/multi_ptr_subscript_op_common.h
steffenlarsen: According to the [spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:accessor.local), `local_accessor` is only allowed in `nd_range` `parallel_for`:
> A local accessor must not be used in a [SYCL kernel function](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sycl-kernel-function) that is invoked via single_task or via the simple form of parallel_for that takes a range parameter. In these cases submitting the kernel to a queue must throw a synchronous exception with the errc::kernel_argument error code.

keryell (approved):
Thanks.
Feel free to change `dtream` to `stream` in latest commit message.

tests/multi_ptr/multi_ptr_subscript_op_common.h
keryell: Where is the comment in the code explaining this? :-)

kopylovanat:
> Thanks. Feel free to change `dtream` to `stream` in latest commit message.

Thank you! Fixed.

PR #669, opened 2023-05-29
Reduce accessor build and run time

kopylovanat:
Reduced the accessor test build and run time by moving checks for some types to FULL_CONFORMANCE mode.
In regular mode vec and marray type as accessor data type run only for int and for the reduced number of sizes.

### DIFF
--- a/tests/accessor/accessor_common.h
+++ b/tests/accessor/accessor_common.h
@@ -197,6 +197,18 @@ inline auto get_conformance_type_pack() {
 #endif  // SYCL_CTS_ENABLE_FULL_CONFORMANCE
 }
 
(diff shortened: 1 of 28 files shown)


bader (approved):

tests/accessor/accessor_common.h
```diff
@@ -197,6 +197,18 @@ inline auto get_conformance_type_pack() {
 #endif  // SYCL_CTS_ENABLE_FULL_CONFORMANCE
 }
 
+template <template <typename, typename...> class action,
```
keryell, line 200: Add a function comment explaining the intent

keryell (approved):
Thanks.

PR #670, opened 2023-05-30
Add test plan for address space deduction

vladimirkhashev:
Added a test plan of address space deduction for C++ pointers used in kernel function that described in section [5.9. Address-space deduction](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_address_space_deduction) of the specification.

### DIFF
--- a/test_plans/pointer_address_space_deduction.asciidoc
+++ b/test_plans/pointer_address_space_deduction.asciidoc
@@ -1,0 +1,47 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan of address space deduction for C++ pointer type variables in kernel function
+
+This is a test plan of address space deduction for C++ pointers described in
+section https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_address_space_deduction[5.9. Address-space deduction]
+of the specification.
+
+== Testing scope
+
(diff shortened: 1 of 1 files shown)


test_plans/pointer_address_space_deduction.asciidoc
```diff
@@ -0,0 +1,49 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan of address space deduction for C++ pointer type variables in kernel function
```
gmlueck, line 4: My personal belief is that this area of the spec is too poorly defined to write CTS tests.  The tests proposed in this plan are extremely simple, but even these simple tests do not seem to be guaranteed by the spec.  For example, the section "Checking the address space deducing for function return value of pointer type" defines a function that does pointer arithmetic on a function parameter.  I don't see anything in [section 5.9.4](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:inferredAddressSpace) that guarantees inference rules for pointer arithmetic.

My guidance is to wait for a future version of the spec to clarify this area before writing CTS tests.

vladimirkhashev: Ok, got it, thank you. I'll close the PR for now.

PR #671, opened 2023-05-30
Cover gap for range by-value semantic

aleksmesh:
Add checks in kernel for range and refactor common code in common/semantics_by_value.h.

### DIFF
--- a/tests/common/semantics_by_value.h
+++ b/tests/common/semantics_by_value.h
@@ -45,10 +45,10 @@ enum class current_check : size_t {
   equal_other_symmetry,
   not_equal_other,
   not_equal_other_symmetry,
-  SIZE  // This should be last
+  size  // This should be last
 };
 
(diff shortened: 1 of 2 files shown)


tests/common/semantics_by_value.h
```diff
@@ -30,49 +30,87 @@
...
+  equal_copy_symmetry,
+  not_equal_copy,
+  not_equal_copy_symmetry,
+  transitivity_equal,
+  transitivity_not_equal,
+  equal_other,
+  equal_other_symmetry,
+  not_equal_other,
+  not_equal_other_symmetry,
+  SIZE  // This should be last
```
keryell, line 48: The uppercase is strange regarding the coding style since it is not a macro.
Or at least I hope this is not a macro, by the way... :-)

keryell (approved):
Thanks.

PR #674, opened 2023-05-31
sycl::is_device_copyable tests implementation

AmirIpma:
Created tests for `sycl::is_device_copyable` type trait according to [test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/is_device_copyable.asciidoc)

### DIFF
--- a/tests/is_device_copyable/is_device_copyable_api.cpp
+++ b/tests/is_device_copyable/is_device_copyable_api.cpp
@@ -21,6 +21,7 @@
 #include "../common/common.h"
 #include "../common/type_list.h"
 
+#if SYCL_DEVICE_COPYABLE == 1
 class kernel_name;
 
 /** @brief User defined class, must satisfy the
@@ -100,7 +101,6 @@ TEST_CASE("is_device_copyable UnaryTrait requirements",
 
 TEST_CASE("is_device_copyable specialization for user defined class",
           "[is_device_copyable]") {
-#if SYCL_DEVICE_COPYABLE == 1
   INFO(
       "User defined type must be device copyable after specialization for "
       "sycl::is_device_copyable<T>");
@@ -152,8 +152,10 @@ TEST_CASE("is_device_copyable specialization for user defined class",
     INFO("Check that device object has expected values on the host");
     CHECK(device_object == reference_object);
   }
+}
 
 #else
+TEST_CASE("Skip tests", "[is_device_copyable]") {
   SKIP("SYCL_DEVICE_COPYABLE is not defined!");
-#endif
 }
+#endif // SYCL_DEVICE_COPYABLE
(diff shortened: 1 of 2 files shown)


tests/is_device_copyable/is_device_copyable_api.cpp
```diff
@@ -0,0 +1,161 @@
...
+  if constexpr (std::is_same_v<T, char>) {
+    return 'c';
+  }
+  return 0;
+}
+
+TEST_CASE("is_device_copyable UnaryTrait requirements",
+          "[is_device_copyable]") {
+  using device_copyable_t = sycl::is_device_copyable<user_def_type>;
+  using integral_constant_t = std::integral_constant<bool, true>;
```
keryell, line 75: Besides the fact I find your name unclear, you are redefining `std::true_type`.

tests/is_device_copyable/is_device_copyable_api.cpp
```diff
@@ -0,0 +1,161 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+*******************************************************************************/
+
+#include "../common/common.h"
+#include "../common/type_list.h"
+
+#if SYCL_DEVICE_COPYABLE == 1
```
ProGTX, line 24:
```suggestion
#if defined(SYCL_DEVICE_COPYABLE) && (SYCL_DEVICE_COPYABLE == 1)
```
From the spec:
>  If the implementation has this support, it must predefine the preprocessor macro SYCL_DEVICE_COPYABLE to 1, and it must not predefine this preprocessor macro if it does not have this support.

We could potentially also add a check that if the macro is defined, it must be defined to 1. But the above suggestion should be sufficient to make impls that don't support that pass this test.

tests/is_device_copyable/is_device_copyable_api.cpp
```diff
@@ -0,0 +1,161 @@
...
+  const user_def_type reference_object(
+      expected_val<int>(), expected_val<float>(), expected_val<char>());
+
+  bool host_res;
+  user_def_type host_object(reference_object);
+  user_def_type device_object;
+  {
+    sycl::buffer<bool, 1> host_res_buf(&host_res, {1});
+    sycl::buffer<user_def_type, 1> host_data_buf(&host_object, {1});
+    sycl::buffer<user_def_type, 1> dev_data_buf(&device_object, {1});
```
ProGTX, line 119:
```suggestion
    sycl::buffer<bool> host_res_buf(&host_res, {1});
    sycl::buffer<user_def_type> host_data_buf(&host_object, {1});
    sycl::buffer<user_def_type> dev_data_buf(&device_object, {1});
```

ProGTX (approved):

tests/is_device_copyable/is_device_copyable_api.cpp
AmirIpma: Applied.

tests/is_device_copyable/is_device_copyable_api.cpp
AmirIpma: I don't think it was really necessary, but applied.

keryell (approved):
Thanks.

PR #680, opened 2023-06-02
sycl::item missing operators tests

AmirIpma:
Added `operator size_t()` and `operator item<Dimensions, true>()` tests for `item`.

### DIFF
--- a/tests/item/item_api.cpp
+++ b/tests/item/item_api.cpp
@@ -144,7 +144,6 @@ class kernel_item {
     type_acc[get_range] =
         std::is_same_v<sycl::range<dims>, decltype(localRange)>;
 
-
 #if SYCL_CTS_ENABLE_DEPRECATED_FEATURES_TESTS
     sycl::id<dims> offset = item.get_offset();
     api_acc_deprecated[item] = (offset == offset_exp) &&
--- a/tests/item/item_api.cpp
+++ b/tests/item/item_api.cpp
@@ -23,6 +23,7 @@
 #include "catch2/catch_test_macros.hpp"
 
 #include "../common/common.h"
+#include "../common/disabled_for_test_case.h"
 
 namespace item_api_test {
 using namespace sycl_cts;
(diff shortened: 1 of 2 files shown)


tests/item/item_api.cpp
```diff
@@ -148,43 +217,70 @@ void test_item() {
...
+  std::array<std::array<bool, nSize>, nMethodsCount> apiData;
+  std::for_each(apiData.begin(), apiData.end(),
+                [](std::array<bool, nSize>& arr) { arr.fill(true); });
+
+  std::array<std::array<bool, nSize>, nMethodsCount> typeData;
+  std::for_each(typeData.begin(), typeData.end(),
+                [](std::array<bool, nSize>& arr) { arr.fill(true); });
 
-  std::vector<int> dataOutDeprecated(nSize);
+  std::vector<int> apiDataDeprecated(nSize);
```
keryell, line 228: Just curious: why a `vector` here and `array` elsewhere?

tests/item/item_constructors.cpp
```diff
@@ -31,7 +31,8 @@ template <int numDims>
 struct item_constructors_kernel;
 
 enum class current_check {
-  copy_constructor,
+  copy_constructor = 0,
```
keryell, line 34: Isn't it the default?

tests/item/item_constructors.cpp
```diff
@@ -124,6 +125,15 @@ class TEST_NAME : public util::test_base {
                 check_equality(successAcc, current_check::copy_constructor,
                                copied, item);
 
+                if constexpr (numDims == 1) {
+                  // Check deafult Dimension and offset
```
keryell, line 129:
```suggestion
                  // Check default Dimension and offset
```
Use a spell checker?

tests/item/item_api.cpp
AmirIpma: I prefer to use vectors rather than arrays, but they are not suitable with bool type. If it bothers you I can replace all with arrays

tests/item/item_constructors.cpp
AmirIpma: Yes, it is.

tests/item/item_constructors.cpp
AmirIpma: Excuse me, I will work on my spelling

PR #682, opened 2023-06-02
sycl::join exception throwing tests

AmirIpma:
Implemented tests for `sycl::join` `errc::invalid` exception matching when passing a vector with different context as [spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_joining_kernel_bundles) says.

### DIFF
--- a/tests/kernel_bundle/sycl_join_kernel_bundles_with_diff_ctx.cpp
+++ b/tests/kernel_bundle/sycl_join_kernel_bundles_with_diff_ctx.cpp
@@ -1,0 +1,68 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2023 The Khronos Group Inc.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


tests/kernel_bundle/sycl_join_kernel_bundles_with_diff_ctx.cpp
```diff
@@ -0,0 +1,68 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
```
keryell, line 4: Comments inside comments are even more confusing.
```suggestion
  
   SYCL 2020 Conformance Test Suite
  
```

keryell (approved):
Thanks.

tests/kernel_bundle/sycl_join_kernel_bundles_with_diff_ctx.cpp
bader: @keryell, this header is recommended in our [coding guidelines](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/docs/README.adoc#copyright-header) and it's the same for all source code files in this project. I suggest we merge this PR "as is" and open a separate issue for changing the guidelines and updating the source code base. Does it sound okay to you?

keryell: I completely forgot about this coding guide. I should look at it. :-)
Yes, you can merge this.

PR #684, opened 2023-06-05
Reduce number of iterations in reduction tests

jzc:
In a case of `reduction_without_identity_param_fp16`, a reduction with a half multiplication functor using the "each_work_item_twice" kernel is invoked. The result should be the product of the list `0, 0, 1, 1, 2, 2, ..., 8, 8, 9, 9`. Depending on how the product is computed, the result can either be `0` (e.g. left to right multiplication) or `nan` (e.g. right to left will overflow for half and then compute `inf*0=nan`). Both of these should be allowed as the SYCL spec does specify the order the the reduction must be computed. However, the assertion checks the reduction result against result from `std::accumulate` (a left to right reduction), so there is a potential for a discrepancy (as is the case for DPC++). The `number_iterations` variable determines the input range used in the reduction tests, so this PR reduces it to `7` to remove the possibility of this discrepancy.

### DIFF
--- a/tests/reduction/reduction_common.h
+++ b/tests/reduction/reduction_common.h
@@ -33,7 +33,13 @@ namespace reduction_common {
 constexpr bool with_property{true};
 constexpr bool without_property{false};
 
-constexpr size_t number_iterations{7};
+// The buffer prepared from get_buffer will be the sequence 0, 1, ...,
+// number_iterations-1. This number should be small enough to prevent
+// overflow issues. (e.g. in a half multiplication reduction, for
+// number_iterations >= 10, then a left-to-right reduction results in
+// 0, while a right-to-left reduction results in nan because 9! * 0 =
+// inf * 0 = nan.)
+constexpr size_t number_iterations{4};
 
 constexpr int identity_value{0};
 
--- a/tests/reduction/reduction_common.h
+++ b/tests/reduction/reduction_common.h
@@ -33,7 +33,7 @@ namespace reduction_common {
 constexpr bool with_property{true};
 constexpr bool without_property{false};
 
-constexpr size_t number_iterations{10};
+constexpr size_t number_iterations{7};
 
 constexpr int identity_value{0};
 


bader (approved):
I think `8!` can be represented in half as non-`inf` value.
At the same time, why using 7 is better than 3 or 4? Can we reduce it further to the minimum possible value, which is big enough to test the correctness of reduction implementation?

jzc:
> I think `8!` can be represented in half as non-`inf` value. At the same time, why using 7 is better than 3 or 4? Can we reduce it further to the minimum possible value, which is big enough to test the correctness of reduction implementation?

`8!` can be represented as a half, although `8!*2` is not, which is needed for the "each_work_item_twice" case. I choose 7 because I that is the first number where `n!*2` can be represented as a half. Although, I overlooked the fact that the range used in the test is an exclusive range, so  with `number_iterations = 7`, really the test is computing up to `6!*2`. However, the choice is essentially arbitrary, as long as it won't run into overflow issues. Any thoughts on whether to change the value to something other than 7?

keryell:
Whatever you want but please add a big comment explaining the rationale of the magic numbers. :-)

bader:
I would use the smallest number which provides the same coverage. Reducing from 10 to 7 seems to be fine. I didn't inspect the code carefully, but I guess 4 should work as well. 
@aleksmesh, @kopylovanat, thoughts?

kopylovanat:
> I would use the smallest number which provides the same coverage. Reducing from 10 to 7 seems to be fine. I didn't inspect the code carefully, but I guess 4 should work as well. @aleksmesh, @kopylovanat, thoughts?

I agree that 4 seems to be enough to check correctness.

bader:
> Whatever you want but please add a big comment explaining the rationale of the magic numbers. :-)

@keryell, please, confirm that you are okay with the comment added with https://github.com/KhronosGroup/SYCL-CTS/pull/684/commits/919eb745941455bb63f645c741465ed9afa2d96a.

keryell (approved):
Great comment. Go on. :-)

PR #686, opened 2023-06-06
Test plan for oneapi extension local memory

vladimirkhashev:
Add test plan for `sycl_ext_oneapi_local_memory` extension according to the [extension specification](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_local_memory.asciidoc)

### DIFF
--- a/test_plans/oneapi_local_memory.asciidoc
+++ b/test_plans/oneapi_local_memory.asciidoc
@@ -65,39 +65,75 @@ We should check that the allocating memory is actually available for all
 work-items in the work-group and the allocated objects are being created
 correctly.
 
+To test correctly initialization of the custom type object after call of _(2)_ 
+the custom type should have default constructor which initializes custom type
+data members by predefined known values.
+Apart from the default constructor the custom type should provide constructors
+taking one, two and three arguments to provide better testing for the args
+parameter pack to _(1)_.
+
 === Test description
 
 ==== Check initial value of allocated object and the type of returned value
 
 Define and submit a work-group data parallel SYCL kernel with work group size
-1.
+`n`.
 
 For the `group_local_memory(Group g, Args&&... args)` in kernel scope do the
 following:
 
 * Allocate local memory for an object of type `T`
-* Pass the initial value to the allocated object constructor through the `args`
-  parameter pack
+* Pass the chosen set of initial values to the constructor of type `T` through
+  the `args` parameter pack
 * Check the type of allocating function return value
-* Check that returned value is equal to the predefined initial value
+* Check the allocated object of type `T` has the predefined initial value
+
+For the custom type test should be performed for sets of initial values
+containing one, two and three values.
 
 For the `group_local_memory_for_overwrite(Group g)` in kernel scope do the
 following:
 
 * Allocate local memory for an object of type `T`
 * Check the type of allocating function return value
+* If the `T` is the custom type, check that object default constructor
+  was called, by comparing the allocated object data members with expected
+  values
+
+==== Check correctly initialization with different set of initial values in each work-group
+
+Define and submit a work-group data parallel SYCL kernel with work group size
+`n` and total execution range `num_work_group x n`.
+
+In kernel scope do the following:
+
+* Allocate local memory for an object of type `T` using function _(1)_
+* Each work-group chooses a different set of initial values used to
+  construct `T`
+* All work-items in a group pass the same set of initial values through the
+  `args` parameter pack
+* Work-items in each group check that the object of type `T` has the expected
+  initial value
+
+For the custom type test should be performed for sets of initial values
+containing one, two and three values.
 
 ==== Check availability of local memory by all work-items in the work-group
 
 Define and submit a work-group data parallel SYCL kernel with work group size
-n.
+`n`.
 
 In kernel scope do the following:
 
 * Allocate local memory for an `array[n]` of objects of type `T`
-* Assign the value `val` which is defined as `nd_item.get_local_linear_id()`
-  to the `array[(n - 1) - val]` (where n is work group size)
+* Assign the value `nd_item.get_local_linear_id() & 1` to the 
+  `array[(n - 1) - nd_item.get_local_linear_id()]` (where `n` is work group size)
 * Use group barrier to synchronize all work-items
-* Check the value `array[val]` is equal to `(n - 1) - val` for each work-item
+* Check the value `array[nd_item.get_local_linear_id()]` is equal to
+  `(n - 1) - nd_item.get_local_linear_id()` for each work-item
 
 Perform that test for allocating function _(1)_ and _(2)_.
+
+To perform this test with an object of the custom type, the custom type must
+provide a conversion constructor from `size_t` and
+`operator==(const CustomType&)`.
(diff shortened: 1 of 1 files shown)


test_plans/oneapi_local_memory.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+
+Let `T` is the type of an object to be allocated within local memory space.
+
+Tests are performed using the following types as `T`:
+
+In regular mode:
+
+* `int`
+* `float`
+* custom trivially destructible type
```
gmlueck, line 32: If you plan to use this custom type with `group_local_memory_for_overwrite` then the custom type must have a default constructor.

It would also be nice to test the case when the constructor takes more than one argument, so I think the custom type should also have a constructor that takes several arguments.  This will provide better testing for the `args` parameter pack to `group_local_memory`.  Alternatively, you could have several custom types, one that has a default constructor, and another whose constructor takes several arguments.

test_plans/oneapi_local_memory.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+* Pass the initial value to the allocated object constructor through the `args`
+  parameter pack
+* Check the type of allocating function return value
+* Check that returned value is equal to the predefined initial value
+
+For the `group_local_memory_for_overwrite(Group g)` in kernel scope do the
+following:
+
+* Allocate local memory for an object of type `T`
+* Check the type of allocating function return value
```
gmlueck, line 88: For the custom type, you should also verify that the default constructor is executed.  For example, the default constructor could initialize the custom type with some specific value, and the test can verify that the object has that value.

Tagging @Pennycook here to make sure this is the intended behavior.  The spec says that `group_local_memory_for_overwrite` does "default initialization".  For a class type, this means that there must be a default constructor and "default initialization" will call that default constructor.  Is this what we intend, or did we expect that `group_local_memory_for_overwrite` simply allocates uninitialized memory?

test_plans/oneapi_local_memory.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+Define and submit a work-group data parallel SYCL kernel with work group size
+n.
+
+In kernel scope do the following:
+
+* Allocate local memory for an `array[n]` of objects of type `T`
+* Assign the value `val` which is defined as `nd_item.get_local_linear_id()`
+  to the `array[(n - 1) - val]` (where n is work group size)
+* Use group barrier to synchronize all work-items
+* Check the value `array[val]` is equal to `(n - 1) - val` for each work-item
```
gmlueck, line 101: It's not clear how this test will work when `T` is `bool` or the custom type.  If you want it to work for `char`, then `n` must be 127 or less.

test_plans/oneapi_local_memory.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+We should check that the allocating memory is actually available for all
+work-items in the work-group and the allocated objects are being created
+correctly.
+
+=== Test description
+
+==== Check initial value of allocated object and the type of returned value
+
+Define and submit a work-group data parallel SYCL kernel with work group size
+1.
```
gmlueck, line 73: I think it would be a more interesting test if the work-group size was greater than 1.  All work items can pass the same values for the `args` parameter pack.  Each work-item can then check the value of the object.

test_plans/oneapi_local_memory.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+
+For the `group_local_memory(Group g, Args&&... args)` in kernel scope do the
+following:
+
+* Allocate local memory for an object of type `T`
+* Pass the initial value to the allocated object constructor through the `args`
+  parameter pack
+* Check the type of allocating function return value
+* Check that returned value is equal to the predefined initial value
+
```
gmlueck, line 83: I'd suggest another test that stresses this line of the spec:

> All arguments in `args` must be the same for all work-items in the group.

This allows `args` to be different for each group.  The test could be like this:

* Allocate local memory for an object of type `T`
* Each work-group chooses a different set of initialization values used to construct `T`.
* All work-items in a group pass the same set of initial values through the `args` parameter pack.
* Work-items in each group check that the object of type `T` has the expected initial value.

test_plans/oneapi_local_memory.asciidoc
```diff
@@ -0,0 +1,103 @@
...
+=== Test description
+
+==== Check initial value of allocated object and the type of returned value
+
+Define and submit a work-group data parallel SYCL kernel with work group size
+1.
+
+For the `group_local_memory(Group g, Args&&... args)` in kernel scope do the
+following:
```
gmlueck, line 76: It would be interesting to test this API in the case when `T` is an array.  The spec isn't clear how `args` are interpreted in this case, though.  Can there be one `arg` for each element of the array?  Is each element of the array initialized with the entire `args` pack?

@Pennycook, can you clarify the spec here?

test_plans/oneapi_local_memory.asciidoc
vladimirkhashev: Maybe we can use values which are defined as `val & 1`? These are suitable for all cases.

gmlueck: That would solve the `bool` case, but you'd still need to clarify how it works with the custom type.  Another option is to limit this test to fewer types.

test_plans/oneapi_local_memory.asciidoc
Pennycook: I think this is the intended behavior.  We deliberately tried to align with C++ functions like `make_unique_for_overwrite`, which perform default initialization.

test_plans/oneapi_local_memory.asciidoc
Pennycook: I think the intent was that `args` wasn't allowed in the array case.

Looking at the definition of `make_unique` (see [cppreference](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique)), though, I'm now confused because they don't allow array types at all.  @rolandschulz, do you have any opinion on what the correct fix is here?  I don't remember why we decided to allow array types.

test_plans/oneapi_local_memory.asciidoc
vladimirkhashev: The custom type can have one or more data members. We can define conversion constructor from `size_t` for the custom type that assigns `val & 1` (that was passed as an argument) to all data members. And then we should define `operator==(const CustomType&)` for the custom type that compares values of all data members.

gmlueck: Yes, that could work.  Add this to the test plan.

test_plans/oneapi_local_memory.asciidoc
vladimirkhashev: Do you suggest adding such a test in addition, or instead of the current version, which also needs to be changed?

gmlueck: I was thinking that it would be an additional test.  If you prefer to change the existing test, that could be OK too, though.

vladimirkhashev: Added a related note.

vladimirkhashev (commented):
Applied the comments.

gmlueck (approved):
Test plan looks OK.

I'm following up offline with the spec authors to resolve the questions about the spec that came up in this review.

test_plans/oneapi_local_memory.asciidoc
Pennycook: Decided offline that we will support aggregate initialization.

PR #687, opened 2023-06-07
Fix unexpected failures of sycl::join exception tests

AmirIpma:
The `REQUIRE` macro causes tests to fail if the number of devices is less than 2, when it should skip tests instead.


fraggamuffin:
for 2 devices it should be guarded.
for 1 or 1+ device(s) it is special and need to say if your are conformant to

bader (approved):

PR #351, opened 2022-06-07
Remove stub from accessor tests

andreyromanof:
This removes using of `stub` classes from accessor tests.

As SYCL 2020 `accessors` were not ready at the moment of tests development, stub classes with the same api as `sycl::accessor` were used for the verification that tests can be compiled successfully. There is no need for the stub classes in future lifetime of tests, so they were removed.

PR #689, opened 2023-06-07
Add checks for ranges/id-like types member 'dimensions'

aleksmesh:
Covered gap with checks for 'dimensions' member of ranges/id-like types according to [4.9.1. Ranges and index space identifiers](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#ranges-identifiers) and PR [#351](https://github.com/KhronosGroup/SYCL-Docs/pull/351).

### DIFF
--- a/tests/common/range_index_space_id.h
+++ b/tests/common/range_index_space_id.h
@@ -78,7 +78,7 @@ void check_members_test(const std::string& type_name) {
             [=] { check_members<T, Dim>(res_acc); });
       });
     }
-    for (int i = 0; i < 2; /*check_count;*/ ++i) {
+    for (size_t i = 0; i < check_count; ++i) {
       INFO(get_error_string(i, type_name));
       CHECK(result[i]);
     }
@@ -87,7 +87,7 @@ void check_members_test(const std::string& type_name) {
                  "' and dim = " + std::to_string(Dim) + " on host";
   SECTION(section_name) {
     check_members<T, Dim>(result);
-    for (int i = 0; i < check_count; ++i) {
+    for (size_t i = 0; i < check_count; ++i) {
       INFO(get_error_string(i, type_name));
       CHECK(result[i]);
     }
--- a/tests/common/range_index_space_id.h
+++ b/tests/common/range_index_space_id.h
@@ -1,0 +1,97 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2017-2022 Codeplay Software LTD. All Rights Reserved.
+//  Copyright (c) 2022-2023 The Khronos Group Inc.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


tests/common/range_index_space_id.h
```diff
@@ -0,0 +1,97 @@
...
+  result[index] = true;
+}
+
+inline std::string get_error_string(int code, const std::string& type_name) {
+  return error_strings[code] + " for type T '" + type_name + "'";
+}
+
+template <typename T, int Dim, typename ResultArray>
+void check_members(ResultArray& result) {
+  if (std::is_same_v<const int, decltype(T::dimensions)>) {
```
ProGTX, line 53:
```suggestion
  if constexpr (std::is_same_v<const int, decltype(T::dimensions)>) {
```

tests/common/range_index_space_id.h
```diff
@@ -0,0 +1,97 @@
...
+    set_success_operation<check_codes::dimensions_type>(result);
+  }
+  if (Dim == T::dimensions) {
```
ProGTX, line 56:
```suggestion
  if constexpr (Dim == T::dimensions) {
```

tests/common/range_index_space_id.h
```diff
@@ -0,0 +1,97 @@
...
+  SECTION(section_name) {
+    {
+      sycl::buffer<bool, 1> res_buf(result, sycl::range(check_count));
+      queue.submit([&](sycl::handler& cgh) {
+        sycl::accessor res_acc(res_buf, cgh);
+        cgh.single_task<kernel_range_id<T, Dim>>(
+            [=] { check_members<T, Dim>(res_acc); });
+      });
+    }
+    for (int i = 0; i < 2; /*check_count;*/ ++i) {
```
ProGTX, line 81:
```suggestion
    for (int i = 0; i < check_count; ++i) {
```

ProGTX (approved):

tests/common/range_index_space_id.h
aleksmesh: Fixed. Thanks!

tests/common/range_index_space_id.h
aleksmesh: I prefer constexpr too, but in this case we'll get CI failure at compile time instead of fails at runtime. I thought runtime fails would be better. What do you think?

tests/common/range_index_space_id.h
aleksmesh: Same point as below.

tests/common/range_index_space_id.h
ProGTX: Hmm, that's generally a bit of a philosophical question of whether to do compile time checks at compile time (we certainly don't everywhere in the CTS), but in this case it would have to compile regardless of `constexpr`, because `set_success_operation` will still need to be run at runtime for the test to pass. It's a minor change anyway, I don't mind much either way.

PR #694, opened 2023-06-08
Fix optional kernel features test case

jzc:
The test case description is "Kernel does not use the tested feature but is decorated with the corresponding attribute
`[[sycl::device_has()]]`", but this call of `run_separate_lambda_with_accessor` tests lambda's that are not decorated with `[[sycl::device_has()]]` and that do not use any optional features, which makes it incorrectly assert that a exception with `kernel_not_supported` error code must be thrown.

### DIFF
--- a/tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
+++ b/tests/optional_kernel_features/kernel_features_device_has_exceptions.cpp
@@ -198,12 +198,6 @@ DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(ComputeCpp, hipSYCL)
   constexpr sycl::errc expected_errc = sycl::errc::kernel_not_supported;
 
   {
-    run_separate_lambda_with_accessor<kname, FeatureTypeT, FeatureAspectT,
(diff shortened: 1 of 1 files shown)


bader:
@aleksmesh, FYI.

bader:
> The test case description is "Kernel does not use the tested feature but is decorated with the corresponding attribute
> `[[sycl::device_has()]]`", but this call of `run_separate_lambda_with_accessor` tests lambda's that are not decorated with `[[sycl::device_has()]]` and that do not use any optional features, which makes it incorrectly assert that a exception with `kernel_not_supported` error code must be thrown.

I'd like to hear @aleksmesh opinion as the test author. It sounds like a bug the test which we should fix rather than remove the test case. Am I right?

aleksmesh:
@bader @jzc I apologize, it's my mistake. Fixed it in PR [#700](https://github.com/KhronosGroup/SYCL-CTS/pull/700).

bader:
@jzc, I suggest we go with PR #700, so I'm closing this one.

PR #695, opened 2023-06-09
Kernel attributes tests implementation

AmirIpma:
Created tests for kernel attributes according to [test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/kernel_attributes.asciidoc)

### DIFF
--- a/tests/kernel/kernel_attributes_vec_hint.cpp
+++ b/tests/kernel/kernel_attributes_vec_hint.cpp
@@ -29,7 +29,6 @@ using namespace kernel_attributes;
 #ifndef SYCL_CTS_COMPILING_WITH_DPCPP
 
 #define RUN_TEST(K_NAME1, K_NAME2, K_NAME3, VEC_T, FUNC1, FUNC2, FUNC3)       \
-                                                                              \
   {                                                                           \
     auto queue = sycl_cts::util::get_cts_object::queue();                     \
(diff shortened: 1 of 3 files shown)


tests/kernel/kernel_attributes_vec_hint.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+#include "../common/common.h"
+#include "../common/disabled_for_test_case.h"
+#include "kernel_attributes.h"
+
+using namespace kernel_attributes;
+
+// FIXME: enable when [[sycl::vec_type_hint(<type>)]] is implemented
+// https://github.com/intel/llvm/issues/9836
+#ifndef SYCL_CTS_COMPILING_WITH_DPCPP
```
ProGTX, line 29: Shouldn't the test compile even if the impl doesn't support the attribute? Unknown attributes should be just ignored by the C++ compiler.

ProGTX (approved):

tests/kernel/kernel_attributes_vec_hint.cpp
AmirIpma: I have compilation failures with attributes shown in issue above

ProGTX: Interesting, definitely seems like a DPC++ bug then.

PR #709, opened 2023-06-14
Test plan for oneapi "queue empty" extension

vladimirkhashev:
Added test plan according to the [spec](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_queue_empty.asciidoc).

### DIFF
--- a/test_plans/oneapi_queue_empty.asciidoc
+++ b/test_plans/oneapi_queue_empty.asciidoc
@@ -45,7 +45,7 @@ queue are completed and `false` otherwise.
 ===== Test 1:
 
 * Create queue
-* Define and submit two kernels performing some actions in long loop, save
+* Define and submit a kernel performing some actions in long loop, save
   returned event E
 * Call `ext_oneapi_empty()`
 * If `E.get_info<command_execution_status>() != complete` check that returned
--- a/test_plans/oneapi_queue_empty.asciidoc
+++ b/test_plans/oneapi_queue_empty.asciidoc
@@ -5,7 +5,7 @@
 
 This is a test plan for the extension adding a new API that tells whether a
 queue is empty. The extension is described in
-https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_local_memory.asciidoc[sycl_ext_oneapi_local_memory.asciidoc].
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_queue_empty.asciidoc[sycl_ext_oneapi_queue_empty.asciidoc].
 
 == Testing scope
 
@@ -13,14 +13,6 @@ https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_o
 
 All of the tests described below are performed only on the default device that
 is selected on the CTS command line.
-
-[NOTE]
-====
-Currently the extension support for OpenCL backend is limited, API 
-introduced by this extension can be called only for `in-order` queues which
-doesn’t have `discard_events` property. So the queue used for the tests should
-be created with `in-order` property if backend is OpenCL.
-====
 
 === Feature test macro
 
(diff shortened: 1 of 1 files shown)


test_plans/oneapi_queue_empty.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+All of the tests described below are performed only on the default device that
+is selected on the CTS command line.
+
+[NOTE]
+====
+Currently the extension support for OpenCL backend is limited, API 
+introduced by this extension can be called only for `in-order` queues which
+doesn’t have `discard_events` property. So the queue used for the tests should
+be created with `in-order` property if backend is OpenCL.
```
gmlueck, line 22: I don't think we should do this.  The CTS should test for the correct behavior according to the spec.  It should not work around bugs in the DPC++ implementation.  I expect this will lead to failures when the test is run on the OpenCL backend, but this is OK.  The failure will correctly point out a deficiency in the DPC++ implementation.

test_plans/oneapi_queue_empty.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+* Check that returned value is `true`
+
+==== `ext_oneapi_empty()` is called after commands submission
+
+===== Test 1:
+
+* Create queue
+* Define and submit two kernels performing some actions in long loop
+* Call `ext_oneapi_empty()`
+* Check that returned value is `false`
```
gmlueck, line 58: This test is unsafe because there is no guarantee that the kernel is still running by the time you call `ext_oneapi_empty` here.  I'd suggest a small modification:

```
event e = q.submit(/*...*/);
info::event_command_status status = e.get_info<info::event::command_execution_status>();
bool is_empty = q.ext_oneapi_empty();
if (status != info::event_command_status::complete)
  assert(!is_empty);
```

The status will almost never be `info::event_command_status::complete` because the kernel is long running.  However, in the rare case when the kernel finishes first, the test still does not fail.

test_plans/oneapi_queue_empty.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+* Create queue
+* Define two kernels performing some actions in long loop, let it be kernels
+  `A` and `B`
+* Kernel `B` must depends on kernel `A` so that it can only be started its work
+  after kernel `A` has completed
+* Submit kernel `A` and `B`, save events `Ea` and `Eb`
+* Call `wait()` for event `Ea`
+* Call `ext_oneapi_empty()`
+* Check that returned value is `false`
```
gmlueck, line 73: Same comment here.  You can use the status of event `Eb` in the same way I describe above.

test_plans/oneapi_queue_empty.asciidoc
```diff
@@ -0,0 +1,76 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl extension oneapi queue empty
+
+This is a test plan for the extension adding a new API that tells whether a
+queue is empty. The extension is described in
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_local_memory.asciidoc[sycl_ext_oneapi_local_memory.asciidoc].
```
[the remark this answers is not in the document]
vladimirkhashev: Fixed.

test_plans/oneapi_queue_empty.asciidoc
vladimirkhashev: Got it, removed.

test_plans/oneapi_queue_empty.asciidoc
vladimirkhashev: Added.

test_plans/oneapi_queue_empty.asciidoc
vladimirkhashev: Added.

gmlueck (approved):

PR #712, opened 2023-06-16
Base DPC++ docker container on the image from intel/llvm.

bader:
intel/llvm project provides daily built docker images with pre-built
DPC++ compiler. Using this image saves us time to build the compiler
from sources.


psalz (approved):
Good idea!

PR #713, opened 2023-06-16
[group_functions] Reduce test size to avoid overflow

Nuullll:
This avoids meaningless overflowed results for reducing over smaller types (e.g. `char`).

Signed-off-by: Yilong Guo <yilong.guo@intel.com>
Co-authored-by: Jie Zhang <jiez.zhang@intel.com>

### DIFF
--- a/tests/group_functions/group_reduce.h
+++ b/tests/group_functions/group_reduce.h
@@ -23,7 +23,8 @@
 #include "group_functions_common.h"
 #include <optional>
 
-constexpr size_t init = 1412;
+constexpr size_t init = 8;
+constexpr size_t test_size = 4;
 static const auto Dims = integer_pack<1, 2, 3>::generate_unnamed();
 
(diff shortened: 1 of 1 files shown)


vladimirkhashev:
Maybe it make sense to limit the size of the workgroup for `reduce` tests, because in the case when the size of the workgroup is large enough and the operation type is `sycl::multiplies<T>`, the test results will also be incorrect due to overflow. Also maybe init value (currently equals to 1442) for `reduce` operations can be changed that it can be represented by `char` type?

ProGTX (approved):

keryell (approved):
Thanks.
You need to fix the merge conflicts.

Nuullll:
Thanks. I'll update per @vladimirkhashev 's suggestion.

Nuullll:
Adding co-author: @jiezzhang

bader (approved):

PR #714, opened 2023-06-16
Test plan for "discard queue events" oneapi extension

vladimirkhashev:
Added test plan according to the [spec](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_discard_queue_events.asciidoc).

### DIFF
--- a/test_plans/oneapi_discard_queue_events.asciidoc
+++ b/test_plans/oneapi_discard_queue_events.asciidoc
@@ -24,13 +24,14 @@ be skipped if feature is not supported.
 
 Extension introduces new `ext::oneapi::property::queue::discard_events`
 property for `sycl::queue` and new enumerator value for the
-`sycl::info::event_command_status` enumeration. When the `queue` is created
-with the `discard_events` property the `queue` member functions returning
+`sycl::info::event_command_status` enumeration. When the queue is created
+with the `discard_events` property the queue member functions returning
 events will return _invalid_ events.
 Behavior of some _invalid_ event member functions is different from the default
 event behavior and needs to be tested. The behavior when _invalid_ event is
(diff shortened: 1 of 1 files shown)


test_plans/oneapi_discard_queue_events.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+property for `sycl::queue` and new enumerator value for the
+`sycl::info::event_command_status` enumeration. When the `queue` is created
+with the `discard_events` property the `queue` member functions returning
+events will return _invalid_ events.
+Behavior of some _invalid_ event member functions is different from the default
+event behavior and needs to be tested. The behavior when _invalid_ event is
+passed into `handler::depends_on()` function is also different and needs to be
+tested.
+
+=== Test description
```
gmlueck, line 35: All the tests listed here are negative tests -- checking various error conditions.  There should be some positive tests also:

* Do all of the following twice: with a queue constructed with `discard_events` and with a queue constructed with both `discard_events` and `in_order`:
  * Submit a single_task kernel, wait on the queue, validate that the kernel ran.
  * Submit a simple parallel_for kernel, wait on the queue, validate that the kernel ran.
  * Submit an nd-range kernel, wait on the queue, validate that the kernel ran.
  * Submit a single task kernel using assert (where the assert does not trigger), wait on the queue, validate that the kernel ran.
  * Submit a single task kernel using `stream`, wait on the kernel, validate that the kernel ran.
  * Submit a single task kernel using a buffer accessor, wait on the kernel, validate that the kernel ran.

vladimirkhashev: Added.

gmlueck (approved):
Looks good!

PR #715, opened 2023-06-16
Test plan for Intel's Extensions for Device Information

AmirIpma:
Created test plan for [extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_device_info.md)

### DIFF
--- a/test_plans/intel_device_info.asciidoc
+++ b/test_plans/intel_device_info.asciidoc
@@ -1,0 +1,142 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for Intel's Extensions for Device Information
+
+This is a test plan for the APIs described in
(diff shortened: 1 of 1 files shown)


test_plans/intel_device_info.asciidoc
```diff
@@ -0,0 +1,142 @@
...
+----
+auto devices = device::get_devices();
+
+std::for_each(devices.cbegin(), devices.cend(), [](const auto& dev) {
+  /* tests */
+});
+----
+
+Get available devices by calling `get_devices()` and run the code in the sections below for each device.
+Code should consist checking if device has aspect for this `info` and checking type.
```
gmlueck, line 33: This paragraph contradicts the "Device coverage" section, where you say that the tests are only run on the device that is selected on the CTS command line.  Our normal policy is to run CTS tests only that selected device.  Why should this test be different?

AmirIpma: Forgot about that, removed.

gmlueck (approved):

PR #716, opened 2023-06-19
Cover nd_range semantic by-value gaps

aleksmesh:
Added checks for copy and move constructors, added tests in kernel function and refactored.


bader (approved):

keryell (approved):
Thanks.

PR #719, opened 2023-06-20
Remove checks for bitwise group operations on floating point types

steffenlarsen:
The SYCL 2020 specification defines bitwise operations as being "equivalent" to the binary operators for the given operation. Since the operators are not explicitly defined for `float`, `double` or `sycl::half` the CTS should not test the group operations with these types and bitwise operations. This commit makes that distinction.


bader (approved):
Please, resolve merge conflicts.

ProGTX (approved):

PR #722, opened 2023-06-21
Test plan for intel_kernel_restrict_all extension

AmirIpma:
Created test plan for [intel_kernel_restrict_all](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_kernel_args_restrict.asciidoc) extension

### DIFF
--- a/test_plans/kernel_restrict_all.asciidoc
+++ b/test_plans/kernel_restrict_all.asciidoc
@@ -1,0 +1,54 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for kernel_restrict_all
+
+This is a test plan for the APIs described in https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_kernel_args_restrict.asciidoc[sycl_intel_kernel_restrict_all]
+
+== Testing scope
+
+=== Device coverage
+
+All of the tests described below are performed only on the default device that
+is selected on the CTS command line.
+
+=== Feature test macro
+
+All of the tests should use `#ifdef SYCL_EXT_INTEL_KERNEL_ARGS_RESTRICT` so they can be skipped
+if feature is not supported.
+
+== Tests
+
+All tests should use following methods of kernels submission unless stated otherwise:
+
+* parallel_for_work_group
+* parallel_for
+* single_task
+
+Kernel should be defined:
+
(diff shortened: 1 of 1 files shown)


keryell (approved):
I guess it should be tested in combination with USM since the kernel argument interface might be handled differently by the compiler.

test_plans/kernel_restrict_all.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+auto acc1 = out_buf_1.get_access<access_mode::write>(cgh);
+auto acc2 = out_buf_2.get_access<access_mode::write>(cgh);
+cgh.parallel_for<lambda_foo<dims>>(range<dims>(N),
+    [=](id<dims> wiid) [[intel::kernel_args_restrict]] {
+        acc1[wiid] = dims;
+        acc2[wiid] = dims * 2;
+    });
+});
+/* check values in buffers */
+----
```
gmlueck, line 54: I think it would be useful to add a test like this where the kernel arguments don't alias, but there is aliasing inside the kernel:

```
int func(int *p1, int *p2, int *p3) {
  *p1 = 1;
  *p2 = 2;
  *p3 = 3;
  return *p1;
}

void test(queue q) {
  int *ptr1 = malloc_device(1, q);
  int *ptr2 = malloc_device(1, q);
  q.single_task([=]() [[intel::kernel_args_restrict]] {
    int val = func(ptr1, ptr2, ptr1);
    if (val != 3)
      /* test fails */
  });
}
```

test_plans/kernel_restrict_all.asciidoc
```diff
@@ -0,0 +1,54 @@
...
+
+Kernel should be defined:
+
+* through lambda in submission call
+* through a separate functor
+* through a separate lambda
+
+For following dimensions `dims = 1, 2, 3`:
+
+* Create kernel with attribute `[[intel::kernel_args_restrict]]`
```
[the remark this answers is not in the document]
AmirIpma: Removed square brackets.

test_plans/kernel_restrict_all.asciidoc
AmirIpma: Applied.

gmlueck (approved):

PR #725, opened 2023-06-21
Test plan for intel extension queue index

vladimirkhashev:
Add test plan according to the [spec](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_queue_index.asciidoc)

### DIFF
--- a/test_plans/intel_queue_index.asciidoc
+++ b/test_plans/intel_queue_index.asciidoc
@@ -1,0 +1,70 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl extension intel queue index
+
+This is a test plan for the extension exposes an "index" to a device’s work
+submission queue. The extension is described in
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_queue_index.asciidoc[sycl_ext_intel_queue_index.asciidoc].
+
+== Testing scope
+
+=== Device coverage
+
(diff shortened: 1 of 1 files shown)


keryell (approved):

test_plans/intel_queue_index.asciidoc
```diff
@@ -0,0 +1,70 @@
...
+==== Kernel execution check
+
+Do the following for each queue index available for the device:
+
+* Create a queue with one of the available queue index
+* Submit a `single_task` kernel, wait on the queue, validate that the kernel
+  ran
+* Submit a simple `parallel_for` kernel, wait on the queue, validate that the
+  kernel ran
+* Submit an `nd-range` kernel, wait on the queue, validate that the kernel ran
```
gmlueck, line 70: I think it would be more interesting to submit one kernel for each index, and then wait for them all to complete:

> * Query the number of queue indices available on the device.
> * Create one queue for the device for each index
> * Submit a `single_task` kernel to each queue
> * Wait for them all to complete and validate that each kernel ran
>
> Repeat the above test using a simple `parallel_for` kernel instead of `single_task`.
>
> Repeat the above test using an ND-range `parallel_for` kernel instead of `single_task`.

vladimirkhashev: Done.

gmlueck (approved):

PR #727, opened 2023-06-23
mem_channel extension test plan

AmirIpma:
Created test plan for [mem_channel](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_mem_channel_property.asciidoc) extension.

### DIFF
--- a/test_plans/mem_channel.asciidoc
+++ b/test_plans/mem_channel.asciidoc
@@ -40,8 +40,8 @@ std::array<int, 10> arr2{0};
     buffer<T, 1> buf2(arr2.data(), range{10}, mc_object);
 
     queue.submit([&](handler& cgh) {
-        sycl::accessor acc1{buf1, cgh, access_mode::write};
-        sycl::accessor acc2{buf2, cgh, access_mode::write};
+        sycl::accessor acc1{buf1, cgh, access_mode::write_only};
+        sycl::accessor acc2{buf2, cgh, access_mode::write_only};
         cgh.parallel_for([=](id<1> i) {
             // some data manipulations
         });
--- a/test_plans/mem_channel.asciidoc
+++ b/test_plans/mem_channel.asciidoc
@@ -40,8 +40,8 @@ std::array<int, 10> arr2{0};
     buffer<T, 1> buf2(arr2.data(), range{10}, mc_object);
 
     queue.submit([&](handler& cgh) {
-        sycl::accessor acc1{buf1, cgh, write_only};
-        sycl::accessor acc2{buf2, cgh, write_only};
+        accessor acc1{buf1, cgh, write_only};
+        accessor acc2{buf2, cgh, write_only};
         cgh.parallel_for([=](id<1> i) {
             // some data manipulations
         });
--- a/test_plans/mem_channel.asciidoc
+++ b/test_plans/mem_channel.asciidoc
@@ -1,0 +1,63 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for mem_channel
+
+This is a test plan for the APIs described in
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_mem_channel_property.asciidoc[sycl_ext_intel_mem_channel]
+
+== Testing scope
+
+=== Device coverage
+
+All of the tests described below are performed only on the default device that
+is selected on the CTS command line.
+
+=== Feature test macro
+
+All of the tests should use `#ifdef SYCL_EXT_INTEL_MEM_CHANNEL_PROPERTY` so they can be skipped
+if feature is not supported.
+
+== Tests
+
+=== buffer with property
+
+* Create a property list with `property::buffer::mem_channel` property;
+* Create a `mem_channel` instance using `property::buffer::mem_channel::mem_channel(cl_uint channel)` constructor;
+* Create buffers sing these property list and property instance;
+* Verify that `buffer::has_property<mem_channel>()` returns true;
+* Verify that `buffer::get_property<mem_channel>()` does not throw any exceptions.
+* Run the code below to make sure that buffers work correctly:
+[source, c++]
+----
+property_list pl = {mem_channel(0)};
+mem_channel mc_object(1);
+
+std::array<int, 10> arr1;
+std::array<int, 10> arr2;
+{
+    buffer<T, 1> buf1(arr1.data(), range{10}, pl);
+    buffer<T, 1> buf2(arr2.data(), range{10}, mc_object);
+
+    queue.submit([&](handler& cgh) {
+        auto acc1 = buf1.get_access<access_mode::read_write>(cgh);
+        auto acc2 = buf2.get_access<access_mode::read_write>(cgh);
+        cgh.parallel_for([=](id<1> i) {
+            // some data manipulations
+        });
+    });
+}
+// verify data in arrays
+----
+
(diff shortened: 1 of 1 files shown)


test_plans/mem_channel.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+
+std::array<int, 10> arr1;
+std::array<int, 10> arr2;
+{
+    buffer<T, 1> buf1(arr1.data(), range{10}, pl);
+    buffer<T, 1> buf2(arr2.data(), range{10}, mc_object);
+
+    queue.submit([&](handler& cgh) {
+        auto acc1 = buf1.get_access<access_mode::read_write>(cgh);
+        auto acc2 = buf2.get_access<access_mode::read_write>(cgh);
```
keryell, line 44: To update to a `sycl::accessor` constructor.

test_plans/mem_channel.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+* Verify that `buffer::has_property<mem_channel>()` returns true;
+* Verify that `buffer::get_property<mem_channel>()` does not throw any exceptions.
+* Run the code below to make sure that buffers work correctly:
+[source, c++]
+----
+property_list pl = {mem_channel(0)};
+mem_channel mc_object(1);
+
+std::array<int, 10> arr1;
+std::array<int, 10> arr2;
```
keryell, line 37: To initialize to some values.

test_plans/mem_channel.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+
+* Use property list and `mem_channel` object to call `get_channel() member function`:
+[source, c++]
+----
+auto channel_1 = pl.get_property<mem_channel>().get_channel();
+auto channel_2 = mc_object.get_channel();
+----
+
+* Check that `get_channel()` returns expected values
+* Check that `std::is_same_v<decltype(channel_1), cl_uint>` and `std::is_same_v<decltype(channel_1), cl_uint>` are `true`.
```
keryell, line 63: I guess the extension spec itself should be updated to use `unsigned int` or `std::uint32_t` instead.

test_plans/mem_channel.asciidoc
AmirIpma: Do you mean using deduction tags?

test_plans/mem_channel.asciidoc
AmirIpma: Do I need to create an issue to change return type?

test_plans/mem_channel.asciidoc
AmirIpma: Applied.

test_plans/mem_channel.asciidoc
gmlueck: We generally prefer to construct accessors like this:

```
accessor acc1{buf1, cgh};
accessor acc2{buf2, cgh};
```

This is much less verbose and easier to read.  Since `read_write` is the default, you don't need a deduction tag.  If you want to show it anyway for clarity, then do this:

```
accessor acc1{buf1, cgh, read_write};
accessor acc2{buf2, cgh, read_write};
```

AmirIpma: @gmlueck, done.

test_plans/mem_channel.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+
+std::array<int, 10> arr1{0};
+std::array<int, 10> arr2{0};
+{
+    buffer<T, 1> buf1(arr1.data(), range{10}, pl);
+    buffer<T, 1> buf2(arr2.data(), range{10}, mc_object);
+
+    queue.submit([&](handler& cgh) {
+        sycl::accessor acc1{buf1, cgh, access_mode::write};
+        sycl::accessor acc2{buf2, cgh, access_mode::write};
```
gmlueck, line 44: The third parameter must be one of the [deduction tags](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_deduction_tags), which are not the same as the `access_mode` enums.  Therefore, do one of these:

```
sycl::accessor acc1{buf1, cgh, sycl::write_only};
sycl::accessor acc2{buf2, cgh, sycl::write_only};
```

or:

```
sycl::accessor acc1{buf1, cgh, sycl::read_write};
sycl::accessor acc2{buf2, cgh, sycl::read_write};
```

gmlueck: Sorry, this is still not right.  The word "access_mode" is wrong.  The identifier "write_only" is directly in the "sycl" namespace.  See the example above.

AmirIpma: Excuse me, didn't check

test_plans/mem_channel.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+
+std::array<int, 10> arr1{0};
+std::array<int, 10> arr2{0};
+{
+    buffer<T, 1> buf1(arr1.data(), range{10}, pl);
+    buffer<T, 1> buf2(arr2.data(), range{10}, mc_object);
+
+    queue.submit([&](handler& cgh) {
+        sycl::accessor acc1{buf1, cgh, write_only};
+        sycl::accessor acc2{buf2, cgh, write_only};
```
keryell, line 44: I am not a great fan of `using namespace` but since this choice seems to be made here:
```suggestion
        accessor acc1{buf1, cgh, write_only};
        accessor acc2{buf2, cgh, write_only};
```

test_plans/mem_channel.asciidoc
```diff
@@ -0,0 +1,63 @@
...
+* Verify that `buffer::has_property<mem_channel>()` returns true;
+* Verify that `buffer::get_property<mem_channel>()` does not throw any exceptions.
+* Run the code below to make sure that buffers work correctly:
+[source, c++]
+----
+property_list pl = {mem_channel(0)};
+mem_channel mc_object(1);
+
+std::array<int, 10> arr1{0};
+std::array<int, 10> arr2{0};
```
keryell, line 37:
```suggestion
std::array<int, 10> arr1{};
std::array<int, 10> arr2{};
```

keryell (approved):
Thanks.

GarveyJoe:
> The extension says there is an aspect aspect::ext_intel_mem_channel. What does this aspect mean? Can we test it somehow in this test plan?

I believe our implementation responds to the aspect query by telling you if the device does something useful with the mem_channel property as opposed to telling you whether the device supports the extension (which all our devices can trivially support as the property is a hint).  I think we can remove the aspect, but I think we're internally relying on it in the runtime so we'll have to change that code to not use the aspect.  

> One of the comments above notes that it does not make sense for the extension to use the OpenCL C type cl_uint, and I agree. Can this be changed in the extension spec to either unsigned int or uint32_t?

Yes, uint32_t would be fine.  This is a relic from when the OpenCL types were defined in the SYCL spec, but since they no longer are, this should be updated.

bader:
@gmlueck, ping.

gmlueck (approved):
The test plan looks OK, so I approved it.

We should make some changes to the extension spec (and implementation), though:

* Remove use of `cl_uint` from the API and use `uint32_t` instead.
* Either deprecate or remove the `ext_intel_mem_channel` aspect.

@GarveyJoe is this something your team can do soon, or should we open an internal tracker for this?

PR #729, opened 2023-06-26
Remove invalid use of get_pointer on placeholder accessor

steffenlarsen:
The SYCL specification states that get_pointer on an accessor or local_accessor can only be called inside a command. This commit fixes a use of get_pointer outside a command. Additionally it moves the surrounding code out of the encapsulating submission.


keryell:
It is unclear to me where `get_pointer` is.

steffenlarsen:
`get_pointer` is the method tested in `test_accessor_ptr`. Other tests use these on `host_accessor` which is perfectly valid.

steffenlarsen:
> Doesn't this reduce the amount of testing for other kinds of accessors as well?

The test explicitly calls the check on a placeholder accessor and the info text for the case states so too. We shouldn't be hitting the case for `host_accessor` or the like.

bader:
@keryell, do you have any follow-up questions?

keryell (approved):
Thanks.

PR #731, opened 2023-06-27
Fix faulty accessor tests

steffenlarsen:
This commit makes the following accessor test changes:
* Fixes the expected first and last element of an accessor with offset.
* Removes the exclusion of 0-dimensional accessor from the swap test.
* Fixes the expected first value in the iterator tests for 0-dimensional local_accessor.


keryell (approved):
Thanks.

ProGTX (approved):

PR #732, opened 2023-06-27
Test plan for intel extension cslice 

vladimirkhashev:
Added test plan according to the [spec](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_cslice.asciidoc).

### DIFF
--- a/test_plans/intel_cslice.asciidoc
+++ b/test_plans/intel_cslice.asciidoc
@@ -1,0 +1,127 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl extension intel compute slice (cslice)
+
+This is a test plan for the extension adding a new form of partitioning -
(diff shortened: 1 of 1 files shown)


gmlueck:
I'm not sure it makes sense to add a test for this extension to the CTS.  The features in this extension can only be tested by using the Level Zero driver and setting the environment variable `ZEX_NUMBER_OF_CCS`.  Is there any infrastructure in the CTS for setting this environment variable?

vladimirkhashev:
> I'm not sure it makes sense to add a test for this extension to the CTS. The features in this extension can only be tested by using the Level Zero driver and setting the environment variable `ZEX_NUMBER_OF_CCS`. Is there any infrastructure in the CTS for setting this environment variable?

I'm not sure, so close the PR for now.

PR #734, opened 2023-06-28
Tests for queue empty oneAPI extension

vladimirkhashev:
Added tests for queue empty oneapi extension according to the [test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/oneapi_queue_empty.asciidoc).


tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+TEST_CASE(
+    "Test for \"Queue empty\" extension, check extension functionality"
+    "after commands are submitted to the queue, test 1",
+    "[oneapi_queue_empty]") {
+#ifndef SYCL_EXT_ONEAPI_QUEUE_EMPTY
+  SKIP("SYCL_EXT_ONEAPI_QUEUE_EMPTY is not defined");
+#else
+  auto queue = sycl_cts::util::get_cts_object::queue();
+
+  size_t array[buf_size];
```
keryell, line 63: Do you need these arrays?

tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+  sycl::buffer<size_t, 1> array_buf(array, buf_size);
+
+  auto e = queue.submit([&](sycl::handler& cgh) {
+    auto array_acc =
+        array_buf.template get_access<sycl::access_mode::write>(cgh);
```
keryell, line 68: Use `sycl::accessor` constructor instead everywhere.

keryell (changes requested):

tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
vladimirkhashev: Done.

tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
vladimirkhashev: I added these arrays to avoid compiler optimizations in kernel code.

bader:
@keryell, ping.

tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
keryell: Which compiler optimization?
But my point is that would it work with just
```c++
sycl::buffer<size_t> array_buf(buf_size);
```
instead?

tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+    "Test for \"Queue empty\" extension, check extension functionality"
+    "after commands are submitted to the queue, test 1",
+    "[oneapi_queue_empty]") {
+#ifndef SYCL_EXT_ONEAPI_QUEUE_EMPTY
+  SKIP("SYCL_EXT_ONEAPI_QUEUE_EMPTY is not defined");
+#else
+  auto queue = sycl_cts::util::get_cts_object::queue();
+
+  size_t array[buf_size];
+  sycl::buffer<size_t, 1> array_buf(array, buf_size);
```
bader, line 64:
```suggestion
  sycl::buffer<size_t, 1> array_buf(buf_size);
```

tests/extension/oneapi_queue_empty/queue_empty_ext_test.cpp
bader: Done.

keryell (approved):
Thanks.

PR #737, opened 2023-06-28
Cover gaps for use_kernel_bunde()

vladimirkhashev:
Updated the tests for `use_kernel_bundle()` with incompatible device. Added check if exception with error code `kernel_not_supported` is thrown in case of `use_kernel_bundle()` call inside command group with incompatible device of secondary queue.


keryell (approved):
Thanks.

PR #740, opened 2023-06-28
sycl::sub_group by-value semantics tests

AmirIpma:
Implemented tests for `sycl::sub_group` by-value semantics, provide copy/move constructors/assignment and reflexivity/symmetry/transitivity checks.


keryell (approved):
Thanks.

ProGTX (approved):

PR #743, opened 2023-06-29
Update of test plan for extension oneapi local memory

vladimirkhashev:
Added array as one of the types to be tested.

### DIFF
--- a/test_plans/oneapi_local_memory.asciidoc
+++ b/test_plans/oneapi_local_memory.asciidoc
@@ -30,6 +30,7 @@ In regular mode:
 * `int`
 * `float`
 * custom trivially destructible type
+* `T[5]`, where `T` is one of above types 
 
 In full conformance mode:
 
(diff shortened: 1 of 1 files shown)


test_plans/oneapi_local_memory.asciidoc
```diff
@@ -113,13 +117,16 @@ In kernel scope do the following:
...
-containing one, two and three values.
+containing one, two and three values. If `T` is array, `args` parameter pack
+must have enough values to initialize each array element.
 
 ==== Check availability of local memory by all work-items in the work-group
 
+Skip the test if `T` is array.
+
 Define and submit a work-group data parallel SYCL kernel with work group size
 `n`.
```
gmlueck, line 132: Let's add one more test that is specific to arrays:

> === Check array initialization with fewer arguments than array elements
>
> Run this test only when `T` is an array of `N` elements.
>
> * Use `group_local_memory` to allocate an object of type `T` (which is an array).
> * Specify `M` "args" to that call, where `M < N`.
> * Verify that the first `M` elements of the array have the specified initial value.

gmlueck (approved):

PR #738, opened 2023-06-28
Tests for oneAPI bfloat16 extension

AmirIpma:
Created tests for [bfloat16](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_bfloat16.asciidoc) extension

PR #747, opened 2023-07-10
Avoid some bit_cast in bfloat16 test plan

gmlueck:
Addresses a review comment in the test implementation (#738).


keryell (approved):
Thanks for the simplification!

PR #750, opened 2023-07-17
[group_functions] Insert barrier after each validation

Nuullll:
Otherwise, the initialization performed by the following code may change the result buffer.

Signed-off-by: Yilong Guo <yilong.guo@intel.com>

### DIFF
--- a/tests/group_functions/group_barrier.cpp
+++ b/tests/group_functions/group_barrier.cpp
@@ -146,10 +146,10 @@ TEMPLATE_TEST_CASE_SIG("Group barriers", "[group_func][dim]", ((int D), D), 1,
                          "fence_scope) is wrong\n");
 
       // test of default barrier
-      local_acc[llid] = 0;
+      local_acc[llid] = llid;
       sycl::group_barrier(group);
 
-      if (local_acc[max_id - llid] != 0)
+      if (local_acc[max_id - llid] != max_id - llid)
         std::get<s::test>(group_barriers_acc[0]) = false;
       // make sure we check all items before moving on
       sycl::group_barrier(group);
@@ -166,11 +166,12 @@ TEMPLATE_TEST_CASE_SIG("Group barriers", "[group_func][dim]", ((int D), D), 1,
         auto& barrier = group_barriers_acc[i];
 
         if (std::get<s::support>(barrier)) {
-          local_acc[llid] = 0;
-          global_acc[llid] = 0;
+          local_acc[llid] = llid;
+          global_acc[llid] = llid;
           sycl::group_barrier(group);
 
-          if (local_acc[max_id - llid] != 0 || global_acc[max_id - llid] != 0)
+          if (local_acc[max_id - llid] != max_id - llid ||
+              global_acc[max_id - llid] != max_id - llid)
             std::get<s::test>(barrier) = false;
           sycl::group_barrier(group);
 
@@ -208,10 +209,10 @@ TEMPLATE_TEST_CASE_SIG("Group barriers", "[group_func][dim]", ((int D), D), 1,
                          "memory_scope fence_scope) is wrong\n");
 
       // test of default barrier
-      local_acc[llid] = 0;
+      local_acc[llid] = llid;
       sycl::group_barrier(sub_group);
 
-      if (local_acc[max_id - llid] != 0)
+      if (local_acc[max_id - llid] != max_id - llid)
         std::get<s::test>(sub_group_barriers_acc[0]) = false;
       sycl::group_barrier(sub_group);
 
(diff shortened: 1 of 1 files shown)


tests/group_functions/group_barrier.cpp
```diff
@@ -154,6 +154,8 @@ TEMPLATE_TEST_CASE_SIG("Group barriers", "[group_func][dim]", ((int D), D), 1,
 
       if (local_acc[max_id - llid] != 1)
         std::get<s::test>(group_barriers_acc[0]) = false;
+      // make sure we check all items before moving on
+      sycl::group_barrier(group);
```
bader, line 158: This change looks good to me, but I don't quite understand the surrounding code.
```c++
      // test of default barrier
      local_acc[llid] = 0;
      sycl::group_barrier(group);
      local_acc[llid] = 1;
      sycl::group_barrier(group);

      if (local_acc[max_id - llid] != 1)
        std::get<s::test>(group_barriers_acc[0]) = false;
      // make sure we check all items before moving on
      sycl::group_barrier(group);
```
1. initialize the local memory with `0`. If I get it right to make sure initial value is not `1`, which will check for later.
2. Fill local memory with `1`.
3. Check that memory is filled with `1` using reverse memory traversal order.

I think we must check that step `1` was successful and that local memory is filled with `0` before we can write `1`.

Later we test sub-group barrier in the same kernel, using the same local memory. The code is copy-pasted, but I don't think we need to apply the same memory initialization routine. We tested all values in the memory before, so we can skip the initialization.
The better way is to move sub-group barrier validation to a separate test.

Tagging original test authors @Melirius and @nolmoonen.

bader (approved):

tests/group_functions/group_barrier.cpp
nolmoonen: The PR author is correct regarding these changes, the barriers must be inserted to guarantee that the test's results match what is tested.

I think your understanding of the code is correct, see also the [group functions test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/group_functions.asciidoc#252-group_barrier) (though concise):
> when each of workitems writes 1 into zero-initialized local memory array (for `sub_group` and `work_group` memory scope) or zero-initialized global memory array (for `device` and `system` memory scope) after barrier all array values red by workitems in reverse order are equal to 1.

I agree that the confidence in the test would increase, if it is checked that the local memory is successfully set to `0`, before writing `1`.

I don't see how initialization for the sub-group part of the test can be skipped. If the group part of the test executed successfully, all values are set to `1`, they need to be `0` at the start of the sub-group part.

Nuullll: Thanks for input. I've added checks for initialization

Nuullll: > I don't see how initialization for the sub-group part of the test can be skipped. If the group part of the test executed successfully, all values are set to `1`, they need to be `0` at the start of the sub-group part.

I think @bader meant that we could test buffers with a different value (e.g. `2`) for sub-group part, so that we can skip initialization for sub-group part since we have checked that all values were set to `1` in the group part.

nolmoonen: Alright, in that case I agree that it can be skipped.

Nuullll:
gentle ping

tests/group_functions/group_barrier.cpp
```diff
@@ -149,11 +149,17 @@ TEMPLATE_TEST_CASE_SIG("Group barriers", "[group_func][dim]", ((int D), D), 1,
       local_acc[llid] = 0;
       sycl::group_barrier(group);
 
+      if (local_acc[max_id - llid] != 0)
+        std::get<s::test>(group_barriers_acc[0]) = false;
+      // make sure we check all items before moving on
+      sycl::group_barrier(group);
```
keryell, line 155: I have been puzzled a little bit by this one.
If the barrier does not work at the first place we will have UB since we will read some non initialized values. But anyway, if there is the barrier not working, everything is UB at the first place, so, not a big deal. ;-)
Furthermore, if it happens that the implementation initialize the local accessor to 0, this might be unnoticed.
So what about for example writing `53 + llid` instead of 0 on line 149 and update accordingly line 152?

keryell (approved):
I think this is good.
More generally, there is a problem in the existing test about just writing full 0 or 1 might hide some bugs. But I guess this is outside of this PR and hopefully the bug would show up in another context.

tests/group_functions/group_barrier.cpp
Nuullll: Done. Changed to `llid` for initializations.

PR #751, opened 2023-07-18
Enable dimension item/id/range dimension checks for DPC++

steffenlarsen:
This commit enables a number of test cases previously disabled for DPC++ due to missing `dimensions` member in `id`, `range`, `nd_range`, `h_item`, `item`, and `nd_item`.

### DIFF
--- a/tests/id/id.cpp
+++ b/tests/id/id.cpp
@@ -284,9 +284,7 @@ TEMPLATE_TEST_CASE_SIG("id supports get() and operator[]", "[id]", ((int D), D),
   }
 }
 
-// FIXME: re-enable when sycl::id<>::dimensions is implemented
-// Issue link https://github.com/intel/llvm/issues/9786
-DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(DPCPP, hipSYCL)
+DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(hipSYCL)
 ("id provides static constexpr member 'dimensions'", "[id]", ((int D), D), 1, 2,
  3)({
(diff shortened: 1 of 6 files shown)


bader (approved):

tests/id/id.cpp
```diff
@@ -284,16 +284,21 @@ TEMPLATE_TEST_CASE_SIG("id supports get() and operator[]", "[id]", ((int D), D),
...
+  KCHECK(EVAL((std::is_same_v<decltype(sycl::id<1>::dimensions), const int>)));
+  KCHECK(EVAL((std::is_same_v<decltype(sycl::id<2>::dimensions), const int>)));
+  KCHECK(EVAL((std::is_same_v<decltype(sycl::id<3>::dimensions), const int>)));
+
+  CHECK(sycl::id<1>::dimensions == 1);
+  CHECK(sycl::id<2>::dimensions == 2);
+  CHECK(sycl::id<3>::dimensions == 3);
+  KCHECK(EVAL(sycl::id<1>::dimensions) == 1);
+  KCHECK(EVAL(sycl::id<2>::dimensions) == 2);
+  KCHECK(EVAL(sycl::id<3>::dimensions) == 3);
```
steffenlarsen, line 301: Note: The names given to the kernel in each instantiation of this test case would previously not be unique, hence the split here.

keryell: This is interesting design information that should appear in a comment.

keryell (approved):

tests/id/id.cpp
steffenlarsen: Good point. I've added a comment about it.

PR #752, opened 2023-07-18
CI: Set ccache directory explicitly

psalz:
Follow up to #749, now everything should work!


bader (approved):
LGTM.
Should we also revert https://github.com/KhronosGroup/SYCL-CTS/pull/749?

psalz:
No, both patches are needed to make things work again! Unfortunately it looks like the DPC++ build timed out at 7 hours, I just restarted it. We'll have to get it through at least once to prime the cache :-/.

PR #753, opened 2023-07-19
Make accessor test cases templated on non-type arguments

steffenlarsen:
Currently the tests for accessors iterate through all combinations of valid arguments each time they try to run a specific section for a specific combination. Though the irrelevant combinations for a given section run no checks or do any significant work, they still introduce an overhead that accumulates, causing it to increase drastically when any of the argument spaces increase, e.g. as the type arguments do in full-conformance mode.

To decrease the overhead, this commit changes the accessor tests to use "templated" test cases which each get a combination of arguments and will not check for their relevant sections under other combinations. This should not affect the coverage of the tests, but should decrease the width of the search each section run has to do.

Note that this only currently templates the test cases on access mode, dimensions, and targets. Including the types in this to further focus testing runs is left for future work.

### DIFF
--- a/tests/accessor/generic_accessor_api_common.h
+++ b/tests/accessor/generic_accessor_api_common.h
@@ -438,13 +438,24 @@ class run_api_tests {
   }
 };
 
-template <typename T>
+using test_combinations =
+    typename get_combinations<access_modes_pack, all_dimensions_pack,
+                              targets_pack>::type;
+
+template <typename T, typename ArgCombination>
 class run_generic_api_for_type {
  public:
   void operator()(const std::string &type_name) {
-    const auto access_modes = get_access_modes();
-    const auto dimensions = get_all_dimensions();
-    const auto targets = get_targets();
+    // Get the packs from the test combination type.
+    using AccessModePack = typename std::tuple_element<0, ArgCombination>::type;
+    using DimensionsPack = typename std::tuple_element<1, ArgCombination>::type;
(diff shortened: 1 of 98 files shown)


tests/accessor/generic_accessor_api_common.h
```diff
@@ -438,13 +438,24 @@ class run_api_tests {
...
 class run_generic_api_for_type {
  public:
   void operator()(const std::string &type_name) {
-    const auto access_modes = get_access_modes();
-    const auto dimensions = get_all_dimensions();
-    const auto targets = get_targets();
+    // Get the packs from the test combination type.
+    using AccessModePack = typename std::tuple_element<0, ArgCombination>::type;
+    using DimensionsPack = typename std::tuple_element<1, ArgCombination>::type;
+    using TargetsPack = typename std::tuple_element<2, ArgCombination>::type;
```
keryell, line 452: While you are here, you could change everywhere
```suggestion
    using AccessModePack = typename std::tuple_element_t<0, ArgCombination>;
    using DimensionsPack = typename std::tuple_element_t<1, ArgCombination>;
    using TargetsPack = typename std::tuple_element_t<2, ArgCombination>;
```

keryell (approved):
This is massive.
Mostly, I trust you. :-)

tests/accessor/generic_accessor_api_common.h
steffenlarsen: Absolutely! It has been done. 👍

PR #756, opened 2023-07-21
Fix defects in vector `check_*` functions

0x12CC:
Change the return type of `check_as_result` to `bool` to match its definition.

Fix `check_convert_as_all_dims` and `check_convert_as_all_types` to combine `check_*` results using the `&=` operator rather than the `+=` operator. This change is consistent with the definition of `check_vector_convert_modes`.

Fix `check_vector_convert_result` to use the `rte` or `rtz` conversion functions when the `mode` is `rounding_mode::automatic`.

### DIFF
--- a/tests/common/common_vec.h
+++ b/tests/common/common_vec.h
@@ -433,8 +433,21 @@ bool check_vector_convert(sycl::vec<vecType, N> inputVec) {
 template <typename vecType, int N, typename asType, int asN>
 bool check_as_result(sycl::vec<vecType, N> inputVec,
                      sycl::vec<asType, asN> asVec) {
-  constexpr int n = std::min(N * sizeof(vecType), asN * sizeof(asVec));
-  return std::memcmp(&inputVec[0], &asVec[0], n) == 0;
+  vecType tmp_ptr[N];
+  for (size_t i = 0; i < N; ++i) {
+    tmp_ptr[i] = getElement(inputVec, i);
+  }
+  asType exp_ptr[asN];
+  for (size_t i = 0; i < asN; ++i) {
+    exp_ptr[i] = getElement(asVec, i);
+  }
+  std::memcpy(exp_ptr, tmp_ptr, std::min(sizeof(exp_ptr), sizeof(tmp_ptr)));
+  for (size_t i = 0; i < asN; ++i) {
+    if (exp_ptr[i] != getElement(asVec, i)) {
+      return false;
+    }
+  }
+  return true;
 }
 
 /**


tests/common/common_vec.h
```diff
@@ -426,23 +431,9 @@ bool check_vector_convert(sycl::vec<vecType, N> inputVec) {
...
-  std::memcpy(exp_ptr, tmp_ptr, std::min(sizeof(exp_ptr), sizeof(tmp_ptr)));
-  for (size_t i = 0; i < asN; ++i) {
-    if (exp_ptr[i] != getElement(asVec, i)) {
-      return false;
-    }
-  }
-  return true;
+bool check_as_result(sycl::vec<vecType, N> inputVec,
+                     sycl::vec<asType, asN> asVec) {
+  return std::memcmp(&inputVec, &asVec, N * sizeof(vecType)) == 0;
```
bader, line 436: > Fix check_as_result to prevent reading from uninitialized memory for cases where (N * sizeof(vecType)) < (asN * sizeof(asType)). Also, change its return type to bool to match its definition.

Would we still read uninitialized memory when `(N * sizeof(vecType)) > (asN * sizeof(asType))`?

bader: > This case should work as expected since the source type initializes all of the bytes before conversion.

I don't think this is guaranteed by the spec. The layout of sycl::vec type is not defined, so we can't use memcmp to compare two sycl::vec.

gmlueck: What was the problem with the original code?  The spec says that `vec::as` may only be used when the converted size is the same as the original size:

> Bitwise reinterprets this SYCL vec as a SYCL vec of a different element type and number of elements specified by asT. **The new SYCL vec type must have the same storage size in bytes as this SYCL vec.**

I do think it would be better to use the `vec::byte_size` member function, rather than `sizeof(vecType)` to determine the size of the data.

gmlueck: I think the spec does define the layout in 4.14.2.6:

> The elements of an instance of the SYCL vec class template are stored in memory sequentially and contiguously 

It seems like your updated code could still access uninitialized elements.  Don't you need something like this:

> `std::memcmp(&inputVec, &asVec, std::min(N * sizeof(vecType), asN * sizeof(asType)))`

gmlueck: > I didn't use `std::min` since the opposite conversion (such as from `sycl::vec<int, 4>` to `sycl::vec<float, 3>`) doesn't have uninitialized memory; The original 4-element vector initializes all 16 bytes.

However, the "as" vector only has 3 elements, so I don't think you can assume that the fourth element was copied to the "as" vector.
 
> I can make the change your suggesting, along with the switch to `vec::byte_size`, but I don't believe either will change the test's behavior.

Now that you explained the problem, I think you  cannot use `vec::byte_size`.  That function return the size of all 4 elements in the case when `N` is 3.

keryell: I can imagine different endianness for some architectures with the `sycl::vec<float, 3>` might be either at the end or the beginning of a `sycl::vec<float, 4>`.

gmlueck: Perhaps we should change the definition of `vec::as` to disallow this case, adding the following restriction:

> Bitwise reinterprets this SYCL `vec` as a SYCL `vec` of a different element type and number of elements specified by `asT`. The new SYCL `vec` type must have the same storage size in bytes as this SYCL `vec`, **and the size of the elements in the new `vec` (`NumElements * sizeof(DataT)`) must be the same as the size of the elements in this SYCL `vec`.**

bader: > I think the spec does define the layout in 4.14.2.6:
> 
> > The elements of an instance of the SYCL vec class template are stored in memory sequentially and contiguously

I interpret "The elements" to refer to the vec elements and not the whole vec object. I suppose this vec definition satisfies this requirement too:

```c++
template <typename T, int S>
class vec {
int padding; // e.g. to satisfy the alignment requirement
T data[S]; // 3-element vector is defined separately
}
```

BTW, 4.14.2.6 section says "This is true for both host and device code in order to allow for instances of the vec class template to be passed to SYCL kernel functions.", but I think to requiring the same layout on the host and the device is enough to make it "pass-able" to SYCL kernel functions (+device-copyable).

> 
> It seems like your updated code could still access uninitialized elements. Don't you need something like this:
> 
> > `std::memcmp(&inputVec, &asVec, std::min(N * sizeof(vecType), asN * sizeof(asType)))`

gmlueck: > I think that this would be the best option. The comparison in this PR should remain valid with the proposed change.

If we change the spec as I propose above, then the test case where you convert `sycl::vec<float, 3>` to `sycl::vec<int, 4>` would no longer be valid.  Therefore, the original checking code in `check_as_result` would work in all cases, and it's better than the change you propose because it works even in the bizarre case where element 0 is not at the beginning of the `vec` (as @bader pointed out).

gmlueck: I still think the original code was better.  The spec says that `operator[]` returns a reference to the indexed element.  It does not say that the returned reference can be used as an iterator to the subsequent elements in the vec.  I agree this is pedantic, but I don't see why this is an improvement over the original code.  By storing the elements in an array, we definitely know the memory layout.

BTW: I submitted KhronosGroup/SYCL-Docs#447 to clarify `vec::as` as we discussed above.

bader (approved):

gmlueck (approved):

fraggamuffin:
another reviewer please

ProGTX (approved):

PR #757, opened 2023-07-25
Fix bool check on generic accessor api

jiezzhang:
Replace the first AND assignment with assignment because `res` is initialized as `false`.


bader (approved):

keryell (approved):
Thanks.

maarquitos14:
`res_acc` is being changed in `test_accessor_ptr_device`, so this change is dismissing the results from that function. I think this PR should be reverted.

bader:
> `res_acc` is being changed in `test_accessor_ptr_device`, so this change is dismissing the results from that function. I think this PR should be reverted.

Good eye. I missed the use of `res_acc` by `test_accessor_ptr_device`.

PR #758, opened 2023-07-25
Fix expected return type for get_pointer().

maarquitos14:
According to recent changes in SYCL2020 spec (https://github.com/KhronosGroup/SYCL-Docs/pull/432), update expected return types for `get_pointer()` in different cases.


maarquitos14:
> @maarquitos14, could you resolve merge conflicts, please?

I have left a comment in #757, which is the source of the conflict, because I think it should be reverted. Otherwise, I can just resolve the conflict here, but I think it would be clearer if we first revert #757 and then merge this.

PR #762, opened 2023-07-27
Correctly use fixed-size integers in vec builtins

steffenlarsen:
The SYCL specification uses fixed-size integers in builtins taking and/or returning integer vectors. However, the current version of the CTS generates tests using fundamental integer types. This commit changes the definitions of these builtins to instead use the appropriate fixed-size integer types.

### DIFF
--- a/tests/math_builtin_api/modules/sycl_types.py
+++ b/tests/math_builtin_api/modules/sycl_types.py
@@ -143,20 +143,20 @@ def create_basic_types():
     t_char_0 = argtype("char", "scalar", "char", 1, [])
     type_dic["char"] = t_char_0
 
-    t_vchar_2 = argtype("sycl::vec<char, 2>", "vector", "char", 2, [])
-    type_dic["sycl::vec<char, 2>"] = t_vchar_2
+    t_vint8_2 = argtype("sycl::vec<int8_t, 2>", "vector", "int8_t", 2, [])
+    type_dic["sycl::vec<int8_t, 2>"] = t_vint8_2
 
-    t_vchar_3 = argtype("sycl::vec<char, 3>", "vector", "char", 3, [])
-    type_dic["sycl::vec<char, 3>"] = t_vchar_3
+    t_vint8_3 = argtype("sycl::vec<int8_t, 3>", "vector", "int8_t", 3, [])
+    type_dic["sycl::vec<int8_t, 3>"] = t_vint8_3
 
-    t_vchar_4 = argtype("sycl::vec<char, 4>", "vector", "char", 4, [])
-    type_dic["sycl::vec<char, 4>"] = t_vchar_4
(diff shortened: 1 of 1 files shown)


tests/math_builtin_api/modules/sycl_types.py
```diff
@@ -1226,33 +1196,29 @@ def create_types():
 
 
     # t_char_n = argtype("charn", "NULL", "NULL", 0, ["sycl::char2","sycl::char3","sycl::char4","sycl::char8","sycl::char16"])
```
bader, line 1198: This comment is outdated now. I suggest removing such comments.

bader (approved):

tests/math_builtin_api/modules/sycl_types.py
steffenlarsen: Arguably the aliases are still valid given the aliases map to [fixed-size integer vectors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_aliases) despite using the names of fundamental types. That said, I don't know why the comments were there in the first place, given they seem to be valid definitions that have been commented out. That considered, would you still like me to remove them?

bader: Do you mean that this file is used to generate variants for something else other than builtins?

bader: I'm asking this because technically, your patch is removing "arguably valid aliases".

steffenlarsen: I don't believe so, no. This should all just be for builtins but it seems like we elected to write out the types of the aliases instead of using the aliases defined in the spec. At least that is my guess as to why these were commented out. Note though that these aliases changed from 1.2.1 to 2020 to be more in line with how the corresponding vector types in OpenCL are fixed-size.

gmlueck: Not sure if this is important, but we plan to deprecate these "vec" aliases.  See the comments at the bottom of KhronosGroup/SYCL-Docs#353.

bader: Okay. @steffenlarsen, up to you. I'm waiting for another approval. I'll merge with the comment if you decide to leave it.

steffenlarsen: I think I agree with removing them, but I would prefer to do it in a separate patch as there are also non-integer aliases that this likely affects. I will open a separate patch for removing the comments.

steffenlarsen: Correction. It seems to only be the integer ones, so I will include it in this patch.

ProGTX (approved):

PR #763, opened 2023-07-27
Revert "Fix bool check on generic accessor api"

bader:
Reverts KhronosGroup/SYCL-CTS#757

From https://github.com/KhronosGroup/SYCL-CTS/pull/757#issuecomment-1653958589.

> `res_acc` is being changed in `test_accessor_ptr_device`, so this change is dismissing the results from that function.


bader:
@jiezzhang, FYI.

PR #767, opened 2023-07-28
Remove ComputeCPP support

bader:


psalz (approved):
Pretty crazy to see how many workarounds we had in place! LGTM

PR #771, opened 2023-08-01
Add SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS CMake option

steffenlarsen:
This commit adds the SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS as a shortcut for enabling all oneapi extension tests. It will only overwrite the other options if turned ON. In the OFF state the option has no effect.


bader:
Can we make `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` to set a default value for the tests, which can be overridden by test specific macro?

Let's consider this set of options: SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS=ON SYCL_CTS_ENABLE_EXT_ONEAPI_MEMCPY2D_TESTS=OFF. With the current patch the second option is ignored.

keryell:
Perhaps the semantics could be that `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` has an effect only on other options that have not explicitly set either `ON` or `OFF`.
But the problem with boolean options in CMake is that they are considered as `OFF`. https://cmake.org/cmake/help/latest/command/option.html
A solution could be to have these options as strings, so now we can have tri-state  `ON`, `OFF` and nothing.
That said, is this something to pursue?

psalz:
> Perhaps the semantics could be that `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` has an effect only on other options that have not explicitly set either `ON` or `OFF`. But the problem with boolean options in CMake is that they are considered as `OFF`. https://cmake.org/cmake/help/latest/command/option.html A solution could be to have these options as strings, so now we can have tri-state `ON`, `OFF` and nothing. That said, is this something to pursue?

Couldn't this be achieved by simply passing `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` as the default value for each of the other options? I.e.,

```CMake
add_cts_option(SYCL_CTS_ENABLE_EXT_ONEAPI_SUB_GROUP_MASK_TESTS
    "Enable extension oneAPI sub_group_mask tests" ${SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS})
```

steffenlarsen:
> Couldn't this be achieved by simply passing `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` as the default value for each of the other options? I.e.,
> 
> ```cmake
> add_cts_option(SYCL_CTS_ENABLE_EXT_ONEAPI_SUB_GROUP_MASK_TESTS
>     "Enable extension oneAPI sub_group_mask tests" ${SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS})
> ```

That was my initial thought too, but the downside to it is that the other options won't change if you set `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` after the initial configuration.

The use-case in mind for this option was to allow for enabling all extension tests in the given group (here `oneapi` but the same strategy could be used wider) so automated systems won't have to be changed every time a new extension gets tests in the suite. As such a simple overwrite is sufficient, but if it is possible to make it work with selective disabling in tandem with this option, it shouldn't break the use-case.

keryell:
@MathiasMagnus any feedback on this?

cmake/AddCTSOption.cmake
```diff
@@ -11,10 +11,12 @@ set(SYCL_CTS_DETAIL_OPTION_COMPILE_DEFINITIONS "")
 #            Use `#if <OPTION>` instead of `#ifdef <OPTION>`.
 #
 # Optional parameters:
-#  - WARN_IF_OFF <msg>     Print a warning message if this option is set to OFF.
+#  - WARN_IF_OFF <msg>    Print a warning message if this option is set to OFF.
+#  - FORCE_ON <value>     If <value> is ON the option will be overwritten as ON.
 #
 function(add_cts_option option_name option_description option_default)
     cmake_parse_arguments(PARSE_ARGV 3 args "" "WARN_IF_OFF" "")
+    cmake_parse_arguments(PARSE_ARGV 3 args "" "FORCE_ON" "")
```
MathiasMagnus, line 19: No need to parse args twice. Add `FORCE_ON` to the previous invocation. (`"WARN_IF_OFF;FORCE_ON"`) if we don't feed the option, one-value, multi-value args from variables.

MathiasMagnus (changes requested):
Mostly good.

psalz:
> That was my initial thought too, but the downside to it is that the other options won't change if you set `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` after the initial configuration.

Fair, but couldn't the same argument be made in reverse then? I.e., once `SYCL_CTS_ENABLE_EXT_ONEAPI_TESTS` is set, it is no longer possible to selectively disable a subset of the extension tests?

> The use-case in mind for this option was to allow for enabling all extension tests in the given group (here `oneapi` but the same strategy could be used wider) so automated systems won't have to be changed every time a new extension gets tests in the suite.

Yes that makes sense. An alternative solution to this might be to ship a set of [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).

fraggamuffin:
needs another week

keryell (approved):
The solution sounds good for the envisioned use case.

PR #766, opened 2023-07-28
Fix marray comparison in group function tests

steffenlarsen:
This commit fixes equality comparison of marray in group function tests.

PR #773, opened 2023-08-02
Handle `marray` in `equal` helper for group function tests

jchlanda:
Before a recent fix to `marray` implementation
(https://github.com/intel/llvm/pull/10596) it was possible to convert `marray<T, N>` to `T` (incorrect behaviour), after the fix, we require a helper to traverse all elements of the array to check for equality.

### DIFF
--- a/tests/group_functions/group_functions_common.h
+++ b/tests/group_functions/group_functions_common.h
@@ -42,6 +42,20 @@ bool equal_impl(const T& a, const U& b) {
 }
 
 template <typename T, int N>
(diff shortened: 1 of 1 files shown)


bader:
@jchlanda, I think these helpers are already added by @steffenlarsen in https://github.com/KhronosGroup/SYCL-CTS/pull/766.

jchlanda:
My bad, sorry. I must have been behind the tip when testing this and did not notice the change after pulling. Closing now.

PR #776, opened 2023-08-02
Prevent signed integer overflow in `check_scan_over_group`

0x12CC:
Check that the maximum value of `T` is large enough to store the results of the scan functions. This check prevents overflow for small types. Also, remove the largest `check_scan` test case.

### DIFF
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -238,7 +238,7 @@ struct init_joint_scan_group {
 
       size_t work_group_size = work_group_range.size();
 
-      const size_t sizes[3] = {5, work_group_size / 2, 3 * work_group_size};
+      const size_t sizes[3] = {5, work_group_size / 2};
       for (size_t size : sizes) {
         check_scan<D, T, U, sycl::group<D>, true, I>(
             queue, size, executionRange, OperatorT(), op_name);
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -303,6 +303,10 @@ void check_scan_over_group(sycl::queue& queue, sycl::range<D> range, OpT op,
   bool ret_type_e = false;
   bool ret_type_i = false;
 
+  if (std::sqrt(std::numeric_limits<T>::max()) + T(init) <= range.size()) {
+    return;
+  }
+
   std::vector<size_t> local_id(range_size, 0);
 
   sycl::nd_range<D> executionRange(range, range);
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -180,7 +180,7 @@ struct joint_scan_group {
 
       sycl::nd_range<D> executionRange(work_group_range, work_group_range);
 
-      const size_t sizes[3] = {5, work_group_size / 2, 3 * work_group_size};
+      const size_t sizes[2] = {5, work_group_size};
       for (size_t size : sizes) {
         check_scan<D, T, U, sycl::group<D>, false>(queue, size, executionRange,
                                                    OperatorT(), op_name);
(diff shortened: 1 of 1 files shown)


tests/group_functions/group_scan.h
```diff
@@ -303,6 +303,10 @@ void check_scan_over_group(sycl::queue& queue, sycl::range<D> range, OpT op,
   bool ret_type_e = false;
   bool ret_type_i = false;
 
+  if (std::sqrt(std::numeric_limits<T>::max()) + T(init) <= range.size()) {
+    return;
+  }
+
```
bader, line 309: I don't think it's a good idea silently skipping all checks in this case.
Although this sounds like a test issue, rather than a bug in the tested implementation, we probably can use REQUIRE macro to "Check that the maximum value of T is large enough to store the results of the scan functions.".

bader: Why do we launch the configuration causing the overflow? Isn't this a bug?

bader: Right. This where I think we should change the logic to avoid an overflow. 
work_goup_size here is the maximum possible # of work-items the device can execute. Do we really need to the maximum # of work-items to test the scan functionality? If so, the input buffer must be initialized with zeros to avoid overflow.
But I don't think we need to run so many work-items, it's a good stress test, but an overkill for this test purpose.

https://github.com/KhronosGroup/SYCL-CTS/blob/47a05179ba09a364c4db03ce07fc387564eebdac/tests/group_functions/group_scan.h#L68 - this seems to be the right place to make sure that we can run the test w/o an overflow. BTW, it's better to use [the sum of arithmetic progression](https://en.wikipedia.org/wiki/Arithmetic_progression#Sum) instead of sqrt for the overflow check.

tests/group_functions/group_scan.h
```diff
@@ -238,7 +238,7 @@ struct init_joint_scan_group {
 
       size_t work_group_size = work_group_range.size();
 
-      const size_t sizes[3] = {5, work_group_size / 2, 3 * work_group_size};
+      const size_t sizes[3] = {5, work_group_size / 2};
```
bader, line 241: Either declare 2-element array or explicitly initialize the 3rd element.

Why the second element is `work_group_size / 2`? Why can't we put 1 or 2 instead (+ check that HW is able to run the work of that size)?

tests/group_functions/group_scan.h
```diff
@@ -303,6 +303,11 @@ void check_scan_over_group(sycl::queue& queue, sycl::range<D> range, OpT op,
   bool ret_type_e = false;
   bool ret_type_i = false;
 
+  if (((range_size * (range_size + 1) / 2) + T(init)) >
+      std::numeric_limits<T>::max()) {
+    return;
+  }
+
```
bader, line 310: I still think that silently skipping checks is a bad idea. Please, replace with an assertion.

AerialMantis (approved):

fraggamuffin:
Ivan, please have a look. Thx.

bader:
I'll wait for the feedback from @Melirius to merge.

keryell (approved):
Thanks.

fraggamuffin:
Will wait one more week.

bader:
Friendly reminder for @Melirius.

bader:
Friendly reminder №2 for @Melirius.

@0x12CC, could you resolve merge conflicts with SYCL-2020 branch, please?

tests/group_functions/group_scan.h
keryell: That means which should even more comment the code.

Melirius:
> Friendly reminder №2 for @Melirius.

Sorry, I'm at vacations now, be back on the next week.

bader:
> > Friendly reminder №2 for @Melirius.
> 
> Sorry, I'm at vacations now, be back on the next week.

@Melirius, friendly reminder №3.

We have required approvals, so it's ready to be merged. I'm going to merge this PR next Wednesday if there are no additional requests for changes.
@Melirius, feel free provide your feedback regardless of the merge status. I'll make sure that we address post-merge review comments too.

PR #777, opened 2023-08-03
Fix kernel_bundle bugs

KornevNikita:
1. Device images may already be filtered according to the device before the Selector call, so there is no gurantee that the Selector will meet all kernels. Need to check the kernel for compatibility to see if it was already filtered out.
2. Also need to check for compatibility before calling get_kernel_bundle for a certain kernel. It may lead to an exception if the kernel is not compatible.

### DIFF
--- a/tests/kernel_bundle/get_kernel_bundle.h
+++ b/tests/kernel_bundle/get_kernel_bundle.h
@@ -94,6 +94,7 @@ struct run_tests_for_overloads_that_obtain_kernel_name {
         kernel_bundle::get_restrictions<KernelDescriptorT,
                                         sycl::bundle_state::executable>()};
     using kernel = typename KernelDescriptorT::type;
+    if (!sycl::is_compatible<kernel>(device)) return;
     auto k_id{sycl::get_kernel_id<kernel>()};
 
     // verifications for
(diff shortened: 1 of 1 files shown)


tests/kernel_bundle/get_kernel_bundle.h
```diff
@@ -94,6 +94,7 @@ struct run_tests_for_overloads_that_obtain_kernel_name {
         kernel_bundle::get_restrictions<KernelDescriptorT,
                                         sycl::bundle_state::executable>()};
     using kernel = typename KernelDescriptorT::type;
+    if (!sycl::is_compatible<kernel>(device)) return;
```
keryell, line 97: Add a comment from what you say in tour PR description.

tests/kernel_bundle/get_kernel_bundle.h
```diff
@@ -131,13 +132,14 @@ struct helper_device_image_has_kernel {
...
-                  unsigned int index) {
+  void operator()(const sycl::device_image<BundleState>& device_image,
+                  std::bitset<KernelDescriptorCount>& has_kernel,
+                  unsigned int index, const sycl::device& device) {
     using kernel = typename KernelDescriptorT::type;
     auto kernel_id{sycl::get_kernel_id<kernel>()};
 
-    if (device_image.has_kernel(kernel_id)) {
+    if (device_image.has_kernel(kernel_id) ||
+        !sycl::is_compatible({kernel_id}, device)) {
```
keryell, line 142: Add a comment from what you say in tour PR description.

tests/kernel_bundle/get_kernel_bundle.h
```diff
@@ -155,7 +157,8 @@ struct helper_device_image_has_kernel_dev {
     using kernel = typename KernelDescriptorT::type;
     auto kernel_id{sycl::get_kernel_id<kernel>()};
 
-    if (device_image.has_kernel(kernel_id, device)) {
+    if (device_image.has_kernel(kernel_id, device) ||
+        !sycl::is_compatible({kernel_id}, device)) {
```
keryell, line 161: Add a comment from what you say in tour PR description.

KornevNikita:
@keryell all done with [09f2533](https://github.com/KhronosGroup/SYCL-CTS/pull/777/commits/09f25330c188d2b3ed913a9efb67d99b2064f173)

ProGTX (approved):

keryell (approved):
Thanks for the clarifications!

PR #779, opened 2023-08-04
Fix sub_group test

KornevNikita:
Passing sub_group instances captured from outside is UB.

### DIFF
--- a/tests/sub_group/sub_group_semantics.cpp
+++ b/tests/sub_group/sub_group_semantics.cpp
@@ -122,17 +122,31 @@ TEST_CASE("Check sycl::sub_group equality", "[sub_group]") {
       to_integral(common_by_value_semantics::current_check::size);
   bool result[code_count];
   std::fill(result, result + code_count, false);
-  auto items = store_instances<2, invoke_sub_group<3, setup_kernel>>();
   {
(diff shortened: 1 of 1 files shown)


keryell:
I am unsure the problem is what you describe, since technically in a single task there is only 1 work-item, so there cannot be anything in-between.
But I agree that using subgroups captured from outside is kind of surprising.

bader:
> I am unsure the problem is what you describe, since technically in a single task there is only 1 work-item, so there cannot be anything in-between.

Great, so you agree with changing `single_task` to `parallel_for`. Right?

> But I agree that using subgroups captured from outside is kind of surprising.

My guess is the description refers the deleted line 125. After switching to `parallel_for` we have more than 1 work-item. 

@keryell, what would you like to improve in this PR to approve the merge? Just description?

KornevNikita:
@keryell the problem is here:
`But I agree that using subgroups captured from outside is kind of surprising.`
Probably my commit message is a bit confusing. Yes, I mean taking the sub_group instance from one command_group and then passing it to another is wrong - so it leads to wrong test results.

ProGTX (approved):

keryell (approved):
Thanks!

PR #783, opened 2023-08-15
Fix sub_group_mask_extract_bits expectation and comparison

steffenlarsen:
This commit fixes an issue where the sub_group_mask_extract_bits would not use a valid comparison operation for marray argument types. This also revealed a faulty expectation in the test for marray as only the first element would previously have been checked. This faulty expectation is also fixed by this patch.

### DIFF
--- a/tests/extension/oneapi_sub_group_mask/sub_group_mask_extract_bits.cpp
+++ b/tests/extension/oneapi_sub_group_mask/sub_group_mask_extract_bits.cpp
@@ -49,6 +49,14 @@ void get_expected_bits(T &out, uint32_t mask_size, int pos) {
   }
 }
 
+// For marray the extracted bits start from the first element and moves position
+// into the consecutive elements.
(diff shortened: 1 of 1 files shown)


tests/extension/oneapi_sub_group_mask/sub_group_mask_extract_bits.cpp
```diff
@@ -49,16 +49,24 @@ void get_expected_bits(T &out, uint32_t mask_size, int pos) {
   }
 }
 
+// For marray the extracted bits start from the first element and moves position
+// into the consecutive elements.
+template <typename T, size_t N>
+void get_expected_bits(sycl::marray<T, N>& out, uint32_t mask_size, int pos) {
```
keryell, line 55: Should this be a common function in some header?

keryell (approved):
Thanks.

fraggamuffin:
LGTM

PR #784, opened 2023-08-16
Fix oneapi_bfloat16 compfail

dm-vodopyanov:
This patch fixes the following compfail in oneapi_bfloat16 test:

```
error: no matching function for call to 'bit_cast'
  bfloat16 bf_min = sycl::bit_cast<bfloat16>(0b0000000010000000);
                    ^~~~~~~~~~~~~~~~~~~~~~~~
```


keryell (approved):
Good point!

PR #789, opened 2023-08-23
[CI] Remove the list of excluded tests for ComputeCPP

bader:
This is a leftover after the ComputeCPP support removal.


keryell (approved):
Good catch.

PR #781, opened 2023-08-09
Fix host_accessor_api_core test

KornevNikita:
Reduce the numbers, so linear_index doesn't exceed CHAR_MAX. Otherwise the test fails here with the char type:

CHECK(value_operations::are_equal(acc_ref1, linear_index));

As acc_ref1 contains corrupted by the overflow value.

PR #790, opened 2023-08-24
Fix generic_accessor_api_core test

KornevNikita:
Same as: https://github.com/KhronosGroup/SYCL-CTS/pull/781

### DIFF
--- a/tests/accessor/generic_accessor_api_common.h
+++ b/tests/accessor/generic_accessor_api_common.h
@@ -310,18 +310,24 @@ class run_api_tests {
       SECTION(
           get_section_name<dims>(type_name, access_mode_name, target_name,
                                  "Check api for ranged accessor with offset")) {
-        constexpr size_t acc_range_size = 4;
-        constexpr size_t buff_range_size = 8;
-        constexpr size_t buff_size = (dims == 3)   ? 8 * 8 * 8
-                                     : (dims == 2) ? 8 * 8
-                                                   : 8;
-        constexpr size_t offset = 4;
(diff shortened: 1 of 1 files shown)


bader:
> Same as: #781

Please, copy the description to make clear commit message.

Can we move this code to a function to avoid code (and bug fixes) duplication?

KornevNikita:
> > > Same as: #781
> > 
> > 
> > Please, copy the description to make clear commit message.
> > Can we move this code to a function to avoid code (and bug fixes) duplication?
> 
> Yes, probably we can outline this block with numbers to a separate function accepting a lambda with the rest of code. And then call this function here and [there](https://github.com/KornevNikita/SYCL-CTS/blob/d5d98dd089220e9d5473823da6365c1de6e5a69f/tests/accessor/host_accessor_api_common.h#L119)

Ok, this change looks too big and difficult. Can we leave it as is? Because the test still failing with dpcpp and I'm not sure it it's test issue or implementation, so I'm trying to fix the second bug.

bader:
I'm okay with making the refactoring separately after fixing the test.
Meanwhile, additional comment that this code duplicates similar implementation in `tests/accessor/host_accessor_api_common.h` might be useful.

bader:
@ProGTX, @keryell, this fix is similar to the #781, which you previously approved. This should be easy to review.

ProGTX (approved):

keryell (approved):
Thanks.

KornevNikita:
@bader please merge

PR #791, opened 2023-08-25
Fix results validation in group_scan tests

AlexeySachkov:
Two issues were fixed:
- incorrect (static, not dynamic) index used in `joint_*` functions validation
- incorrectly computed indicies into results buffer, which caused validation only of 3/4th of the results

### DIFF
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -81,23 +81,27 @@ struct JointScanDataStruct {
                         init_value, op);
     std::inclusive_scan(ref_input.begin(), ref_input.end(), reference_i.begin(),
                         op, init_value);
+    // res consists of 4 series of results: two pairs of exclusive and inclusive
+    // scan results made over 'group' and 'sub_group' accordingly.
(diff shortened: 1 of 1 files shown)


tests/group_functions/group_scan.h
```diff
@@ -330,9 +334,12 @@ struct ScanOverGroupDataStruct {
     CHECK(ret_type[3]);
 
     T init_value = with_init ? T(init) : sycl::known_identity<OpT, T>::value;
+    // res consists of 4 series of results: two pairs of exclusive and inclusive
+    // scan results made over 'group' and 'sub_group' accordingly.
     for (int group_i = 0; group_i < 2; group_i++) {
       std::string group_name = group_i == 0 ? "group" : "sub_group";
-      size_t group_offset = range_size * group_i;
+      // Each group contains two sets of results.
+      size_t group_offset = 2 * range_size * group_i;
```
AlexeySachkov, line 342: Note: more changes are required for this test, because `local_id` has size `2 * range_size` and there will be out-of-bounds access on line 334

ProGTX (approved):

PR #793, opened 2023-08-31
Use correct types in JointScanDataStruct and fix out of range error.

maarquitos14:
Prior to this PR, we were misusing types, having `InPtr` where we should be using `OutPtr` and the other way around, causing multiple validation issues and even segmentation faults. Also, this PR fixes an out of range access.

### DIFF
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -57,8 +57,8 @@ auto joint_exclusive_scan_helper(Group group, T* v_begin, T* v_end,
 template <typename T, typename U>
 struct JointScanDataStruct {
   JointScanDataStruct(size_t range_size)
-      : ref_input(range_size), res(range_size * 4, T(-1)) {
-    std::iota(ref_input.begin(), ref_input.end(), U(1));
+      : ref_input(range_size), res(range_size * 4, U(-1)) {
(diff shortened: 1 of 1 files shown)


frasercrmck:
I have no authority to approve but I can confirm that this fixes the joint scan tests for me - thank you! I was previously seeing incorrect results and memory corruptions when running with DPC++ and the oneAPI construction kit.

It seems like hipSYCL skips the mixed-type cases. Is this bug the real reason behind that? Might those tests now work with hipSYCL?

bader:
> It seems like hipSYCL skips the mixed-type cases. Is this bug the real reason behind that? Might those tests now work with hipSYCL?

Tagging @illuhad in case he has something to share.

keryell (commented):
Good catch!
Interesting example of bad naming (`U`, `V`) preventing some mental double checking.
Perhaps you could suggest a better naming now you have suffered on this? :-)

keryell (approved):

maarquitos14:
> Good catch! Interesting example of bad naming (`U`, `V`) preventing some mental double checking. Perhaps you could suggest a better naming now you have suffered on this? :-)

I think we could follow the spec and use `InPtr` and `OutPtr` here.

PR #795, opened 2023-09-06
Fix incorrect all_of validation when sub-group size is 1

frasercrmck:
When the sub-group size is 1, the 'one_true' and 'some_true' predicates return true, so we should instead be checking that `all_of` returns true.

### DIFF
--- a/tests/group_functions/group_of.h
+++ b/tests/group_functions/group_of.h
@@ -197,6 +197,11 @@ void predicate_function_of_group(sycl::queue& queue) {
 
   sycl::range<D> work_group_range = sycl_cts::util::work_group_range<D>(queue);
   size_t work_group_size = work_group_range.size();
+  // Note that the predicates 'one_true' and 'some_true' are in fact 'all_true'
+  // for a work-group size of 1, but we don't consider that an interesting case
+  // to test: assert to make this apparent to anything modifying the test case.
+  assert(work_group_size != 1 &&
+         "Not all test checks hold when the work-group size is 1");
 
   // array to return results: 4 predicates * 3 functions
   bool res[test_matrix * test_cases] = {false};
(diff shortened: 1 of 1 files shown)


frasercrmck:
This problem technically lso applies to the corresponding work-group tests, but I'm assuming we'll never run those tests with a work-group size of 1.

bader (approved):
Thanks for fixing this!

> I'm assuming we'll never run those tests with a work-group size of 1.

Probably it's worth adding an assertion for that + a comment which of existing checks not work for a work-group size of 1.

keryell (approved):
Good catch!
Could you add a comment in front of all your changes explaining the problem?

frasercrmck:
Thanks for comments, will address later in the day.

I have another question, though. I was looking at this again with your feedback in mind, but is it valid for the `predicate_function_of_sub_group` test to be using `get_global_linear_id` as it does? Should we not be using the sub-group local linear ID ?

I ask because I thought it was invalid to make assumptions about the mapping of work-items to sub-group items. The `predicate_function_of_sub_group` test seems to rely on the first sub-group with ID 0 having all of the work-item IDs from `0` to `(sub-group-size)-1`. The `bool_function_of_sub_group` test doesn't do this, as it runs on all work-items of all sub-groups and uses the sub-group linear ID.

But also (noticed while writing this comment) is the `bool_function_of_sub_group` not overwriting its results for every sub-group? It seems to run over all sub-groups but only write to a fixed set of indices. Does that mean that some sub-groups might report incorrect results but get ignored?

frasercrmck:
I've updated with the comments requested.

frasercrmck:
> @frasercrmck do you plan to dive into your discoveries in your last comment?

If you agree they're real issues, yeah I can certainly take a look. I would prefer to keep that to a separate task/PR though if that's alright; I might need some guidance and a bit more back-and-forth as I'm not actually very that used to writing SYCL. Is that okay?

keryell:
> > @frasercrmck do you plan to dive into your discoveries in your last comment?
> 
> If you agree they're real issues, yeah I can certainly take a look. I would prefer to keep that to a separate task/PR though if that's alright; I might need some guidance and a bit more back-and-forth as I'm not actually very that used to writing SYCL. Is that okay?

sub-groups are very confusing to me. After looking at the SYCL specification, I have the feeling that the problem you expose is real. 
`T local_var(item.get_global_linear_id() + 1);` should be `T local_var =  sub_group.get_local_linear_id()`
and so on.

PR #797, opened 2023-09-12
Fix potential segfault in kernel_info test

frasercrmck:
If the device doesn't report any valid sub-group sizes, it's better to halt the test than to segfault when dereferencing `std::max_element`.

### DIFF
--- a/tests/kernel/kernel_info.cpp
+++ b/tests/kernel/kernel_info.cpp
@@ -203,6 +203,10 @@ TEST_CASE("Test kernel info", "[kernel]") {
                               "device_specific::max_sub_group_size>(dev)");
   {
     auto sub_group_sizes = dev.get_info<sycl::info::device::sub_group_sizes>();
+    {
+      INFO("The device must report at least one sub-group size");
+      REQUIRE(!sub_group_sizes.empty());
+    }
     uint32_t max =
         *std::max_element(sub_group_sizes.begin(), sub_group_sizes.end());
     INFO(


frasercrmck:
I'm taking it as a given that the device must report at least one sub-group size, though I couldn't find this mentioned in the SYCL spec.

We just spotted this because the oneAPI construction kit doesn't currently advertise any sub-group sizes. We support sub-groups but via vectorization, so it's fairly flexible which sizes we support and currently the compiler chooses the best one.

This actually causes us a problem in another test, `sub_group_api`, where the compiler generates a vectorized sub-group but we don't report it via `sycl::info::device::sub_group_sizes`. I bring this up in case anyone has any thoughts - is it really valid for the SYCL CTS to require that _all_ sub-group sizes reported by a kernel are in turn reported via `sycl::info::device::sub_group_sizes`? I can see the argument either way, but it's definitely awkward for a SYCL target that doesn't have a fixed hardware-like notion of a sub-group.

bader (approved):

keryell:
> I'm taking it as a given that the device must report at least one sub-group size, though I couldn't find this mentioned in the SYCL spec.

That is a good point to discuss in the committee.

> We just spotted this because the oneAPI construction kit doesn't currently advertise any sub-group sizes. We support sub-groups but via vectorization, so it's fairly flexible which sizes we support and currently the compiler chooses the best one.

Interesting.

> This actually causes us a problem in another test, `sub_group_api`, where the compiler generates a vectorized sub-group but we don't report it via `sycl::info::device::sub_group_sizes`. I bring this up in case anyone has any thoughts - is it really valid for the SYCL CTS to require that _all_ sub-group sizes reported by a kernel are in turn reported via `sycl::info::device::sub_group_sizes`? I can see the argument either way, but it's definitely awkward for a SYCL target that doesn't have a fixed hardware-like notion of a sub-group.

An easy solution could be to consider no sub-group support as a trivial sub-group with 1 element.

But it does not solve the case for a program wanting to force in your case above some vectorization for some non-advertized "sub-group".

tomdeakin:
Ready to merge test - open issue against spec.

PR #573, opened 2023-03-24
Create test plan for root_group extension

kopylovanat:
Create test plan for extension [sycl_ext_oneapi_root_group](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_root_group.asciidoc).

PR #799, opened 2023-09-12
Implement tests for root_group extension

0x12CC:
Implement the tests from the test plan in https://github.com/KhronosGroup/SYCL-CTS/pull/573.

### DIFF
--- a/tests/extension/oneapi_root_group/root_group.cpp
+++ b/tests/extension/oneapi_root_group/root_group.cpp
@@ -1,0 +1,256 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2023 The Khronos Group Inc.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+*******************************************************************************/
+
+#include "../../common/common.h"
+
+namespace root_group::tests {
+
+#ifdef SYCL_EXT_ONEAPI_ROOT_GROUP
+
+template <typename KernelName, size_t... Dims>
+static void check_root_group_api() {
+  constexpr int Dimensions = sizeof...(Dims);
+  auto q = sycl_cts::util::get_cts_object::queue();
+  auto bundle =
+      sycl::get_kernel_bundle<sycl::bundle_state::executable>(q.get_context());
+  auto kernel = bundle.get_kernel<KernelName>();
+  auto maxWGs = kernel.template ext_oneapi_get_info<
+      sycl::ext::oneapi::experimental::info::kernel_queue_specific::
+          max_num_work_group_sync>(q);
+  REQUIRE(maxWGs >= 1);
+  auto local_range = sycl::range<Dimensions>(Dims...);
+  auto global_range = local_range;
+  global_range[0] *= maxWGs;
+  REQUIRE(global_range.size() == local_range.size() * maxWGs);
+  auto nd_range = sycl::nd_range<Dimensions>{global_range, local_range};
+  auto props = sycl::ext::oneapi::experimental::properties{
+      sycl::ext::oneapi::experimental::use_root_sync};
+
+  struct checks {
+    enum {
+      get_group_id,
+      get_local_id,
+      get_group_range,
+      get_local_range,
+      get_max_local_range,
+      get_group_linear_id,
+      get_local_linear_id,
+      get_group_linear_range,
+      get_local_linear_range,
+      leader,
+      get_child_group___root_group,
+      get_child_group___group,
+      COUNT,
+    };
+  };
+
+  using CheckResults = bool[checks::COUNT];
+  sycl::buffer<CheckResults, Dimensions> results_buffer{global_range};
+
+  q.submit([&](sycl::handler& cgh) {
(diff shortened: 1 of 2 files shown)


tests/extension/oneapi_root_group/root_group.cpp
```diff
@@ -0,0 +1,256 @@
...
+TEST_CASE(
+    "Test for \"Root Group\" extension, check root_group class and "
+    "get_child_group API functionality",
+    "[oneapi_root_group_api]") {
+#ifndef SYCL_EXT_ONEAPI_ROOT_GROUP
+  SKIP("SYCL_EXT_ONEAPI_ROOT_GROUP is not defined");
+#else
+  check_root_group_api<struct RootGroup1D, 4>();
+  check_root_group_api<struct RootGroup2D, 4, 4>();
+  check_root_group_api<struct RootGroup3D, 4, 4, 4>();
```
keryell, line 239: Avoid using the same numbers as it might hide some bugs.

tests/extension/oneapi_root_group/root_group.cpp
```diff
@@ -0,0 +1,256 @@
...
+      get_group_linear_id,
+      get_local_linear_id,
+      get_group_linear_range,
+      get_local_linear_range,
+      leader,
+      get_child_group___root_group,
+      get_child_group___group,
+      COUNT,
+    };
+  };
```
AlexeySachkov, line 62: Why not use a [scoped `enum`](https://en.cppreference.com/w/cpp/language/enum)?

```suggestion
  enum class checks {
    get_group_id,
    get_local_id,
    get_group_range,
    get_local_range,
    get_max_local_range,
    get_group_linear_id,
    get_local_linear_id,
    get_group_linear_range,
    get_local_linear_range,
    leader,
    get_child_group___root_group,
    get_child_group___group,
    COUNT,
  };
```

tests/extension/oneapi_root_group/root_group.cpp
```diff
@@ -0,0 +1,256 @@
...
+    for (const auto& it : acc) {
+      passed &= it[check];
+    }
+    results[check] = passed;
+  }
+
+  using rg = sycl::ext::oneapi::experimental::root_group<Dimensions>;
+
+  // type aliases
+  CHECK(std::is_same_v<typename rg::id_type, sycl::id<Dimensions>>);
```
AlexeySachkov, line 112: I wonder what is a strategy/preferred way of handling type checks in CTS? Should such checks be implemented using a `static_assert` instead to allow earlier detection of bugs?

tests/extension/oneapi_root_group/root_group.cpp
```diff
@@ -0,0 +1,256 @@
...
+static void check_root_group_api() {
+  constexpr int Dimensions = sizeof...(Dims);
+  auto q = sycl_cts::util::get_cts_object::queue();
+  auto bundle =
+      sycl::get_kernel_bundle<sycl::bundle_state::executable>(q.get_context());
+  auto kernel = bundle.get_kernel<KernelName>();
+  auto maxWGs = kernel.template ext_oneapi_get_info<
+      sycl::ext::oneapi::experimental::info::kernel_queue_specific::
+          max_num_work_group_sync>(q);
+  REQUIRE(maxWGs >= 1);
```
AlexeySachkov, line 37: I don't think that the query or the `use_root_sync`  property is necessary for this test, according to the [extension spec](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_root_group.asciidoc#overview):

> An instance of the `root_group` class can be accessed by any ND-range kernel, since all information it exposes via its member functions is equivalent to information already available via members of `sycl::nd_item`.

Therefore, I would that that it makes sense to check that `root_group` class is still accessible and returns the right values regardless of the property or launch configuration. Perhaps, this particular test case can have a few variations like with, or without the property.

Tagging @Pennycook here for additional opinion on the matter.

tests/extension/oneapi_root_group/root_group.cpp
```diff
@@ -0,0 +1,256 @@
...
+  check_root_group_api<struct RootGroup1D, 4>();
+  check_root_group_api<struct RootGroup2D, 6, 4>();
+  check_root_group_api<struct RootGroup3D, 8, 6, 4>();
+#endif
+}
+
+TEST_CASE(
+    "Test for \"Root Group\" extension, check root_group synchronization using "
+    "group_barrier function"
+    "[oneapi_root_group_barrier]") {
```
AlexeySachkov, line 246: I wonder if those two test cases should instead have the same tag which corresponds to the extension, to be able to specifically launch just those two test cases?

Whether we need extra tags for additional filtering (like `api` or `barrier`) I'm not sure - this test looks small enough so that doesn't seem necessary. No idea if there is some CTS-wide guideline about tags usage, any recommendations from other reviewers?

tests/extension/oneapi_root_group/root_group.cpp
Pennycook: I agree that we should be checking the API works for kernels without the root-sync property.

If we want to cut down on the number of combinations to check, we could run the API test at the same time as the barrier test.  Then we'd only need two tests: one that checks the API works in kernels without the root-sync property; and another that checks the API (including barriers) works in kernels with the root-sync property.

tests/extension/oneapi_root_group/root_group.cpp
AlexeySachkov: > I didn't know it's valid to reuse a tag

Based on the [documentation](https://github.com/catchorg/Catch2/blob/v2.x/docs/test-cases-and-sections.md#tags), those tags don't have to be unique, but rather intended to provide functionality to group tests into various categories to have more flexibility when launching them.

tests/extension/oneapi_root_group/root_group.cpp
AlexeySachkov: Good point about `static_cast`, I forgot about it. In that case I agree that the current approach is better

AlexeySachkov (approved):

keryell (approved):

PR #800, opened 2023-09-13
Fix invalid argument for nd-range kernels

steffenlarsen:
This commit fixes the use of item and id arguments in nd_range parallel_for kernels.


ProGTX (approved):

bader:
@steffenlarsen, please, drop tests/optional_kernel_features/kernel_features_speculative_compilation.cpp changes. They conflict with https://github.com/KhronosGroup/SYCL-CTS/pull/798.

PR #801, opened 2023-09-13
Fix vector_api test

KornevNikita:
1. According to https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#swizzled-vec-class __swizzled_vec__ shouldn't escape the expression it was constructed in. So I refactored the test to create __swizzled_vec__ and then call a required function in the same place.

2. There is a bug in check_vector_convert() Since check_vector_convert_modes() accepts sycl::vec argument, we don't really test convert() method for __swizzled_vec__ as it's impicitly converted to sycl::vec. Also refactored this part to fix it.

### DIFF
--- a/tests/common/common_vec.h
+++ b/tests/common/common_vec.h
@@ -211,13 +211,35 @@ void handleFPToUnsignedConv(sycl::vec<vecType, N>& inputVec) {
   }
 }
 
+#define DO_OPERATION_ON_SWIZZLE(inputVec, ResVariable, Op)                   \
+  if constexpr (N == 1) {                                                    \
+    ResVariable = inputVec.template swizzle<sycl::elem::s0>() Op             \
+  } else if constexpr (N == 2) {                                             \
+    ResVariable =                                                            \
+        inputVec.template swizzle<sycl::elem::s0, sycl::elem::s1>() Op       \
+  } else if constexpr (N == 3) {                                             \
+    ResVariable = inputVec.template swizzle<sycl::elem::s0, sycl::elem::s1,  \
+                                            sycl::elem::s2>() Op             \
+  } else if constexpr (N == 4) {                                             \
+    ResVariable =                                                            \
+        inputVec.template swizzle<sycl::elem::s0, sycl::elem::s1,            \
+                                  sycl::elem::s2, sycl::elem::s3>() Op       \
+  } else if constexpr (N == 8) {                                             \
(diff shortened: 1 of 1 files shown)


tests/common/common_vec.h
```diff
@@ -334,7 +317,8 @@ bool check_vector_size_byte_size(sycl::vec<vecType, N> inputVec) {
   if (count != N || !noexcept(inputVec.size())) {
     return false;
   }
-  count = vector_swizzle_check<vecType, N>::get_swizzle(inputVec).size();
+  DO_OPERATION_ON_SWIZZLE(inputVec, count, .size();)
```
AlexeySachkov, line 320: I would suggest to put trailing `;` into the macro to avoid repeating it everywhere

KornevNikita: [1bd3720](https://github.com/KhronosGroup/SYCL-CTS/pull/801/commits/1bd372017ddb1be9c9876d19d7e13900dfdd11d9)

AlexeySachkov (approved):

tests/common/common_vec.h
```diff
@@ -211,13 +211,45 @@ void handleFPToUnsignedConv(sycl::vec<vecType, N>& inputVec) {
   }
 }
 
+#define DO_OPERATION_ON_SWIZZLE(inputVec, ResVariable, Op)                    \
+  if constexpr (N == 1) {                                                     \
```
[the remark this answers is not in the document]
KornevNikita: Done 5884e93

keryell (commented):
Looking at this macro. I wonder whether it would not be possible to have a function or lambda instead and just passing the operation as a a pointer to member function https://en.cppreference.com/w/cpp/language/pointer instead.

AlexeySachkov:
> Looking at this macro. I wonder whether it would not be possible to have a function or lambda instead and just passing the operation as a a pointer to member function https://en.cppreference.com/w/cpp/language/pointer instead.

Calls through function pointers are not allowed in device code - this would probably be the main blocker here

ProGTX (approved):

tests/common/common_vec.h
keryell: While I understand the economy, on the other hand I find this confusing while not looking inside the macro.
It might also confuse the indentation done with editors and formatters.
But if it works with `clang-format` I am OK with this.

PR #804, opened 2023-09-26
Store scan results by group local ID instead of global ID

steffenlarsen:
Since sub-group IDs are not required to follow the same ordering as the group, the scan test also needs to consider this when computing the results of the sub-group scan. This commit fixes it so the test makes no ordering assumption about neither group nor sub-group, avoiding both faulty expectations and invalid accesses.


keryell:
All this code becomes quite convoluted with quite a lot of implicit cognitive burden. Since it might be clear in your brain right now, could you comment the code with the information from the PR plus a lot more :-) so even next week we can understand the code and the context? ;-)

steffenlarsen:
@keryell - I added some comments. Are these in line with what you were thinking?

keryell (approved):
Great! Thanks.
I wish there were a lot of comments like these in all the PR. :-)

PR #805, opened 2023-09-27
Explicitly include <cmath> where required

frasercrmck:
These tests were all relying on `<cmath>` being included by `<sycl/sycl.hpp>`, which cannot be relied upon.

See https://github.com/intel/llvm/commit/9bbbc77 for a change which exposed this problem when building with DPC++.


bader (approved):

keryell (approved):
Thanks.

PR #806, opened 2023-09-27
Info-print all types in joint scan tests

frasercrmck:


bader (approved):

keryell (approved):
Thanks.
In the future we need to rename U and V to something sensible.

steffenlarsen (approved):
More info is more good. 👍

PR #807, opened 2023-09-27
Fix sub-group 'of' tests

frasercrmck:
As discovered/discussed in
https://github.com/KhronosGroup/SYCL-CTS/pull/795, the `predicate_function_of_sub_group` test was using the global linear ID to perform sub-group 'of' checks.

The checks were only being run on the first sub-group (with ID 0). This was implicitly relying on the first sub-group containing the work-items with IDs from 0 up to the sub-group size, as the checks included conditions like exactly one ID being 1. This is not correct in general, as the mapping between work-items and sub-group items is not defined.

In addition, this commit also enhances the sub-group tests to check that *all* sub-groups in the work-group return the correct result. Before we were either checking only the *first* sub-group
(`predicate_function_of_sub_group`) or checking all sub-groups but overwriting the results each time, so in effect only checking the *last* sub-group.

Now the results are initialized to 'true' and each sub-group is run and its results are merged with all of the other results.

### DIFF
--- a/tests/group_functions/group_of.h
+++ b/tests/group_functions/group_of.h
@@ -310,12 +310,22 @@ void predicate_function_of_sub_group(sycl::queue& queue) {
         sycl::sub_group sub_group = item.get_sub_group();
         T size = sub_group.get_local_linear_range();
 
+        // Use the sub-group local ID (plus 1) as a variable against which to
+        // test our predicates. Note that this has a well-defined set of values
+        // [1,2,...,N] where N is the sub-group size. Note that the sub-group
+        // could also just be of size 1.
         T local_var(sub_group.get_local_linear_id() + 1);
 
         // predicates
+        // The variable is never 0 for any member of the sub-group
         auto none_true = [&](T i) { return i == 0; };
+        // Exactly one member of the sub-group has value 1 (the first)
         auto one_true = [&](T i) { return i == 1; };
+        // Some (or all, for sub-groups of size 1) members of the sub-group have
+        // this value
         auto some_true = [&](T i) { return i > size / 2; };
+        // The variable is less than or equal to the sub-group size for all
+        // members of the sub-group.
         auto all_true = [&](T i) { return i <= size; };
 
         {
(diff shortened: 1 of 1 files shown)


keryell:
That sounds good. Probably apply some of https://github.com/KhronosGroup/SYCL-CTS/pull/804#issuecomment-1739068171
Do not forget that when we read the code in the future we do not have access to the PR description.

bader (approved):

frasercrmck:
> That sounds good. Probably apply some of [#804 (comment)](https://github.com/KhronosGroup/SYCL-CTS/pull/804#issuecomment-1739068171) Do not forget that when we read the code in the future we do not have access to the PR description.

I've added some extra comments now. I'm not sure exactly what suggestions you were making, so hopefully this is what you had in mind.

tests/group_functions/group_of.h
```diff
@@ -306,47 +310,57 @@ void predicate_function_of_sub_group(sycl::queue& queue) {
...
 
-        T local_var(item.get_global_linear_id() + 1);
+        // Use the sub-group local ID (plus 1) as a variable against which to
+        // test our predicates. Note that this has a well-defined set of values
+        // [1,2,...,N] where N is the sub-group size. Note that the sub-group
+        // could also just be of size 1.
+        T local_var(sub_group.get_local_linear_id() + 1);
 
         // predicates
+        // The variable is never 0 for any member of the sub-group
```
keryell, line 320:
```suggestion
        // The variable is never 1 for any member of the sub-group
```
or
```suggestion
        // The variable is always 0 for any member of the sub-group
```
?

tests/group_functions/group_of.h
```diff
@@ -493,23 +511,32 @@ void bool_function_of_sub_group(sycl::queue& queue) {
...
         T size = sub_group.get_local_linear_range();
 
+        // Use the sub-group local ID (plus 1) as a variable against which to
+        // test our predicates. Note that this has a well-defined set of values
+        // [1,2,...,N] where N is the sub-group size. Note that the sub-group
+        // could also just be of size 1.
         T local_var(sub_group.get_local_linear_id() + 1);
 
         // predicates
+        // The variable is never 0 for any member of the sub-group
```
keryell, line 521: Same as above?

keryell (commented):
This is quite clearer! \o/

tests/group_functions/group_of.h
frasercrmck: I'll go with the former as it better matches the predicate's name. Thanks!

tests/group_functions/group_of.h
frasercrmck: Done, thanks!

frasercrmck:
I don't suppose someone would be able/willing to squash this branch when merging? It's become a little messy (at least to my mind). No worries either way, though.

keryell (approved):
Thanks. I prefer keeping the history even it is messy, otherwise this is no longer history. :-)

PR #808, opened 2023-09-27
Change local accessor exception test to not use USM

RossBrunton:
The "exception for local accessor" test previously used a USM shared memory pointer as a write destination to produce a side effect. However, since not all devices support Shared USM and it doesn't seem to be a requirement of the test, it has been replaced with a "normal" accessor.


bader (approved):

keryell (approved):
Thanks.

steffenlarsen (approved):

PR #809, opened 2023-10-04
Create test plan for `auto_local_range` extension

0x12CC:
This PR adds a test plan for the [`auto_local_range`](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_auto_local_range.asciidoc) extension.

### DIFF
--- a/test_plans/auto_local_range.asciidoc
+++ b/test_plans/auto_local_range.asciidoc
@@ -1,0 +1,37 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl_ext_oneapi_auto_local_range
+
+This is a test plan for the API described in
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_auto_local_range.asciidoc[sycl_ext_oneapi_auto_local_range].
+
+
+== Testing scope
+
+=== Device coverage
+
+All of the tests described below are performed only on the default device that
+is selected on the CTS command line.
+
+=== Feature test macro
+
+All of the tests should use `#ifdef SYCL_EXT_ONEAPI_AUTO_LOCAL_RANGE` so they can be skipped
+if feature is not supported.
+
+== Tests
+
+* All following tests run with `Dimensions` = 1, 2, 3
+
+=== auto_range function
+
+Check that `auto_range<Dimensions>()` return type is `range<Dimensions>`
+
+=== auto_range in parallel_for
+
+For following `parallel_for` functions:
+
+* `queue::parallel_for`
+* `handler::parallel_for`
+
+Check that `auto_range<Dimensions>()` as the local range parameter behaves as expected: when each work item writes 1 into zero-initialized global memory all array values read after the `parallel_for` invocation are equal to 1 and the number of array values is equal to the global range.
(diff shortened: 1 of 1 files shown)


test_plans/auto_local_range.asciidoc
```diff
@@ -0,0 +1,37 @@
...
+Check that `auto_range<Dimensions>()` return type is `range<Dimensions>`
+
+=== auto_range in parallel_for
+
+For following `parallel_for` functions:
+
+* `queue::parallel_for`
+* `handler::parallel_for`
+
+Check that `auto_range<Dimensions>()` as the local range parameter behaves as expected: when each work item writes 1 into zero-initialized global memory all array values read after the `parallel_for` invocation are equal to 1 and the number of array values is equal to the global range.
```
gmlueck, line 37:
```suggestion
Check that `auto_range<Dimensions>()` as the local range parameter behaves as expected: when each work item writes 1 into zero-initialized global memory all array values read after the `parallel_for` invocation are equal to 1.
```

I don't think it makes sense to check the number of array elements.  The test will allocate some fixed number of array elements either in a buffer or a USM allocation, and the `parallel_for` won't change the number of array elements.  It only makes sense to test the contents of the array.

gmlueck, line 37: I think there should be some more complicated test that makes use of the `group` or `sub_group` object, since that is the main purpose of this extension.

@Pennycook can you suggest something?  It seems to me that many code patterns are hard to use with this extension because you would need to know either the number of work-groups or the size of a work-group prior to launching the kernel.  (For example, if you want to allocate some SLM for each work-item in a work-group.)  However, this extension won't allow that because you can't know the number of work-groups or their size until after the kernel is launched.

Pennycook: Something like a histogram, maybe?  The amount of local memory required in that case is derived from the size of the number of bins in the histogram, rather than the size of the group.

The code would do something like:
- Pass in an input pointer, an output pointer, and a local accessor of some small number of bins.
- Distribute the input range over the (automatic) group, and update the histogram in local memory (atomically).
- Issue a group barrier.
- Copy the histogram into the output pointer.

The dependencies on conformant atomics and algorithms isn't ideal, but that's the most complete example I can think of.

Anything that distributes an input range over the group members would work.  A simpler version of the above would be to do something like:
```c++
int local_accumulator = 0;
for (int i = g.get_local_id(); i < N; i += g.get_group_range()) {
  local_accumulator += input[i];
}
int total = sycl::reduce_over_group(g, local_accumulator, sycl::plus<>());
```

gmlueck: I think `N` is the global range and `input` is an array of `N` elements.  It's kind of a synthetic test, but each work-group ends up summing all the elements in the `input` array.  The test can validate its calculation by doing this:

```
int local_accumulator = 0;
for (int i = g.get_local_id(); i < N; i += g.get_group_range()) {
  local_accumulator += input[i];
}
int total = sycl::reduce_over_group(g, local_accumulator, sycl::plus<>());
if (total != expected_sum)
  /* test fails */
```

(Where `expected_sum` is computed on the host before the kernel is enqueued.)

The purpose of the test is to make sure that a kernel launched with `auto_range` can actually use the `group` object, which is the main reason to use `auto_range` vs. a simple `range` parallel-for.

gmlueck: Oh yes, I think you are right.  I think the `get_group_range` must have been a typo.

Pennycook: Oops.  Yes, that's a typo.  Sorry for the confusion!

gmlueck (approved):
Looks good!

PR #811, opened 2023-10-12
Use implementation name for configuration in conformance script

steffenlarsen:
This commit makes the run_conformance_tests.py script use the implementation name given in its related argument as the SYCL_IMPLEMENTATION argument in the CMake configuration.


steffenlarsen:
Note: I don't know if it was intentionally not used for the purpose previously, but it seems like it should be.

keryell (approved):

PR #813, opened 2023-10-12
Enable range rounding for range tests

jzc:


tests/range/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
 file(GLOB test_cases_list *.cpp)
 
+if(SYCL_IMPLEMENTATION STREQUAL "DPCPP")
+  # For DPCPP, range rounding is disabled globally. However,
+  # the range_api test needs range rounding enabled to pass,
+  # so renable it here.
```
bader, line 6: AFAIK, this is non-standard feature, so it's surprising that range_api test needs it. Could you add more details about the reasons for requiring range rounding, please?

jzc: I've edited the comment. And to clarify once more, range rounding **enabled** is DPCPP's default behavior. SYCL-CTS's cmake disables range rounding initially as an optimization. Perhaps this shouldn't be disabled in the first place to match DPCPP's default behavior as closely as possible?

https://github.com/KhronosGroup/SYCL-CTS/blob/cf553900567efa756bb5f560e0332cb6968acc2c/cmake/FindDPCPP.cmake#L47-L48

bader: > However, range rounding enables launching of kernels over >32bit ranges,

According to my understanding the range rounding is performance optimization, so it's not clear how it extends the supported nd-range. Doesn't it round the range to wg-size divisible value? If so, how does it enable nd-ranges which are not supported before the rounding? The rounding can only increase the nd-range size, right?

I would rather expect `-fsycl-id-queries-fit-in-int` to impact the supported nd-ranges.

jzc: Since https://github.com/intel/llvm/pull/10864, the range rounding behavior was extended to decrease the global range if it's too big, and having each work item perform multiple invocations of the kernel, allowing for global ranges larger than the backend allows. (only for range parallel_for, not nd_range parallel_for.)

bader: > having each work item perform multiple invocations of the kernel

I'm surprised that this behavior is enabled by "range rounding" macro as it does not sound to be related to range rounding.

Anyway. This is the question for DPC++ implementation. 

Thanks for clarification!

bader (approved):

keryell (approved):

PR #817, opened 2023-10-18
[NFC] Refactor `check_lo_hi_odd_even` helper

AlexeySachkov:
Turned it into a more generic function template to avoid explicitly specializing it for every vector size.


AlexeySachkov:
Note for reviewers: I took a version which we had for 16-element vector and generalized it by:
- using new `DO_OPERATION_ON_SWIZZLE` macro for swizzle-based operations
- doing some special-casing for 3-element vectors to skip certain checks (which we should probably do, but this PR is a mere refactoring)

keryell (approved):
Quite shorter!

PR #819, opened 2023-10-20
Fix sub_group_mask_constructors

KornevNikita:
`&value + shift` is invalid.

It's assumed to get the address of &value + (N = shift) bytes.
But we're actually getting &value + sizeof(value) * shift and running out of buffer.

### DIFF
--- a/tests/extension/oneapi_sub_group_mask/sub_group_mask_constructors.cpp
+++ b/tests/extension/oneapi_sub_group_mask/sub_group_mask_constructors.cpp
@@ -37,7 +37,7 @@ void copy_ulong_long2marray(sycl::marray<T, dim>& marray,
       elem_size * dim > sizeof(value) ? sizeof(value) : elem_size * dim;
   size_t shift = 0;
   size_t i = 0;
-  char* valuePtr = reinterpret_cast<char*>(&value);
+  auto valuePtr = reinterpret_cast<char*>(&value);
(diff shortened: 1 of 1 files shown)


KornevNikita:
@keryell @ProGTX could you please take a look?

tests/extension/oneapi_sub_group_mask/sub_group_mask_constructors.cpp
```diff
@@ -37,8 +37,9 @@ void copy_ulong_long2marray(sycl::marray<T, dim>& marray,
       elem_size * dim > sizeof(value) ? sizeof(value) : elem_size * dim;
   size_t shift = 0;
   size_t i = 0;
+  char* valuePtr = reinterpret_cast<char*>(&value);
```
keryell, line 40: Do not repeat yourself.
```suggestion
  auto valuePtr = reinterpret_cast<char*>(&value);
```

tests/extension/oneapi_sub_group_mask/sub_group_mask_constructors.cpp
```diff
@@ -52,8 +53,9 @@ void copy_marray2ulong_long(unsigned long long& value,
...
```
keryell, line 56: DRY
```suggestion
  auto valuePtr = reinterpret_cast<char*>(&value);
```

keryell (approved):
Thanks.

PR #822, opened 2023-10-25
Improve support detection for USM copy tests

RossBrunton:
Different USM allocation types are locked behind different aspects. The USM tests account for this by skipping the test if the aspect for the destination allocation is not present. However, two tests (copy and memcpy) have an allocation for the source memory, which is not checked.

This change adds a "supports_device" method to the USM tests that takes a queue and returns whether the test supports that queue. For the copy and memcpy tests, this checks that the source allocation is supported by the device. All other tests just return true.

(For the avoidance of doubt; I'm employed by Codeplay, a subsidiary of Intel)

### DIFF
--- a/tests/usm/usm_api.h
+++ b/tests/usm/usm_api.h
@@ -251,6 +251,13 @@ class copy : public copyGeneric<T, count, sourceAllocation> {
    */
   static constexpr bool has_non_usm_support() { return true; }
 
+  /** @brief This test only works on devices with `sourceAllocation` support
+   */
+  static bool supports_device(sycl_cts::util::logger& log,
+                              const sycl::queue& queue) {
+    return check_device_support<sourceAllocation>(log, queue);
+  }
+
   template <allocation alloc>
   static std::string description() {
     return "copy from " + get_allocation_decription<sourceAllocation>() +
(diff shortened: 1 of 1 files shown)


bader (approved):

illuhad (approved):

tomdeakin:
Ready to merge

tests/usm/usm_api.h
```diff
@@ -251,6 +251,13 @@ class copy : public copyGeneric<T, count, sourceAllocation> {
    */
   static constexpr bool has_non_usm_support() { return true; }
 
+  /** @brief This test only works on devices with `sourceAllocation` support
+   */
+  static bool supports_device(sycl_cts::util::logger& log,
+                              const sycl::queue& queue) {
+    return check_device_support<sourceAllocation>(log, queue);
```
keryell, line 258: That looks good.
Just thinking we could refactor globally the tests to have kinds of traits to express this at the class level to avoid all this copy paste of `supports_device()`.
Such as instead of supports_device
```c++
class copy : public copyGeneric<T, count, sourceAllocation> {
```
we could have just
```c++
class copy : public copyGeneric<T, count, sourceAllocation>, public supports_device<sourceAllocation> {
```

keryell (approved):

tests/usm/usm_api.h
bader: Sorry, hit merge too early. @RossBrunton, wound you mind applying this suggestion, please?

RossBrunton: Sure, I can create a follow up merge request for this.

Unfortunately I've not really worked with traits much in C++, and am a bit unsure on what you are suggesting here. What would a definition of `supports_device` look like?

bader: I can only guess, but one way I see it can be done is by defining two class templates: "no_check" and "sourceAllocation_check". Both of them define `supports_device` method according to their names. `copy` and `memcpy` inherit from "sourceAllocation_check" template. All other classes from "no_check" template.
@keryell is that what you have in mind?

keryell: Yes, this is the idea.

PR #824, opened 2023-10-27
Do not use `assert` in device code

AlexeySachkov:
`assert` is not guaranteed to be supported by a SYCL implementation and therefore we can't really use it in CTS.

Removed `assert` uses from velue operations helpers, which are used by some of CTS tests (`accessor`, `vector_swizzles`).


bader (approved):

keryell (approved):
Thanks.

PR #825, opened 2023-10-31
Refactor `vector_swizzles` test

AlexeySachkov:
This is a partial refactoring of the test, which aims to:
- reduce kernel size, which allows to speed up JIT compilation and therefore test execution
- improve error reporting from the test, by returning more than just a single `bool` value for all of the checks performed by kernels

Main changes:
- introduced new template for swizzle test kernels
- results of "per-element" `.swizzle<s0, s1, ..., sN>()` operations are now reported back to host where their verification happens instead of doing so on device side

### DIFF
--- a/tests/common/common_python_vec.py
+++ b/tests/common/common_python_vec.py
@@ -180,6 +180,42 @@ kernel_template = Template("""  bool resArray[1] = {true};
   CHECK(resArray[0]);
 """)
 
+# Unlike in the template above, some of the checks which are done by tests for
+# swizzles are outlined to host code.
+# This is done to reduce kernel size and have more information about a problem
+# in case of a test failure.
+swizzle_kernel_template = Template("""
+  {
+    auto vecBuffer = sycl::buffer<sycl::vec<${type}, ${size}>, 1>(
+        sycl::range<1>(total_per_element_swizzle_test_cases));
+    bool resArray[1] = {true};
+    {
+      sycl::buffer<bool, 1> boolBuffer(resArray, sycl::range<1>(1));
+      testQueue.submit([&](sycl::handler &cgh) {
+        sycl::accessor resAcc(boolBuffer, cgh, sycl::write_only);
+        sycl::accessor vecAcc(vecBuffer, cgh, sycl::write_only);
+
+        cgh.single_task<class ${kernelName}>([=]() {
+          ${test}
+        });
+      });
+    }
+    INFO("Checking ${testName}");
+    CHECK(resArray[0]);
+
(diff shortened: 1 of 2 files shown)


tests/common/common_python_vec.py
```diff
@@ -419,37 +482,23 @@ class SwizzleData:
...
-            ${type} in_order_vals[] = {${in_order_vals}};
             sycl::vec<${type}, ${size}> inOrderSwizzleFunctionVec {${name}DimTestVec.template swizzle<${in_order_swiz_indexes}>()};
+            vecAcc[in_order] = inOrderSwizzleFunctionVec;
             if (!check_equal_type_bool<sycl::vec<${type}, ${size}>>(inOrderSwizzleFunctionVec)) {
                 resAcc[0] = false;
             }
             if (!check_vector_size<${type}, ${size}>(inOrderSwizzleFunctionVec)) {
                 resAcc[0] = false;
             }
-            if (!check_vector_values<${type}, ${size}>(inOrderSwizzleFunctionVec, in_order_vals)) {
```
AlexeySachkov, line 450: Note for reviewers: there are two templates for kernel code:
- one for vectors with 4 or less elements, which is used to exercise all simple swizzles like `.xyzw`; modified `swizzle_elem_template` is its part
- another one for bigger vectors, where only per-element swizzles are tested; `swizzle_full_test_template` is used for it

Kernels for both categories of vectors do the same checks to ensure that `.swizzle` method returns elements in the right order, they are just described by two different code templates.

AlexeySachkov:
Ideally, we should get a better rewrite of the test to be able to emit separate error message for (almost) every check performed by the test, so SYCL implementors can quickly understand which exact operation doesn't work properly. Plus, we do not test swizzles functionality on host where it should also work, according to the SYCL 2020 specification.

However, I'm not able to provide such a rewrite at the moment. Nevertheless, I would appreciate if this partial refactoring is accepted, as it greatly helps to reduce to reduce amount of code emitted for device. I don't have pure numbers, because I have some local changes applied to `intel/llvm` implementation which also help this particular test, but I would estimate that this patch reduces device code size by 30-40% when CTS is compiled by `intel/llvm` implementation.

keryell (approved):
Thanks.
You can probably add your idea for further refactoring as a TODO comment somewhere so it is not lost.

AlexeySachkov:
> Thanks. You can probably add your idea for further refactoring as a TODO comment somewhere so it is not lost.

Recorded key points in 35fbe6511cd9621700083bdbff1fd558e33ea478

keryell (approved):
Thanks for the on-coming generations of developers. :-)

PR #826, opened 2023-11-02
Fix typos in atomic_ref test names

frasercrmck:
I'm assuming these are typos.

There are some inconsistencies about how we style the test names for ostensibly the same operations: see tests covering one operator (`atomic_ref::operator`) vs those that cover two (`atomic_ref operator/operator` - no colons). I see why it was done, but the inconsistency is noticeable.


keryell (approved):
This is why we need more code introspection in future C++ to avoid writing such boilerplate description manually. :-)

PR #827, opened 2023-11-03
Using vector<unsigned char> instead of vector<bool>

ChengChen002:
Since the std::vector\<bool\> type does not have a data() member function,
Using vector<unsigned char> instead of vector<bool>.

### DIFF
--- a/tests/usm/usm_api.h
+++ b/tests/usm/usm_api.h
@@ -513,8 +513,8 @@ template <typename T, size_t buf_size>
 class event_generator {
   // Change vector<bool> to vector <unsigned char>, since vector<bool> is
   // stored bit-wise, and does not have data() member function.
-  using ContainType = typename std::conditional<std::is_same_v<bool, T>,
-                                                unsigned char, T>::type;
+  using ContainType = typename std::conditional_t<std::is_same_v<bool, T>,
+                                                  unsigned char, T>;
   sycl::range<1> rng{buf_size};
   std::vector<ContainType> arr_src;
   std::vector<ContainType> arr_dst;
--- a/tests/usm/usm_api.h
+++ b/tests/usm/usm_api.h
@@ -575,6 +575,75 @@ class event_generator {
   }
 };
 
+// Since the std::vector<bool> type does not have a data() member function,
+// std::array is used instead.
+template <size_t buf_size>
+class event_generator<bool, buf_size> {
+  sycl::range<1> rng{buf_size};
+  std::array<bool, buf_size> arr_src;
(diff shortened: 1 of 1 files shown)


ChengChen002:
I haven't come up with a cleaner way to fix this yet, please comment if you have any other suggestions! Thanks.

bader:
> I haven't come up with a cleaner way to fix this yet, please comment if you have any other suggestions! Thanks.

Please, push "Ready for review" button if you want this PR to be reviewed.

keryell (commented):
This looks correct but there is a lot of code duplication.
I was thinking about introducing a generic container type that would select either a `std::vector` or a `std::array` according to the value type, but thinking again, why not just using always a `std::array`?

ChengChen002:
> This looks correct but there is a lot of code duplication. I was thinking about introducing a generic container type that would select either a `std::vector` or a `std::array` according to the value type, but thinking again, why not just using always a `std::array`?

This is to avoid using too much stack memory, you can see more information in #823.

keryell (approved):
Thanks! I love shorter code for maintenance. :-)

tests/usm/usm_api.h
```diff
@@ -511,30 +511,34 @@ struct gen_buf_size_selector<0> {
  */
 template <typename T, size_t buf_size>
 class event_generator {
+  // Change vector<bool> to vector <unsigned char>, since vector<bool> is
+  // stored bit-wise, and does not have data() member function.
+  using ContainType = typename std::conditional<std::is_same_v<bool, T>,
+                                                unsigned char, T>::type;
```
keryell, line 517: C++14
```suggestion
  using ContainType = typename std::conditional_t<std::is_same_v<bool, T>,
                                                unsigned char, T>;
```

keryell (changes requested):

tests/usm/usm_api.h
```diff
@@ -511,30 +511,34 @@ struct gen_buf_size_selector<0> {
  */
 template <typename T, size_t buf_size>
 class event_generator {
+  // Change vector<bool> to vector <unsigned char>, since vector<bool> is
+  // stored bit-wise, and does not have data() member function.
+  using ContainType = typename std::conditional_t<std::is_same_v<bool, T>,
+                                                  unsigned char, T>;
```
bader, line 517:
```suggestion
  using ContainType =
      typename std::conditional_t<std::is_same_v<bool, T>, unsigned char, T>;
```

keryell (approved):
Thanks!

PR #828, opened 2023-11-03
[NFC] Drop legacy get/setElement helpers

AlexeySachkov:
SYCL 2020 introduced `operator[]` to `vec` class and we can use it directly instead of going through the helper.


keryell (approved):
Isn't SYCL great? :-)

PR #830, opened 2023-11-08
Compute group local linear id correctly

jiezzhang:
Local linear ID should be group_id * local_range + local_id


steffenlarsen (approved):
Good catch! This one is on me. 😄

frasercrmck (approved):
Thanks! I had _just_ debugging this test locally :slightly_smiling_face: 

Fixes the test for me.

keryell (approved):
Thanks.

PR #831, opened 2023-11-08
Do not check for undefined behavior in abs.

hvdijk:
Unlike OpenCL abs, but like C and C++ abs, in SYCL 2020, taking the absolute value of a signed type produces a signed result and the behavior is undefined if the absolute value is not representable (i.e. if the input is the greatest negative value), so do not check that the SYCL results match the reference results for this case.

### DIFF
--- a/util/math_reference.h
+++ b/util/math_reference.h
@@ -285,10 +285,23 @@ sycl::marray<T, N> select(sycl::marray<T, N> a, sycl::marray<T, N> b,
 
 /* absolute value */
 template <typename T>
-T abs(T x) {
-  return x < 0 ? -x : x;
+sycl_cts::resultRef<T> abs(T x) {
+  using U = std::make_unsigned_t<T>;
+  T result = x < 0 ? T(-U(x)) : x;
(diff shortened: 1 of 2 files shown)


util/math_reference.h
```diff
@@ -285,10 +285,23 @@ sycl::marray<T, N> select(sycl::marray<T, N> a, sycl::marray<T, N> b,
 
 /* absolute value */
 template <typename T>
-T abs(T x) {
-  return x < 0 ? -x : x;
+sycl_cts::resultRef<T> abs(T x) {
+  using U = typename std::make_unsigned<T>::type;
```
keryell, line 289:
```suggestion
  using U = typename std::make_unsigned_t<T>;
```

keryell (changes requested):
Almost there.

util/math_reference.h
hvdijk: Sure, and `typename` is not needed in that case either. Updated. I tried to go with the existing style but I guess you want to update gradually to the newer forms?

hvdijk:
ping @keryell

keryell (approved):
Thanks!

PR #832, opened 2023-11-09
[NFC] USM test device support refactored

RossBrunton:
The USM API tests previously each had a "supports_device" method that either returned true or checked whether a specific USM allocation type was supported.

This change moves that duplicated code to two inheritable classes that provide those methods instead.

----

This is a follow up implementing the suggestion made by @keryell here: https://github.com/KhronosGroup/SYCL-CTS/pull/822#discussion_r1380420474 .

(For the avoidance of doubt; I'm employed by Codeplay, a subsidiary of Intel)

### DIFF
--- a/tests/usm/usm_api.h
+++ b/tests/usm/usm_api.h
@@ -217,11 +217,35 @@ void check_values(const T *begin, const ReferenceIt &reference) {
   }
 }
 
+/** @brief Trait for tests that have no additional device requirements
+ *
+ * Tests implementing this trait do not require any device support beyond
+ * the destination memory location.
+ */
+struct noAdditionalDeviceRequirements {
+  static bool supports_device(sycl_cts::util::logger& log,
+                              const sycl::queue& queue) {
+    return true;
+  }
+};
+/** @brief Trait for tests that require support for a given USM alloc type
+ *
+ * Tests implementing this trait require device support for a specific
+ * type of allocation (e.g. as a source memory location).
+ */
+template <allocation allocationType>
+struct requiresUsmAllocationSupport {
+  static bool supports_device(sycl_cts::util::logger& log,
+                              const sycl::queue& queue) {
+    return check_device_support<allocationType>(log, queue);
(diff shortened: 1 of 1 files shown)


tests/usm/usm_api.h
```diff
@@ -217,11 +217,35 @@ void check_values(const T *begin, const ReferenceIt &reference) {
...
+ *
+ * Tests implementing this trait do not require any device support beyond
+ * the destination memory location.
+ */
+struct noAdditionalDeviceRequirements {
+  static bool supports_device(sycl_cts::util::logger& log,
+                              const sycl::queue& queue) {
+    return true;
+  }
+};
```
keryell, line 230:
```suggestion
  static bool supports_device(sycl_cts::util::logger&,
                              const sycl::queue&) {
    return true;
  }
};

```
to avoid warning about unused arguments and more spacing between function definitions.

keryell (changes requested):
Thanks for the cleaning. Almost there.

tests/usm/usm_api.h
RossBrunton: Ah, using Github's resolve thing doesn't use the correct email. One sec.

tests/usm/usm_api.h
```diff
@@ -217,11 +217,36 @@ void check_values(const T *begin, const ReferenceIt &reference) {
...
+
+/** @brief Trait for tests that require support for a given USM alloc type
+ *
+ * Tests implementing this trait require device support for a specific
+ * type of allocation (e.g. as a source memory location).
+ */
+template <allocation allocationType>
+struct requiresUsmAllocationSupport {
+  static bool supports_device(sycl_cts::util::logger& log,
+                              const sycl::queue& queue) {
```
bader, line 240:
```suggestion
  static bool supports_device(sycl_cts::util::logger&, const sycl::queue&) {
```

keryell (approved):
Thanks!

PR #836, opened 2023-11-22
Change abs_diff return type

steffenlarsen:
Following https://github.com/KhronosGroup/SYCL-Docs/pull/458 the return
type of abs_diff was changed to be the same as the input types. This
commit makes the corresponding changes to the CTS tests for it.

Signed-off-by: Larsen, Steffen <steffen.larsen@intel.com>


keryell (approved):
Thanks.

PR #838, opened 2023-11-22
Fix overflow issues in scan tests.

maarquitos14:
Prior to this patch, there were a number of overflow issues when the output type of scan op was char and the init type was float. After local verification, all tests pass now.

### DIFF
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -25,6 +25,9 @@
 template <int D, typename T, typename U, typename I, typename OpT>
 class joint_scan_group_kernel;
 
+// This should never be higher than std::numeric_limits<T>::max() for the
+// smallest type tested. Currently, the smallest type tested is
+// char/int8_t, so it shouldn't be higher than 127.
 constexpr int init = 42;
 constexpr size_t test_size = 12;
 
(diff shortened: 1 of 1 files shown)


tests/group_functions/group_scan.h
```diff
@@ -54,14 +57,25 @@ auto joint_exclusive_scan_helper(Group group, T* v_begin, T* v_end,
...
+template <typename T, typename U, typename I, typename OpT>
 struct JointScanDataStruct {
-  JointScanDataStruct(size_t range_size)
+  JointScanDataStruct(size_t range_size, OpT op, bool with_init)
       : ref_input(range_size), res(range_size * 4, U(-1)) {
     std::iota(ref_input.begin(), ref_input.end(), T(1));
+    if constexpr (std::is_same_v<OpT, sycl::multiplies<I>> ||
+                  std::is_same_v<OpT, sycl::plus<I>>) {
+      auto identity = sycl::known_identity_v<OpT, I>;
+      auto acc = with_init ? I(init) : identity;
```
keryell, line 68: Since you care about overflowing here, should `I { ... }`  be used all over the place instead of `I( ... )` to test for some potential overflow at compile time?

maarquitos14: Good catch! Thanks!

bader:
@keryell, any other comments?

PR #839, opened 2023-11-24
[group] Refine scan_over_group for sub-group

Nuullll:
Use global linear id to index input data. Store both the sub-group id and item linear id within the sub-group to recover sub-group construction when verifying, so that we don't make any assumption on the implementation-defined sub-group partitioning and ordering.

### DIFF
--- a/tests/group_functions/group_scan.h
+++ b/tests/group_functions/group_scan.h
@@ -322,7 +322,8 @@ struct ScanOverGroupDataStruct {
   ScanOverGroupDataStruct(size_t range_size)
       : ref_input(range_size),
         res(range_size * 4, T(-1)),
-        local_id(range_size * 2, 0) {
+        local_id(range_size, 0),
+        sub_group_id(range_size, 0) {
     std::iota(ref_input.begin(), ref_input.end(), U(1));
   }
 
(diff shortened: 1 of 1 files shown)


tests/group_functions/group_scan.h
```diff
@@ -436,7 +436,7 @@ void check_scan_over_group(sycl::queue& queue, sycl::range<D> range, OpT op,
                 // Use the local id of the item in the sub-group to place
                 // results of the scan operation in the order of the items.
                 auto sg_index = sub_group.get_group_linear_id() *
-                                    sub_group.get_local_linear_range() +
+                                    sub_group.get_max_local_range()[0] +
```
steffenlarsen, line 439:
```suggestion
                                    sub_group.get_max_local_range() +
```
I don't believe we need the indexing. Just a nit though.

steffenlarsen (approved):
This looks right. It is easy to forget not all the groups have the same size.

Nuullll:
> This looks right. It is easy to forget not all the groups have the same size.

Thanks for reviewing. Please allow me to finish some extra testings.

Nuullll:
Neither `sub_group.get_local_linear_range()` nor `sub_group.get_max_local_range()` is reliable:
1. Not all the sub-groups are guaranteed to have the same size, so we shouldn't use `get_local_linear_range()`.
2. Considering the 2-dimensional NDRange (3, 4) with reqd_sub_group_size=16, then the implementation may construct 3 sub-groups (in the highest dimension) where `get_max_local_range() == 16` for all the sub-groups. It's a little counter-intuitive, but as noted in [OpenCL C Spec - cl_intel_required_subgroup_size](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_required_subgroup_size.html#_additions_to_section_6_7_2_optional_attribute_qualifiers):
> The optional __attribute__((intel_reqd_sub_group_size(<int>))) can be used to indicate that the kernel must be compiled and executed with the specified sub-group size. When this attribute is present, get_max_sub_group_size() is guaranteed to return the specified integer value. This is important for the correctness of many sub-group algorithms, and in some cases may be used by the compiler to generate more optimal code.

So using the global linear id directly is probably safer.

tests/group_functions/group_scan.h
```diff
@@ -416,11 +416,9 @@ void check_scan_over_group(sycl::queue& queue, sycl::range<D> range, OpT op,
...
                 sycl::sub_group sub_group = item.get_sub_group();
 
-                // Use the local id of the item in the group to place results of
-                // the scan operation in the order of the items.
-                auto g_index = group.get_group_linear_id() *
-                                   group.get_local_linear_range() +
-                               group.get_local_linear_id();
+                // Use the global linear id of the item in the group to place
+                // results of the scan operation in the order of the items.
+                auto g_index = item.get_global_linear_id();
```
steffenlarsen, line 421: I don't think this is an option. I believe this is exactly what we moved away from a couple of patches ago. The problem is that partial results of scan appear in the order of the items in the sub-group, which is not required to follow the order of the global ID.

We can use it, but then we will also have to store the order of the sub-group items and use that when verifying.

steffenlarsen (changes requested):
Blocking this to avoid regression.

tests/group_functions/group_scan.h
Nuullll: Thanks. I've read your previous PR #804. To make sure we are on the same page:
1. [SYCL spec - group class](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#group-class) requires both the **group ordering** and the **items ordering within each group**, to be consistent with the [linearized dimension](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:multi-dim-linearization), so using `item.get_global_linear_id()` should be equivalent to `group.get_group_linear_id() * group.get_local_linear_range() + group.get_local_linear_id()` for work-groups.
2. However, [SYCL spec - Work-group data parallel kernels](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_work_group_data_parallel_kernels) gives us no guarantees on sub-group ordering: 
> Portable device code must not assume that work-items within a sub-group execute in any particular order, that work-groups are subdivided into sub-groups in a specific way, nor that the work-items within a sub-group provide specific forward progress guarantees.

For the sub-group part, I think we need to store both the `sub_group.get_group_linear_id()` (as an abstract identifier of the sub-group) and the `sub_group.get_local_linear_id()` (as an identifier for the item within the sub-group) and then collect and reorder the reference input data for each sub-group when verifying. So that we permit the implementation the maximal freedom of sub-group partitioning and ordering.

tests/group_functions/group_scan.h
```diff
@@ -337,35 +338,76 @@ struct ScanOverGroupDataStruct {
     T init_value = with_init ? T(init) : sycl::known_identity<OpT, T>::value;
     // res consists of 4 series of results: two pairs of exclusive and inclusive
     // scan results made over 'group' and 'sub_group' accordingly.
-    for (int group_i = 0; group_i < 2; group_i++) {
-      std::string group_name = group_i == 0 ? "group" : "sub_group";
-      size_t group_offset = range_size * group_i;
+    {
+      std::string group_name = "group";
```
steffenlarsen, line 342: We might as well inline these again now that they don't share the loop.

steffenlarsen (approved):
Thank you, @Nuullll ! LGTM!

tests/group_functions/group_scan.h
Nuullll: Done.

Nuullll:
Any further comments? Can we merge this?

keryell (approved):
Great!

PR #842, opened 2023-12-07
Do not check for undefined behavior in abs_diff

steffenlarsen:
Similar to sycl::abs, sycl::abs_diff now has undefined behavior for unrepresentable results. This patch changes the generated tests for abs_diff to avoid UB, similar to how it was done for abs in https://github.com/KhronosGroup/SYCL-CTS/pull/831.

### DIFF
--- a/util/math_reference.h
+++ b/util/math_reference.h
@@ -306,31 +306,15 @@ sycl_cts::resultRef<sycl::marray<T, N>> abs(sycl::marray<T, N> a) {
 /* absolute difference */
 template <typename T>
 sycl_cts::resultRef<T> abs_diff(T a, T b) {
-  if constexpr (std::is_unsigned_v<T>) {
-    T h = (a > b) ? a : b;
-    T l = (a <= b) ? a : b;
-    return h - l;
-  } else {
-    using U = std::make_unsigned_t<T>;
-    constexpr T TMax = std::numeric_limits<T>::max();
-    // For unsigned integers, negative values might not be representable by
-    // an absolute value, so we compute the overflow separately.
-    T ao = a < -TMax ? a + TMax : 0;
-    T bo = b < -TMax ? b + TMax : 0;
-    // Then find the absolute values (without the overflow) and compute the
-    // unsigned difference.
-    U au = static_cast<U>(std::abs(a - ao));
-    U bu = static_cast<U>(std::abs(b - bo));
-    U h = (au > bu) ? au : bu;
-    U l = (au <= bu) ? au : bu;
-    U result = (a < 0) == (b < 0) ? h - l : h + l;
-    // Now re-add the absolute distance in overflow. Both ao and bo will be
-    // <= 0.
-    result += (ao <= bo ? U(bo - ao) : U(ao - bo));
-    // If the unsigned difference is larger than the signed maximum value, we
-    // have UB. Otherwise we have our valid result.
-    return result > U(TMax) ? sycl_cts::resultRef<T>(0, true) : T(result);
-  }
+  using U = std::make_unsigned_t<T>;
+  T h = (a > b) ? a : b;
+  T l = (a <= b) ? a : b;
+  // Use knowledge of 2-complement encoding while we know the difference is
+  // positive, so no overflow/underflow.
+  U result = static_cast<U>(h) - static_cast<U>(l);
+  return result > static_cast<U>(std::numeric_limits<T>::max())
+             ? sycl_cts::resultRef<T>(0, true)
+             : T(result);
 }
 template <typename T, int N>
 sycl_cts::resultRef<sycl::vec<T, N>> abs_diff(sycl::vec<T, N> a,
--- a/util/math_reference.h
+++ b/util/math_reference.h
@@ -309,10 +309,11 @@ sycl_cts::resultRef<T> abs_diff(T a, T b) {
   using U = std::make_unsigned_t<T>;
   T h = (a > b) ? a : b;
   T l = (a <= b) ? a : b;
-  // Use knowledge of 2-complement encoding while we know the difference is
-  // positive, so no overflow/underflow.
+  // Using two's-complement and that unsigned integer underflow is defined as
+  // modulo 2^n we get the result by computing the distance based on signed
+  // comparison.
   U result = static_cast<U>(h) - static_cast<U>(l);
-  return result > static_cast<U>(std::numeric_limits<T>::max())
+  return result > std::numeric_limits<T>::max()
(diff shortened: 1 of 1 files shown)


keryell (commented):
This sounds complicated. Would it not possible to do like the unsigned case but casting first to unsigned and adding a bias of - [ ] -min() on each element first?

steffenlarsen:
> This sounds complicated. Would it not possible to do like the unsigned case but casting first to unsigned and adding a bias of - [ ] -min() on each element first?

I worry we would run into the same problem causing this problem when computing `-min()` if it is the minimum value of an integer which is unrepresentable in its signed form.

keryell:
I mean for `uint8_t`  for example:
```
auto au = static_cast<U>(a) + 0x80;
auto bu = static_cast<U>(b) + 0x80;
auto h = (au > bu) ? au : bu;
auto l = (au <= bu) ? au : bu;
return h - l;
```
So technically `0x80` is `static_cast<U>(std::numeric_limits<T>::max())`

steffenlarsen:
Is something like https://godbolt.org/z/ejbK9qPdj what you are thinking?

util/math_reference.h
```diff
@@ -305,21 +305,45 @@ sycl_cts::resultRef<sycl::marray<T, N>> abs(sycl::marray<T, N> a) {
...
+    U h = (au > bu) ? au : bu;
+    U l = (au <= bu) ? au : bu;
+    U result = (a < 0) == (b < 0) ? h - l : h + l;
+    // Now re-add the absolute distance in overflow. Both ao and bo will be
+    // <= 0.
+    result += (ao <= bo ? U(bo - ao) : U(ao - bo));
+    // If the unsigned difference is larger than the signed maximum value, we
+    // have UB. Otherwise we have our valid result.
+    return result > U(TMax) ? sycl_cts::resultRef<T>(0, true) : T(result);
+  }
```
steffenlarsen, line 333:
```suggestion
  using U = std::make_unsigned_t<T>;
  auto normalize = [](T& x) {
    constexpr U TMin = static_cast<U>(std::numeric_limits<T>::min());
    U xu = static_cast<U>(x);
    if (x < 0)
      xu -= TMin;
    else
      xu += TMin;
    return xu;
  };
  U au = normalize(a);
  U bu = normalize(b);
  U h = (au > bu) ? au : bu;
  U l = (au <= bu) ? au : bu;
  U result = h - l;
  return result > static_cast<U>(std::numeric_limits<T>::max())
             ? sycl_cts::resultRef<T>(0, true)
             : T(result);
```

Suggestion based on https://github.com/KhronosGroup/SYCL-CTS/pull/842#issuecomment-1856165033 to illustrate how it could be integrated here.

keryell: I was thinking to https://godbolt.org/z/9Ma6nvGsz but I think I can even optimizing it further.

steffenlarsen: Doesn't that solution rely on underflowing behavior if either `a` or `b` is negative? AFAIK that is UB.

keryell: The shorter simplified version: https://godbolt.org/z/4Wb8x6GKG

steffenlarsen:
```suggestion
  using U = std::make_unsigned_t<T>;
  T h = (a > b) ? a : b;
  T l = (a <= b) ? a : b;
  // Use knowledge of 2-complement encoding while we know the difference is
  // positive, so no overflow/underflow.
  U result = static_cast<U>(h) - static_cast<U>(l);
  return result > static_cast<U>(std::numeric_limits<T>::max())
             ? sycl_cts::resultRef<T>(0, true)
             : T(result);
```

However, I worry that the comment is misleading. We are potentially causing an underflow if one of the operands are negative. Say for example `a = 1000` and `b = -1000`, the the static cast to unsigned for these would result in the computation `1000 - 4294966296`. That said, it is now for unsigned integer types, which I believe has well-defined overflow/underflow behavior (modulo 2^n).

steffenlarsen: I have applied your suggestion since it should be fine. If you agree that the comment is slightly misleading we can change it.

keryell: Yes, by "so no overflow/underflow." I mean with `signed`.
Once we are back in the `unsigned` world we are good, since as you said we are just doing modulo 2^n arithmetic.
I let you tweak the comment to explain this in a better way. :-)

keryell: The cast is redundant in
```
return result > static_cast<U>(std::numeric_limits<T>::max())
```
since in C/C++ there is this implicit (confusing...) conversion of signed to unsigned.

util/math_reference.h
```diff
@@ -305,21 +305,29 @@ sycl_cts::resultRef<sycl::marray<T, N>> abs(sycl::marray<T, N> a) {
...
-T abs_diff(T a, T b) {
+sycl_cts::resultRef<T> abs_diff(T a, T b) {
+  using U = std::make_unsigned_t<T>;
   T h = (a > b) ? a : b;
   T l = (a <= b) ? a : b;
-  return h - l;
+  // Use knowledge of 2-complement encoding while we know the difference is
+  // positive, so no overflow/underflow.
+  U result = static_cast<U>(h) - static_cast<U>(l);
+  return result > static_cast<U>(std::numeric_limits<T>::max())
```
steffenlarsen, line 315:
```suggestion
  // Using two's-complement and that unsigned integer underflow is defined as
  // modulo 2^n we get the result by computing the distance based on signed
  // comparison.
  U result = static_cast<U>(h) - static_cast<U>(l);
  return result > std::numeric_limits<T>::max()
```

util/math_reference.h
steffenlarsen: I have updated the comment and removed the redundant cast. 👍

util/math_reference.h
```diff
@@ -305,21 +305,30 @@ sycl_cts::resultRef<sycl::marray<T, N>> abs(sycl::marray<T, N> a) {
...
 /* absolute difference */
 template <typename T>
-T abs_diff(T a, T b) {
+sycl_cts::resultRef<T> abs_diff(T a, T b) {
+  using U = std::make_unsigned_t<T>;
   T h = (a > b) ? a : b;
   T l = (a <= b) ? a : b;
-  return h - l;
+  // Using two's-complement and that unsigned integer underflow is defined as
+  // modulo 2^n we get the result by computing the distance based on signed
```
keryell, line 313: Thanks, now I understand the code. :-)
But is it signed? It looks like it uses unsigned comparison.

keryell (approved):
Great!
We can always bike-shed the comment later. ;-)

PR #843, opened 2024-01-02
Fix reqd_work_group_size test

frasercrmck:
The test was using a 1-dimensional kernel attribute even with 2- and 3-dimensional work-group sizes. This is against the semantics of the attribute, and the implementation may throw an `nd_range` exception (though the specification isn't entirely clear).

Regardless, the test is not intentionally testing this aspect of the specification; it's instead checking whether an implementation throws a `kernel_not_supported` exception if N > "max work-group size", and a `nd_range` exception otherwise.


keryell (approved):
Thanks.

PR #844, opened 2024-01-04
Catching the right type of exception

ChengChen002:
It throws sycl::exception instead of sycl::feature_not_supported when failed to test kernel args with samplers.


keryell (approved):
Thanks.

PR #853, opened 2024-01-11
Fixed bug in queue test: `sizeof(int)` was used even if T!=int.

JackAKirk:
Fixed a bug where `sizeof(int)` was used even if T!=int, which meant char and fp16 cases failed. Now all queue tests pass for the CUDA backend.


bader (approved):
Except formatting issues, this change looks good to me.

keryell (approved):
Good catch!

Issue #821, opened 2023-10-24
`marray` test does not test everything (or anything?) on device


AlexeySachkov:
Maybe I'm missing something as I haven't reviewed the whole test, but here is at least one code path which I've traced:

https://github.com/KhronosGroup/SYCL-CTS/blob/eec8701491114d1930c90679ba5072dd2a5162cb/tests/marray/marray_operators_core.cpp#L29-L32

Which brings us to:

https://github.com/KhronosGroup/SYCL-CTS/blob/eec8701491114d1930c90679ba5072dd2a5162cb/tests/marray/marray_operators.h#L482-L486

Then:

https://github.com/KhronosGroup/SYCL-CTS/blob/eec8701491114d1930c90679ba5072dd2a5162cb/tests/marray/marray_operators.h#L150-L155

And finally:

https://github.com/KhronosGroup/SYCL-CTS/blob/eec8701491114d1930c90679ba5072dd2a5162cb/tests/marray/marray_operators.h#L120-L133

From what I can see, there is no mention of queues or devices at all. Perhaps it is hidden somewhere deep in the helpers, but I don't think that it is the case. Looks more like we really only test `marray` functionality on host

bader:
@AlexeySachkov is right. I think we can expect `run_unary_sequence` to run the test on the device, but it's not the case.
Similar approach is used by other tests e.g. atomic_ref. See base class in https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/atomic_ref/atomic_ref_test_base.h and derivative classes to run specific tests defining `run_on_device` method (e.g. https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/atomic_ref/atomic_ref_add_sub_op_all_types_test.h).

According to the [test plan](https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/math_array.asciidoc) the tests are supposed to be executed on the device.

Having said that, I don't expect marray implementation to be different on the host and SYCL devices. Unfortunately, the test doesn't check "device copyable" property neither.

tomdeakin:
Awaiting a PR to fix this.

PR #854, opened 2024-01-12
Test marray on device

steffenlarsen:
This commit makes marray tests run checks on both device and host, where they would only run on host previously. Additionally it splits operator tests up in separate categories to reduce compilation overhead.

Fixes https://github.com/KhronosGroup/SYCL-CTS/issues/821

### DIFF
--- a/tests/marray/marray_constructor.h
+++ b/tests/marray/marray_constructor.h
@@ -31,52 +31,32 @@ template <typename DataT, typename NumElementsT>
 class run_marray_constructor_test {
   static constexpr std::size_t NumElements = NumElementsT::value;
 
+  static constexpr size_t num_test_cases = 7;
+
+  static constexpr const char* check_names[num_test_cases] = {
+      "default constructor",
+      "scalar constructor",
+      "variadic constructor (NumElements DataT instances)",
+      "variadic constructor (one DataT instance, one marray instance)",
+      "variadic constructor (one marray instance, one DataT instance)",
+      "copy constructor",
+      "copy constructor rval reference"};
+
   using marray_t = sycl::marray<DataT, NumElements>;
 
- private:
-  template <size_t num_elements = NumElements,
-            std::enable_if_t<num_elements == 1, bool> = true>
-  void check_constexpr_single_element() {}
-
-  template <size_t num_elements = NumElements,
-            std::enable_if_t<num_elements != 1, bool> = true>
-  void check_constexpr_single_element() {
-    //  one DataT instance, one marray instance
-    {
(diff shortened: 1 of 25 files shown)


tests/marray/marray_constructor.h
```diff
@@ -31,81 +31,125 @@ template <typename DataT, typename NumElementsT>
...
-    INFO("for number of elements \"" << NumElements << "\": ");
+  using marray_t = sycl::marray<DataT, NumElements>;
 
+  template <typename IteratorT>
+  static void run_checks(IteratorT results) {
     // default constructor
     {
       marray_t ma;
-      CHECK(value_operations::are_equal(ma, DataT{}));
+      *(results++) = value_operations::are_equal(ma, DataT{});
```
keryell, line 52: `()` are redundant but it might be clearer too. So no problem.

tests/marray/marray_constructor.h
```diff
@@ -31,81 +31,125 @@ template <typename DataT, typename NumElementsT>
...
       constexpr auto a = marray_common::iota_marray<DataT, NumElements, 1>();
       marray_t ma_inc;
-      std::iota(ma_inc.begin(), ma_inc.end(), 1);
-      CHECK(value_operations::are_equal(ma_inc, a));
+      marray_common::iota(ma_inc.begin(), ma_inc.end(), 1);
+      *(results++) = value_operations::are_equal(ma_inc, a);
 
       // only compiled when NumElements != 1
-      check_constexpr_single_element();
+      if constexpr (NumElements != 1) {
```
keryell, line 71: What is the problem with 1?

steffenlarsen: These test cases try to construct an `marray` using a scalar and an `marray` of size `NumElements - 1`. If `NumElements` is 1, then the smaller `marray` would have to be of size 0. 0-sized `marray`s are not allowed by the [specification](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:marray.type) based on the following:
> The number of elements parameter, NumElements, is a positive value of the std::size_t type.

tests/marray/marray_constructor.h
steffenlarsen: You are absolutely right, but since they are unary operators on each side of the expression I think it's one of those cases where the precedence rules get hard to remember, so I personally prefer the parenthesis. 😄

tests/marray/marray_constructor.h
keryell: Oh I see, by looking at the called function.
You just implement the comment, actually. :-)

keryell (approved):
Thanks!

PR #856, opened 2024-01-17
Fix `-Wc++11-narrowing` warnings

AlexeySachkov:
Newer versions of `clang` make it error by default, which affect clang-based implementations (like DPCPP).

### DIFF
--- a/tests/common/type_list.h
+++ b/tests/common/type_list.h
@@ -162,53 +162,52 @@ struct arrow_operator_overloaded {
   }
 };
 
+// Helper structure to extract element type of vec/marray
+// Fallback for scalars, which returns unmodified T
+template <typename T>
+struct vec_marray_element_type {
+  using type = T;
+};
+
+// vec and marray specializations
+template <typename T, int N>
+struct vec_marray_element_type<sycl::vec<T, N>> {
+  using type = T;
+};
+
+template <typename T, int N>
+struct vec_marray_element_type<sycl::marray<T, N>> {
+  using type = T;
+};
+
 // Returns instance of type T
 template <typename T>
-struct init_value_helper {
-  static constexpr auto get(int x) { return static_cast<T>(x); }
-};
+inline constexpr auto get_init_value(int x) {
+  // Explicit cast is required to silence a warning about narrowing conversion
+  // in case when T is smaller than int.
+  return T{static_cast<typename vec_marray_element_type<T>::type>(x)};
+}
 
-// Returns instance of type T
-template <typename T, int N>
-struct init_value_helper<sycl::vec<T, N>> {
-  static constexpr auto get(int x) {
-    return sycl::vec<T, N>{static_cast<T>(x)};
-  }
-};
-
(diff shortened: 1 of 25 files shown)


AlexeySachkov:
Note for reviewers: the main change is in `tests/common/type_list.h`, the rest is purely mechanical and caused by the helper refactoring (function -> struct static member function).

It is possible to preserve the helper in form of a function, by doing something like:
```c++
template <typename T>
inline constexpr auto get_init_value_helper(int x) {
  return T{ static_cast<typename get_element_type<T>::type>(x) };
}
```

Where `get_element_type` is a new (or existing) type trait to get scalar element type out of `vec` and `marray`. This way the change would be less invasive. However, I do like the new name better, because with the old name one might think that function is returning some kind of helper instead of an init value. Let me know about the preference, I can rework the PR accordingly.

tests/common/type_list.h
```diff
@@ -164,31 +164,51 @@ struct arrow_operator_overloaded {
...
 // Returns instance of type T
 template <typename T>
-inline constexpr auto get_init_value_helper(int x) {
-  return x;
-}
+struct init_value_helper {
+  static constexpr auto get(int x) { return static_cast<T>(x); }
+};
+
+// Returns instance of type T
```
bader, line 171: "instance of type T" or "instance of type vec<T, N>"?
The same applies to the comment below.

tests/spec_constants/spec_constants_common.h
```diff
@@ -48,11 +48,11 @@ constexpr int default_val = 20;
 
 template <typename T, int case_num>
 constexpr sycl::specialization_id<T> spec_const(
-    user_def_types::get_init_value_helper<T>(default_val));
+    user_def_types::init_value_helper<T>::get(default_val));
```
keryell, line 51: What is the value of having `struct` instead of generic functions like
```suggestion
    user_def_types::get_init_value<T>(default_val));
```
?

bader: @keryell, I think @AlexeySachkov is seeking for discussion of this change in this comment: https://github.com/KhronosGroup/SYCL-CTS/pull/856#issuecomment-1895869936

> However, I do like the new name better, because with the old name one might think that function is returning some kind of helper instead of an init value. Let me know about the preference, I can rework the PR accordingly.

keryell: I do not like the word `helper` which does not *help* the understanding. :-)

AlexeySachkov: > What is the value of having struct instead of generic functions like

Yes, that's a good point. We can still have it as a function, but we can also rename it along the way, I will update the PR

AlexeySachkov: Refactored back to function, removed `helper` from the name in 6211bfc04faed9ce593f601af3b1e925e6ea7d33

keryell (approved):
Thanks for simplification!

PR #857, opened 2024-01-19
[NFC] Drop code related to `error_category_for`

AlexeySachkov:
That API was removed from the SYCL 2020 specification in KhronosGroup/SYCL-Docs#431

### DIFF
--- a/util/sycl_exceptions.h
+++ b/util/sycl_exceptions.h
@@ -29,8 +29,6 @@
 
 #define SYCL_CTS_SUPPORT_HAS_EXCEPTION_CODE 1
 #define SYCL_CTS_SUPPORT_HAS_EXCEPTION_CATEGORY 0
-#define SYCL_CTS_SUPPORT_HAS_ERRC_FOR 0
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks.
I am unsure this is really `[NFC]` as you are removing some functionality.

AlexeySachkov:
> Thanks. I am unsure this is really `[NFC]` as you are removing some functionality.

As you can see, I've only changed a single file which defined those "helpers" - that means that they have never been in use and that was essentially a dead code. In that case this probably satisfies `[NFC]` tag.

Note: I'm fine with the tag removal. I can do that myself, or it can be done by a maintainer if squash strategy is used when accepting the PR. Second note: without squash on accept, removal of the tag from the existing commit will require rebase and force-push from my side

bader:
We use merge strategy in this repository.

> > Thanks. I am unsure this is really `[NFC]` as you are removing some functionality.
> 
> As you can see, I've only changed a single file which defined those "helpers" - that means that they have never been in use and that was essentially a dead code. In that case this probably satisfies `[NFC]` tag.

I'm fine merging with existing PR title. @keryell?

PR #858, opened 2024-01-27
Create test plan for `sycl_ext_oneapi_kernel_compiler_spirv` extension

0x12CC:
This PR adds a test plan for the [`sycl_ext_oneapi_kernel_compiler_spirv`](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_kernel_compiler_spirv.asciidoc) extension.

### DIFF
--- a/test_plans/oneapi_kernel_compiler_spirv.asciidoc
+++ b/test_plans/oneapi_kernel_compiler_spirv.asciidoc
@@ -6,7 +6,6 @@
 This is a test plan for the API described in
 https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_kernel_compiler_spirv.asciidoc[sycl_ext_oneapi_kernel_compiler_spirv].
 
-
 == Testing scope
 
 === Device coverage
@@ -16,8 +15,8 @@ is selected on the CTS command line.
 
 === Feature test macro
 
-All of the tests should use `#ifdef SYCL_EXT_ONEAPI_KERNEL_COMPILER_SPIRV` so they can be skipped
-if feature is not supported.
+All of the tests should use `#ifdef SYCL_EXT_ONEAPI_KERNEL_COMPILER_SPIRV` so
+they can be skipped if feature is not supported.
 
 == Tests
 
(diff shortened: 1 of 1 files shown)


test_plans/oneapi_kernel_compiler_spirv.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+All of the tests should use `#ifdef SYCL_EXT_ONEAPI_KERNEL_COMPILER_SPIRV` so they can be skipped
+if feature is not supported.
+
+== Tests
+
+All of the following tests run SPIR-V kernels loaded in binary form.
+
+=== SPIR-V Source Language Test
+
+Load a simple SPIR-V kernel and run it to ensure it has the expected behavior. This kernel should take two parameters: an input pointer and an output pointer. Each of these pointers should have type *OpTypePointer*, with *CrossWorkgroup* storage class pointing to  *OpTypeInt* with width 32. For each work item, the kernel computes the following expression: `out[id] = (a * in[id]) + b`, where `a` and `b` are integer constants. The host code will create this kernel, pass it USM pointer arguments, run it, and assert that the output has the expected values. This test ensures that `create_kernel_bundle_from_source` can be used with `source_language::spirv` to obtain a kernel, set its parameters, and run it.
```
AlexeySachkov, line 28: Should we add a test to check that `ext_oneapi_has_kernel` works correctly (both positive and negative case)?

test_plans/oneapi_kernel_compiler_spirv.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+is selected on the CTS command line.
+
+=== Feature test macro
+
+All of the tests should use `#ifdef SYCL_EXT_ONEAPI_KERNEL_COMPILER_SPIRV` so they can be skipped
+if feature is not supported.
+
+== Tests
+
+All of the following tests run SPIR-V kernels loaded in binary form.
```
AlexeySachkov, line 24: This is probably more of a question to the extension spec:

It seems like it allows to obtain a (more or less) regular `kernel_bundle<executable>` that can be passed by user to `handler::use_kernel_bundle` API and define and launch a kernel using regular `parallel_for` mechanism in the SYCL source.

The question here: is it legal? If not, do we expect a certain exception to be thrown? And if so, there should be a test case for this.

test_plans/oneapi_kernel_compiler_spirv.asciidoc
```diff
@@ -0,0 +1,45 @@
...
+
+=== Parameter Tests
+
+This test checks that kernels can accept parameters for all of the SPIR-V types required by the extension. The required types are the following:
+
+- *OpTypeInt*, width 8, 16, 32, and 64.
+- *OpTypeFloat*, width 16, 32 and 64.
+
+For each type `T`, define a kernel with parameters `T`, *OpTypePointer* with *Workgroup* storage class pointing to `T`, and *OpTypePointer* with *CrossWorkgroup* storage class pointing to `T`. This kernel should compute an expression using all three parameters and store the result. The host code can then check this result to ensure that the parameter types are working.
```
AlexeySachkov, line 37: Shouldn't we keep 80-limit line width?

keryell (approved):
Thanks.

test_plans/oneapi_kernel_compiler_spirv.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+This test checks that kernels can accept *OpTypeStruct* parameters that match
+the constraints specified by the extension. Define a kernel that accepts an
+input *OpTypeStruct* and an output *OpTypePointer* with *CrossWorkgroup* storage
+class pointing to the same *OpTypeStruct*. The struct should contain
+*OpTypeInt*, *OpTypeFloat*, and inner *OpTypeStruct* members. The kernel
+computes an expression using each member from the input and stores the result in
+the corresponding member in the output. The host code then checks the output
+struct to ensure the members have the expected values.
+
+===
```
gmlueck, line 76: Delete this orphaned section heading at the end?

test_plans/oneapi_kernel_compiler_spirv.asciidoc
gmlueck: The extension spec says the following:

> The new APIs added by this extension are an expansion of the existing `kernel_bundle` capabilities. Thus, an application can create a kernel bundle from a source string and then build the bundle into "executable" bundle state. Once the application obtains a `kernel` object, it can use existing APIs from the core SYCL specification to set the value of kernel arguments and enqueue the kernel to a device.

(https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_kernel_compiler.asciidoc#overview)

Therefore, the application must use the overload of `parallel_for` or `single_task` that takes a parameter of type `kernel` if it wants to enqueue a kernel that was created from this extension.  These overloads are defined in the core SYCL specification, and the description contains the following text:

> This invocation function ignores any `kernel_bundle` that was bound to this command group handler via `handler::use_kernel_bundle()` and instead implicitly uses the kernel bundle that contains the `kernelObject`.

Therefore, it is not necessary to call `handler::use_kernel_bundle`.

test_plans/oneapi_kernel_compiler_spirv.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+For the *OpTypeFloat* parameters with width 16 or 64, the host code should use
+`sycl::aspect::fp16` and `sycl::aspect::fp64` to determine if the test kernels
+can be built and run, or if they should be skipped.
+
+=== Struct Parameter Tests
+
+This test checks that kernels can accept *OpTypeStruct* parameters that match
+the constraints specified by the extension. Define a kernel that accepts an
+input *OpTypeStruct* and an output *OpTypePointer* with *CrossWorkgroup* storage
```
gmlueck, line 69: Note that the input parameter should actually be **OpTypePointer** with **Function** storage class pointing to **OpTypeStruct**.  (This is because structures are passed by reference.)

gmlueck (approved):

PR #861, opened 2024-02-02
Fix sub-group broadcast test

AlexeySachkov:
The test made incorrect assumptions about work-items mapping to sub-groups, i.e. it expected that sub-groups are formed from subsequent work-items and thus the last element in the first sub-group of an nd-range would have its global linear id equal to sub-group size.

However, SYCL specification does not mandate any specific mapping of work-items into sub-groups and therefore assumptions made by test are incorrect.

This PR rewrites the test so that kernel stores both original and broadcasted values into corresponding buffers and the former is used as a reference for the latter and checked on host.

### DIFF
--- a/tests/group_functions/group_broadcast.h
+++ b/tests/group_functions/group_broadcast.h
@@ -131,87 +131,127 @@ void broadcast_sub_group(sycl::queue& queue) {
   sycl::range<D> work_group_range = sycl_cts::util::work_group_range<D>(queue);
 
   // array to return results
-  T res[test_matrix + 1] = {splat_init<T>(0)};
+  T origin_values[test_matrix] = {splat_init<T>(0)};
+  T broadcasted_values[test_matrix] = {splat_init<T>(0)};
   {
-    sycl::buffer<T, 1> res_sycl(res, sycl::range<1>(test_matrix + 1));
+    sycl::buffer<T, 1> origin_values_buf(origin_values,
+                                         sycl::range<1>(test_matrix));
+    sycl::buffer<T, 1> broadcasted_values_buf(broadcasted_values,
+                                              sycl::range<1>(test_matrix));
 
     queue.submit([&](sycl::handler& cgh) {
-      auto res_acc =
-          res_sycl.template get_access<sycl::access::mode::read_write>(cgh);
+      auto origin_values_acc =
+          origin_values_buf.template get_access<sycl::access::mode::read_write>(
+              cgh);
+      auto broadcasted_values_acc =
+          broadcasted_values_buf
+              .template get_access<sycl::access::mode::read_write>(cgh);
 
       sycl::nd_range<D> executionRange(work_group_range, work_group_range);
+      // Values computed in a kernel depend on global linear id. We need to make
+      // sure that there are no overflows
+      REQUIRE(executionRange.get_global_range().size() <
+              std::numeric_limits<size_t>::max() / 100);
 
       cgh.parallel_for<broadcast_sub_group_kernel<
           D, T>>(executionRange, [=](sycl::nd_item<D> item) {
         sycl::sub_group sub_group = item.get_sub_group();
 
+        // Each work-item computes a unique value
+        T value_to_broadcast(splat_init<T>(static_cast<size_t>(
+            item.get_global_linear_id() * 100 + sub_group.get_local_id())));
         T local_var(splat_init<T>(0));
 
+        // To simplify the test, we are only checking the first sub-group
         if (sub_group.get_group_id()[0] == 0) {
-          // find local id of last group item
-          sycl::id<1> last_item = sub_group.get_local_range();
-          --last_item[0];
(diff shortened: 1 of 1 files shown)


tests/group_functions/group_broadcast.h
```diff
@@ -131,87 +131,127 @@ void broadcast_sub_group(sycl::queue& queue) {
...
 
+        // To simplify the test, we are only checking the first sub-group
         if (sub_group.get_group_id()[0] == 0) {
-          // find local id of last group item
-          sycl::id<1> last_item = sub_group.get_local_range();
-          --last_item[0];
-
-          // broadcast from the first workitem
-          local_var = splat_init<T>(item.get_global_linear_id() + 1);
+          // Find local id of first, last and some third sub-group item in
```
keryell, line 167:
```suggestion
          // Find local id of first, last and some third sub-group id in
```
To be less confusing, since `sycl::item` is also a thing.
I had to look at the spec to refresh my memory on sub-groups.
It might be clearer to add a comment before `first_id` to say it is the sub-group leader or `sub_group::leader()`?

tests/group_functions/group_broadcast.h
```diff
@@ -131,87 +131,127 @@ void broadcast_sub_group(sycl::queue& queue) {
...
+          // between. Will be used to check different combinations of
+          // broadcasting and receiving work-items
+          sycl::id<1> first_id = 0;
+          sycl::id<1> mid_id = sub_group.get_local_range() / 2;
+          sycl::id<1> last_id = sub_group.get_local_range();
+          --last_id[0];
```
keryell, line 173: Just curious about this in 2 lines instead of just 1?

tests/group_functions/group_broadcast.h
```diff
@@ -131,87 +131,127 @@ void broadcast_sub_group(sycl::queue& queue) {
...
-          if (sub_group.get_local_linear_id() == 0) res_acc[1] = local_var;
+          if (sub_group.get_local_id() == last_id) {
+            // Work-item which does the broadcast stores value to broadcast to
+            // use it later as a reference
+            origin_values_acc[1] = value_to_broadcast;
+          }
+
+          broadcasted_value =
+              sycl::group_broadcast(sub_group, value_to_broadcast,
+                                    sub_group.get_local_linear_range() - 1);
```
keryell, line 205: Is it not possible to use `last_id` here?

keryell (commented):
Thanks for cleaning this recurring `sub_group`-related bug.

tests/group_functions/group_broadcast.h
AlexeySachkov: I should probably leave a comment here, because `group_broadcast` has overloads accepting both `size_t` and `sycl::it` as an index of a work-item which should be used as source.

Were are trying to test both interfaces here. I should also fix the last argument in `ASSERT_RETURN_TYPE` to match that particular overload with plain `size_t`

keryell: The more comments I read, the happier my brain is. ;-)

tests/group_functions/group_broadcast.h
AlexeySachkov: Reworded the comment in 2952e4e0fee35fe8d3d4ff9a23b5e48abd4ff311

tests/group_functions/group_broadcast.h
AlexeySachkov: I just didn't think about changing the existing code much. Shortened to a single line in 2952e4e0fee35fe8d3d4ff9a23b5e48abd4ff311

tests/group_functions/group_broadcast.h
AlexeySachkov: Applied all the mentioned changes in 2952e4e0fee35fe8d3d4ff9a23b5e48abd4ff311

AlexeySachkov:
2952e4e0fee35fe8d3d4ff9a23b5e48abd4ff311 addresses all comments, but I haven't really tried to run the test using that commit, because my machine misbehaves. That new revision seems to be almost NFC, so I don't expect that it introduces any regressions.

Note that I won't be available to react to feedback next week, but maybe @steffenlarsen will back me up in case of further comments/questions.

tests/group_functions/group_broadcast.h
```diff
@@ -164,13 +164,12 @@ void broadcast_sub_group(sycl::queue& queue) {
 
         // To simplify the test, we are only checking the first sub-group
         if (sub_group.get_group_id()[0] == 0) {
-          // Find local id of first, last and some third sub-group item in
-          // between. Will be used to check different combinations of
+          // Find local id of a leader, last and some other work-item in the
```
keryell, line 167:
```suggestion
          // Find local id of the leader, last and some other work-item in the
```
"the leader" in the sense of the `.leader()`?

keryell (commented):
Great! Quite clearer.

keryell (approved):
Thanks!

PR #865, opened 2024-02-05
Add test plan for sycl_ext_oneapi_composite_device

AlexeySachkov:


test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,128 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl_ext_oneapi_composite_device
+
+THis is a test plan for the API described in the
```
keryell, line 6:
```suggestion
This is a test plan for the API described in the 
```

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,128 @@
...
+together.
+
+=== Enumerating composite devices
+
+==== Through the new `get_composite_devices()` API
+
+This test should check that freestanding function `get_composite_devices()`:
+- has the correct signature
+- does not throw an exception
+- returns the same result on subsequent calls
```
keryell, line 37: Need probably some spacing to be interpreted as AsciiDoc item list?

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: Applied in f01911d3f691d211d876b02fd6afb7c876d33851

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: Indeed, fixed in f01911d3f691d211d876b02fd6afb7c876d33851

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,131 @@
...
+document and describe tests with narrower scope, which are aimed to check
+specific functionality. There are sections below which describe bigger test
+cases which are intended to check how different elements of the extension work
+together.
+
+=== Enumerating composite devices
+
+==== Through the new `get_composite_devices()` API
+
+This test should check that freestanding function `get_composite_devices()`:
```
AlexeySachkov, line 34:
```suggestion
This test should check that freestanding function `get_composite_devices()`:

```

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,131 @@
...
+`platform::ext_oneapi_get_composite_devices()` method:
+
+- has the correct signature
+- does not throw an exception
+- returns the same result on subsequent calls
+
+==== Composite devices are not considered as root devices
+
+The test should ensure that no composite devices are returned through the
+`platform::get_device()` method for a platform the selected device belongs to.
```
AlexeySachkov, line 52:
```suggestion
`platform::get_devices()` method for a platform the selected device belongs to.
```

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,132 @@
...
+
+The test should check that a query for the information descriptor has correct
+return type.
+
+The test should also check the value returned for the descriptor:
+
+- if the selected device is a component device, then a synchronous `exception`
+  with `errc::invalid` error code is expected
+- otherwise, returned device should have the `ext_oneapi_is_composite` aspect
+
```
gmlueck, line 80: I think your test for `info::device::component_devices` won't work because the CTS "selected device" will always be some root device, and root devices are never composite devices.  I'd suggest combining the tests for `info::device::component_devices` and `info::device::composite_device` like this:

* If the selected device is not a composite device, `info::device::component_devices` must return an empty list.  (The expectation is that the selected device will never be a composite device.)
* If the selected device is not a component device, `info::device::composite_device` must raise a synchronous `exception` with the `errc::invalid` error code.
* Otherwise:
  * The returned device must have the `ext_oneapi_is_composite` aspect.
  * Calling `info::device::component_devices` must return a list of at least 2 devices.
    * One of these devices must be the selected device.
    * All devices in the list must have the `ext_oneapi_is_component` aspect.

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,132 @@
...
+=== New device aspects
+
+==== `ext_oneapi_is_composite`
+
+The test should check that if the selected device has the aspect, then the
+device should be present in a list returned by both `get_composite_devices()`
+and `platform::ext_oneapi_get_composite_devices()`.
```
gmlueck, line 87: As I mention above, I think the CTS "selected device" will never have this aspect, so this test won't do anything useful.  The test I propose above also checks for this aspect, so that provides some testing for the aspect.

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,132 @@
...
+
+==== `ext_oneapi_is_component`
+
+The test should check that if the selected device has the aspect, then the
+device should *not* be present in a list returned by either
+`get_composite_devices()` or `platform::ext_oneapi_get_composite_devices()`.
```
gmlueck, line 93: It would be more correct to do this test:

> If the selected device does not have the `ext_oneapi_is_composite` aspect, then the device should *not* be present in the list returned by either `get_composite_devices()` or `platform::ext_oneapi_get_composite_devices()`.

(Remember, there could be devices that are neither composite nor component devices.)

However, you already perform this test above under the section "Composite devices are not considered as root devices".  Therefore, I think the test described here is really just a duplicate.

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: What do you think about just extending each test to cover not only the CTS selected device, but also a composite device which is queried from the same platform?

I'm afraid that if we only test the CTS selected device, then we simply won't have coverage for "positive" case (i.e. we don't check `component_devices` query on a composite device at all)

gmlueck: > What do you think about just extending each test to cover not only the CTS selected device, but also a composite device which is queried from the same platform?

What do you mean by "each test"?  Do you mean that we would make some fundamental change to the CTS such that every CTS test also tries to run on a composite device?  Or, are you just talking about the tests in this test plan?

AlexeySachkov: > > What do you think about just extending each test to cover not only the CTS selected device, but also a composite device which is queried from the same platform?
> 
> What do you mean by "each test"? Do you mean that we would make some fundamental change to the CTS such that every CTS test also tries to run on a composite device? Or, are you just talking about the tests in this test plan?

I mean only this test plan and moreover, only a few test cases (for info queries and aspects), which depend on the device they get and therefore would have more coverage if we launched them on both kinds of devices

AlexeySachkov: > Isn't this similar to what I proposed above:

It is, but I think that selecting a wide range of devices first and then performing the same set of checks on every device from that range is more orderly. For example, in the algorithm described by you we never query `composite_device` info on a composite device

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,132 @@
...
+
+The test should check that a query for the information descriptor has correct
+return type.
+
+The test should also check the value returned for the descriptor:
+
+- if the selected device is a component device, an empty vector is expected
+- otherwise, returned vector is expected to be non-empty and each its element
+  should be additionally checked:
+  - each component device should have the `ext_oneapi_is_component` aspect
```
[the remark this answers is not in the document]
AlexeySachkov: Note to self: pay more attention to how things are rendered. Should be fixed by 15431034242fd4163bdbd9dc81ac98691597e6e9 and 01c2d702fc0513305ec7fd1d57071d9e15651df1

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,132 @@
...
+    `info::device::composite_device` query
+
+==== `info::device::composite_device`
+
+The test should check that a query for the information descriptor has correct
+return type.
+
+The test should also check the value returned for the descriptor:
+
+- if the selected device is a component device, then a synchronous `exception`
```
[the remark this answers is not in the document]
AlexeySachkov: Fixed in 15431034242fd4163bdbd9dc81ac98691597e6e9

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: > Your proposal is similar, except you want to iterate over all composite devices (returned from get_composite_devices), and then run the composite device tests on those devices.

My implementation uses `platform::ext_oneapi_get_composite_devices()` to at least limit the list to the tested platform. However, I agree that until we have changed the device selection/testing strategy globally, let's stick to the current philosophy. db80d6d44b4ccdaa24bf2a52c1fee0274935e0b0 applies your device selection logic, but leaves the general structure in place: all tests are executed for potentially more than one device, but list of the checks is consistent for both and doesn't depend (that much) on which device we are testing, but mostly on what we are checking for

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: Should be resolved by 15431034242fd4163bdbd9dc81ac98691597e6e9 and db80d6d44b4ccdaa24bf2a52c1fee0274935e0b0

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,132 @@
...
+
+==== `info::device::component_devices`
+
+The test should check that a query for the information descriptor has correct
+return type.
+
+The test should also check the value returned for the descriptor:
+
+- if the selected device is a component device, an empty vector is expected
+- otherwise, returned vector is expected to be non-empty and each its element
```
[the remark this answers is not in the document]
AlexeySachkov: This is not applicable since 15431034242fd4163bdbd9dc81ac98691597e6e9

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: 9e548b8938ca376b49e847e3d23de872befbc280 rewrites tests for aspect in accordance of how they are implemented (#876):
- there are basic tests to check aspects consistency which use `get_composite_devices()` and `platform::ext_oneapi_get_composite_devices()` results
- there is a test specifically for component devices to check how their partitioning works

Rest of the things related to the aspects and their interaction with other extension functionality is covered by other tests, I think

test_plans/sycl_ext_oneapi_composite_device.asciidoc
```diff
@@ -0,0 +1,138 @@
...
+
+=== Impact on "descendent device"
+
+The test should first obtain both a component and a composite device and it is
+skipped if that is not possible using the selected device.
+
+Having those two, the test should create a `context` using the _composite_
+device and then attempt to create a `queue` by passing the created context in
+there *and* the _component_ device. The test checks that `queue` was created
+successfully, i.e. no exceptions were thrown.
```
gmlueck, line 113: It would be nice to extend this test so that it submits a kernel that uses USM.  This would be a better test of "descended device".  After the queue is created as you describe above:

* Allocate USM memory (it could be either "device" or "shared" USM) using the context that you created.
* Submit a kernel to your queue that writes into that USM memory.
* Verify that the USM memory has the expected values after the kernel completes.

I realize this expands the scope of the test, so maybe it belongs under "More complex test cases".

AlexeySachkov: I actually have it implemented [here](https://github.com/KhronosGroup/SYCL-CTS/pull/876/files#diff-d1fabf9f3122b35406b465f939761c182b9a153dcc13ffcab9e7a7b862f40d65R81) as part of "More complex test cases". I guess I can just expand that section to clarify that we specifically want to test USM as a shared resource

gmlueck: It looks like that test creates a "shared context" -- a context that contains both the composite and component devices.

In order to test that "descendent devices" work properly, the context should contain only the composite device.

maarquitos14 (approved):

maarquitos14:
I know https://github.com/gmlueck/llvm/blob/1f1712ed025c76a30ca251c97175685a59bf58bb/sycl/doc/extensions/proposed/sycl_ext_oneapi_composite_device.asciidoc#impact-to-the-oneapi_device_selector section just impacts Intel's implementation, but I wonder if similar mechanisms might exist for other implementations, and if there's a generic way to test this for any given SYCL implementation. To be specific: composite devices should only be returned if all their components are in the list of root devices returned from `device::get_devices()`, so if there's a way to only select some of the components --like Intel's implementation has with `ONEAPI_DEVICE_SELECTOR`--, then CTS should maybe test it.

test_plans/sycl_ext_oneapi_composite_device.asciidoc
AlexeySachkov: Added the test case under "more complex test cases" in 7b4060b264026aaa7d3e559b4e82da8b9f5bad46

AlexeySachkov:
> I know https://github.com/gmlueck/llvm/blob/1f1712ed025c76a30ca251c97175685a59bf58bb/sycl/doc/extensions/proposed/sycl_ext_oneapi_composite_device.asciidoc#impact-to-the-oneapi_device_selector section just impacts Intel's implementation, but I wonder if similar mechanisms might exist for other implementations, and if there's a generic way to test this for any given SYCL implementation. To be specific: composite devices should only be returned if all their components are in the list of root devices returned from `device::get_devices()`, so if there's a way to only select some of the components --like Intel's implementation has with `ONEAPI_DEVICE_SELECTOR`--, then CTS should maybe test it.

`ONEAPI_DEVICE_SELECTOR` isn't really a part of the language, so I would say that interaction of the extension with it should be tested by the implementation, and not by CTS

gmlueck (approved):

keryell (approved):
Interesting.

PR #866, opened 2024-02-06
Add test plan for oneapi_non_uniform_groups extension

steffenlarsen:
This commit adds a test plan for the [oneapi_non_uniform_groups](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_non_uniform_groups.asciidoc) oneAPI extension.

### DIFF
--- a/test_plans/non_uniform_groups.asciidoc
+++ b/test_plans/non_uniform_groups.asciidoc
@@ -119,8 +119,8 @@ Check that `leader()` return type is `bool` and return value is equal to
 Let `N` be some power-of-two value greater than 1 that is expected to be a
 divisor of the sub-group size of most devices. The `get_fixed_size_group` is
 called with the `sub_group` of the invocation and `N` as the partition size.
-Let `M` be the size of the sub-group the given `fixed_size_group` was created
-from.
+Let `M` be the result of `get_local_range()` on the sub-group the given
+`fixed_size_group` was created from.
 
 ==== Group traits
 
@@ -193,7 +193,7 @@ Check that `leader()` return type is `bool` and return value is equal to
 
 The `get_tangle_group` is called with the `sub_group` of the invocation. This
 will only be called by the first `N` items of the sub-group, where `N` is
-strictly less than the size of the sub-group.
+strictly less than the result of `get_local_range()` on the sub-group.
 
 ==== Group traits
 
@@ -264,7 +264,7 @@ Check that `leader()` return type is `bool` and return value is equal to
 === The `opportunistic_group` class API
 
 The `get_opportunistic_group` is called by all work items.
-Let `M` be the size of the sub-group of the invocation.
+Let `M` be the result of `get_local_range()` on the sub-group of the invocation.
 
 ==== Group traits
 
--- a/test_plans/non_uniform_groups.asciidoc
+++ b/test_plans/non_uniform_groups.asciidoc
@@ -92,22 +92,22 @@ equal to `N1` if the predicate was `true` or equal to `N2` if the predicate was
 ==== get_group_linear_id
 
 Check that `get_group_linear_id()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
+value is equal to `get_group_id()` converted to `linear_id_type`.
 
 ==== get_local_linear_id
 
 Check that `get_local_linear_id()` return type is `linear_id_type` and the
-return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
+return value is equal to `get_local_id()` converted to `linear_id_type`.
 
 ==== get_group_linear_range
 
 Check that `get_group_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
+value is equal to `get_group_range()` converted to `linear_id_type`.
 
 ==== get_local_linear_range
 
 Check that `get_local_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
+value is equal to `get_local_range()` converted to `linear_id_type`.
 
 ==== leader
 
@@ -166,22 +166,23 @@ equal to `N`.
 ==== get_group_linear_id
 
 Check that `get_group_linear_id()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
+value is equal to `get_group_id()` converted to `linear_id_type`.
 
 ==== get_local_linear_id
 
 Check that `get_local_linear_id()` return type is `linear_id_type` and the
-return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
+return value is equal to `get_local_id()` converted to
+`linear_id_type`.
 
 ==== get_group_linear_range
 
 Check that `get_group_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
+value is equal to `get_group_range()` converted to `linear_id_type`.
 
 ==== get_local_linear_range
 
 Check that `get_local_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
+value is equal to `get_local_range()` converted to `linear_id_type`.
 
 ==== leader
 
@@ -238,22 +239,22 @@ equal to `N`.
 ==== get_group_linear_id
 
 Check that `get_group_linear_id()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
+value is equal to `get_group_id()` converted to `linear_id_type`.
 
 ==== get_local_linear_id
 
 Check that `get_local_linear_id()` return type is `linear_id_type` and the
-return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
+return value is equal to `get_local_id()` converted to `linear_id_type`.
 
 ==== get_group_linear_range
 
 Check that `get_group_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
+value is equal to `get_group_range()` converted to `linear_id_type`.
 
 ==== get_local_linear_range
 
 Check that `get_local_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
+value is equal to `get_local_range()` converted to `linear_id_type`.
 
 ==== leader
 
@@ -309,22 +310,22 @@ less than or equal to `M`.
 ==== get_group_linear_id
 
 Check that `get_group_linear_id()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
+value is equal to `get_group_id()` converted to `linear_id_type`.
 
 ==== get_local_linear_id
 
 Check that `get_local_linear_id()` return type is `linear_id_type` and the
-return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
+return value is equal to `get_local_id()` converted to `linear_id_type`.
 
 ==== get_group_linear_range
 
 Check that `get_group_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
+value is equal to `get_group_range()` converted to `linear_id_type`.
 
 ==== get_local_linear_range
 
 Check that `get_local_linear_range()` return type is `linear_id_type` and return
-value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
+value is equal to `get_local_range()` converted to `linear_id_type`.
 
 ==== leader
 
--- a/test_plans/non_uniform_groups.asciidoc
+++ b/test_plans/non_uniform_groups.asciidoc
@@ -193,7 +193,7 @@ Check that `leader()` return type is `bool` and return value is equal to
 
 The `get_tangle_group` is called with the `sub_group` of the invocation. 
 Let `M` be the result of `get_local_range()` on this `sub_group` and let `N` be
-some value strictly less than `M`. `get_tangle_group` is called it two split
+some value strictly less than `M`. `get_tangle_group` is called in two split
 control-flows in an if-else-statement, the if-branch with the first `N` items of
 the sub-group and the else branch with the rest.
 
--- a/test_plans/non_uniform_groups.asciidoc
+++ b/test_plans/non_uniform_groups.asciidoc
@@ -367,5 +367,32 @@ The group algorithms
 
 for `ballot_group`, `fixed_size_group`, `tangle_group`
 and `opportunistic_group` are tested similar to how they are currently tested
-with `group` and `sub_group` in the core CTS. The groups are constructed in the
-same way as for the API testing described above.
+with `group` and `sub_group` in the core CTS.
+
+The groups are constructed as follows:
+
+* `get_ballot_group` is called with a predicate that is `true` for the first `N`
+  work-items in the sub-group.
+* `get_ballot_group` is called with a predicate that is `true` for work-items
+  with odd `sg.get_local_linear_id()` values, where `sg` is the sub-group.
+* `get_ballot_group` is called with a predicate that is `true` for all
+  work-items in the sub-group.
+* `get_ballot_group` is called with a predicate that is `false` for all
+  work-items in the sub-group.
+* `get_fixed_size_group` is called with a partition-size of 1.
+* `get_fixed_size_group` is called with a partition-size of 2, if 2 is greater
+  than or equal to the smallest supported sub-group size on the device.
+* `get_fixed_size_group` is called with a partition-size of 4, if 4 is greater
+  than or equal to the smallest supported sub-group size on the device.
+* `get_fixed_size_group` is called with a partition-size of 8, if 8 is greater
+  than or equal to the smallest supported sub-group size on the device.
+* `get_tangle_group` is called in a branched control-flow with the first `N`
+  work-items in the sub-group.
+* `get_tangle_group` is called in a branched control-flow with work-items with
+  odd `sg.get_local_linear_id()` values, where `sg` is the sub-group.
+* `get_tangle_group` is called by all items in the sub-group.
+* `get_opportunistic_group` is called in a branched control-flow with the first
+  `N` work-items in the sub-group.
+* `get_opportunistic_group` is called in a branched control-flow with work-items
+  with odd `sg.get_local_linear_id()` values, where `sg` is the sub-group.
+* `get_opportunistic_group` is called by all items in the sub-group.
--- a/test_plans/non_uniform_groups.asciidoc
+++ b/test_plans/non_uniform_groups.asciidoc
@@ -4,7 +4,7 @@
 = Test plan for sycl_ext_oneapi_non_uniform_groups
 
 This is a test plan for the APIs described in
-https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_non_uniform_groups.asciidoc[sycl_ext_oneapi_non_uniform_groups].
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_non_uniform_groups.asciidoc[sycl_ext_oneapi_non_uniform_groups].
 
 
 == Testing scope
--- a/test_plans/non_uniform_groups.asciidoc
+++ b/test_plans/non_uniform_groups.asciidoc
@@ -1,0 +1,370 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl_ext_oneapi_non_uniform_groups
+
+This is a test plan for the APIs described in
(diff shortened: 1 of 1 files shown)


test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
+:sectnums:
+:xrefstyle: short
+
+= Test plan for sycl_ext_oneapi_non_uniform_groups
+
+This is a test plan for the APIs described in
+https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_non_uniform_groups.asciidoc[sycl_ext_oneapi_non_uniform_groups].
```
keryell, line 7: URL not working.

steffenlarsen: My bad! It should work now.

gmlueck:
@Pennycook could I delegate my review role to you for this test plan?  I think you'd have better insight since you were the original author of this extension spec.  (I tried adding you in the "Reviewers" section, but for some reason GitHub doesn't find you there.)

Pennycook:
> @Pennycook could I delegate my review role to you for this test plan? I think you'd have better insight since you were the original author of this extension spec. (I tried adding you in the "Reviewers" section, but for some reason GitHub doesn't find you there.)

I think I can't be officially added as a reviewer because I'm not associated with the SYCL-CTS repository.  Regardless, I'll take a look!

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+* `is_fixed_topology_group<root_group>::value` is `true`.
+* `is_fixed_topology_group_v<root_group>` is `true`.
+
+=== The `ballot_group` class API
+
+The `get_ballot_group` is called with the `sub_group` of the invocation and a
+predicate splitting the work-items of the sub-group into uneven groups. Let `N1`
+be the size of the group created with `true` predicate and let `N2` be the size
+of the group created with `false` predicate.
```
Pennycook, line 47: Can/should we say something here about which values of `N` we should test?  I understand that we can't be exhaustive here, but I'm thinking it might be a good idea to check:
- When the predicate is true on all work-items
- When the predicate is false on all work-items
- When the predicate deliberately doesn't result in contiguous groups

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+Check that `leader()` return type is `bool` and return value is equal to
+`get_local_id() == 0`.
+
+=== The `fixed_size_group` class API
+
+Let `N` be some power-of-two value greater than 1 that is expected to be a
+divisor of the sub-group size of most devices. The `get_fixed_size_group` is
+called with the `sub_group` of the invocation and `N` as the partition size.
+Let `M` be the size of the sub-group the given `fixed_size_group` was created
+from.
```
Pennycook, line 123: I'm not sure where to document this, but should we only test the cases where `M` >= `N`?

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== leader
+
+Check that `leader()` return type is `bool` and return value is equal to
+`get_local_id() == 0`.
+
+=== The `fixed_size_group` class API
+
+Let `N` be some power-of-two value greater than 1 that is expected to be a
+divisor of the sub-group size of most devices. The `get_fixed_size_group` is
```
Pennycook, line 120: Similar to the question above: I think we should make sure we check:
- When the size is 1
- When the size is equal to the sub-group size

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+* `id_type` is same as `id<1>`.
+* `range_type` is same as `range<1>`.
+* `linear_id_type` is same as `uint32_t`.
+* `dimensions` is 1.
+* `fence_scope` is equal to `sub_group::fence_scope`.
+
+==== get_group_id
+
+Check that `get_group_id()` return type is `id<1>` and return value is less than
+`M/N`.
```
Pennycook, line 149: I'm not sure about `M/N` here (or for `get_group_range()`).  Is `M` here intended as the result of `sub_group::get_local_range()` or `sub_group::get_max_local_range()`?

I think we want the former -- if you agree, a simple fix would be to clarify what we mean by `M` when it's introduced above.

test_plans/non_uniform_groups.asciidoc
Pennycook, line 47: I also think you need to add something like:
```suggestion
of the group created with `false` predicate. Let `ballot` be the result of `get_ballot_group`.
```
...or find some way to remove `ballot.` from the descriptions below.

test_plans/non_uniform_groups.asciidoc
Pennycook, line 123:
```suggestion
from. Let `fixed` be the result of `get_fixed_size_group`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_range
+
+Check that `get_local_range()` return type is `range_type` and return value is
+equal to `N`.
+
+==== get_group_linear_id
+
+Check that `get_group_linear_id()` return type is `linear_id_type` and return
+value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
```
Pennycook, line 169:
```suggestion
value is equal to `fixed.get_group_id()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_linear_id
+
+Check that `get_local_linear_id()` return type is `linear_id_type` and the
+return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
```
Pennycook, line 174:
```suggestion
return value is equal to `fixed.get_local_id()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_group_linear_range
+
+Check that `get_group_linear_range()` return type is `linear_id_type` and return
+value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
```
Pennycook, line 179:
```suggestion
value is equal to `fixed.get_group_range()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_linear_range
+
+Check that `get_local_linear_range()` return type is `linear_id_type` and return
+value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
```
Pennycook, line 184:
```suggestion
value is equal to `fixed.get_local_range()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+==== leader
+
+Check that `leader()` return type is `bool` and return value is equal to
+`get_local_id() == 0`.
+
+=== The `tangle_group` class API
+
+The `get_tangle_group` is called with the `sub_group` of the invocation. This
+will only be called by the first `N` items of the sub-group, where `N` is
+strictly less than the size of the sub-group.
```
Pennycook, line 195:
```suggestion
The `get_tangle_group` is called with the `sub_group` of the invocation. This
will only be called by `N` items of the sub-group, where `N` is
strictly less than the size of the sub-group.
```

I don't think we can guarantee that it will be the _first_ `N` -- it'll be whichever work-items happen to be active in this control flow.

Pennycook, line 195:
```suggestion
The `get_tangle_group` is called with the `sub_group` of the invocation. This
will only be called by the first `N` items of the sub-group, where `N` is
strictly less than the size of the sub-group. Let `tangle` be the result of
`get_tangle_group`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_range
+
+Check that `get_local_range()` return type is `range_type` and return value is
+equal to `N`.
+
+==== get_group_linear_id
+
+Check that `get_group_linear_id()` return type is `linear_id_type` and return
+value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
```
Pennycook, line 241:
```suggestion
value is equal to `tangle.get_group_id()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_linear_id
+
+Check that `get_local_linear_id()` return type is `linear_id_type` and the
+return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
```
Pennycook, line 246:
```suggestion
return value is equal to `tangle.get_local_id()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_group_linear_range
+
+Check that `get_group_linear_range()` return type is `linear_id_type` and return
+value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
```
Pennycook, line 251:
```suggestion
value is equal to `tangle.get_group_range()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_linear_range
+
+Check that `get_local_linear_range()` return type is `linear_id_type` and return
+value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
```
Pennycook, line 256:
```suggestion
value is equal to `tangle.get_local_range()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
Pennycook, line 195: I think it's important for the CTS to deliberately check usage of a tangle in two sides of a branch:
```c++
if (condition) {
  auto tangle = get_tangle_group(sg);
  // checks
}
else {
  auto tangle = get_tangle_group(sg);
  // identical checks
}
```

Compilers are incredibly likely to see code like this and try to be clever by removing the branch, but that ignores the desired `tangle_group` semantics.  If we don't have a test like this, I don't think we can be confident that an implementation actually understands tangles.

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== leader
+
+Check that `leader()` return type is `bool` and return value is equal to
+`get_local_id() == 0`.
+
+=== The `opportunistic_group` class API
+
+The `get_opportunistic_group` is called by all work items.
+Let `M` be the size of the sub-group of the invocation.
```
Pennycook, line 266: As above: we should clarify if `M` is `get_local_range()` or `get_max_local_range()`.

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
+
+==== leader
+
+Check that `leader()` return type is `bool` and return value is equal to
+`get_local_id() == 0`.
+
+=== The `opportunistic_group` class API
+
+The `get_opportunistic_group` is called by all work items.
```
Pennycook, line 265: An implementation could pass this test by implementing `opportunistic_group` as a wrapper around `sub_group`.  I think we need to make sure we deliberately test `get_opportunistic_group` inside of some divergent control flow, similar to tangles.

test_plans/non_uniform_groups.asciidoc
Pennycook, line 266:
```suggestion
Let `M` be the size of the sub-group of the invocation. Let `opportunistic` be the result
of `get_opportunistic_group`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_range
+
+Check that `get_local_range()` return type is `range_type` and return value is
+less than or equal to `M`.
+
+==== get_group_linear_id
+
+Check that `get_group_linear_id()` return type is `linear_id_type` and return
+value is equal to `ballot.get_group_id()` converted to `linear_id_type`.
```
Pennycook, line 312:
```suggestion
value is equal to `opportunistic.get_group_id()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_linear_id
+
+Check that `get_local_linear_id()` return type is `linear_id_type` and the
+return value is equal to `ballot.get_local_id()` converted to `linear_id_type`.
```
Pennycook, line 317:
```suggestion
return value is equal to `opportunistic.get_local_id()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_group_linear_range
+
+Check that `get_group_linear_range()` return type is `linear_id_type` and return
+value is equal to `ballot.get_group_range()` converted to `linear_id_type`.
```
Pennycook, line 322:
```suggestion
value is equal to `opportunistic.get_group_range()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+==== get_local_linear_range
+
+Check that `get_local_linear_range()` return type is `linear_id_type` and return
+value is equal to `ballot.get_local_range()` converted to `linear_id_type`.
```
Pennycook, line 327:
```suggestion
value is equal to `opportunistic.get_local_range()` converted to `linear_id_type`.
```

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,370 @@
...
+
+The group functions
+
+* `group_barrier`
+* `group_broadcast`
+
+for `ballot_group`, `fixed_size_group`, `tangle_group`
+and `opportunistic_group` are tested similar to how they are currently tested
+with `group` and `sub_group` in the core CTS. The groups are constructed in the
+same way as for the API testing described above.
```
Pennycook, line 344: This may be a good place to call out some of the specific tests or divergence requirements that I mentioned above, but I'll leave it to you to decide.

Pennycook (changes requested):

test_plans/non_uniform_groups.asciidoc
steffenlarsen: I agree, it should be `sub_group::get_local_range()`. I have updated it in all the places it should be clarified.

test_plans/non_uniform_groups.asciidoc
steffenlarsen: I think in general we can simply refer to them by the member function names without an object attached, so I have just removed the object from the descriptions.

test_plans/non_uniform_groups.asciidoc
steffenlarsen: First here refers to the way the control-flow is split, i.e. we pick the ones with `sub_group.get_local_linear_id() < N`. I suppose I can clarify that here, but do we need to be that specific in the test plan? I'm fine with either your suggestion or being very specific.

steffenlarsen: Currently my WIP tests use something like:

```
parallel_for(... [&](sycl::nd_item<1> it) {
  if (!condition)
    return;
  auto tangle = get_tangle_group(sg);
  // checks
});
```

Do you think it would be fine to separate them, i.e. something like:

```
parallel_for(... [&](sycl::nd_item<1> it) {
  if (!condition)
    return;
  auto tangle = get_tangle_group(sg);
  // checks
});

...

parallel_for(... [&](sycl::nd_item<1> it) {
  if (condition)
    return;
  auto tangle = get_tangle_group(sg);
  // identical checks
});
```

test_plans/non_uniform_groups.asciidoc
steffenlarsen: `size == 1` is fine, but I worry about matching the sub-group size exactly. We are already making some pretty heavy assumptions by picking some `N > 1` where `N` is both a divisor and power-of-two. I suspect `2`, `4` and `8` would be safe picks, but AFAIK nothing stops an implementation from only allowing sub-group size 1. What we can do is test for both `1`, `4`, and `8` and do some checks to make sure they won't hit an unsupported sub-group size.

However, if we were to check with a set `sub_group` size I worry that we have to make a finite list of sub-group sizes, as they have to be specified using `[[sycl::reqd_sub_group_size(...)]]` and in that case we might leave out some hardware that could at least test `1` and probably also `4` and `8`.

Pennycook: It's a shame we don't have named sub-groups yet -- then we could just test the device's preferred sub-group size.

Your suggestion to test `1`, `4` and `8` sounds good to me.

test_plans/non_uniform_groups.asciidoc
Pennycook: No, I don't think this is enough.  The behavior we want to test isn't whether tangles can exist for both `condition` and `!condition`, but rather that they can exist _simultaneously_.  If the tests are split out, there's no risk of a compiler optimizing too aggressively.

It might be clearer if I complete my example.  The concern is that a compiler may see this:
```c++
if (condition) {
  auto tangle = get_tangle_group(sg);
  // checks
}
else {
  auto tangle = get_tangle_group(sg);
  // identical checks
}
```

and replace it with this:
```c++
if (condition || !condition) {
  auto tangle = get_tangle_group(sg);
  // checks
}
```

That transformation is legal for most code, but is illegal for code containing tangle groups.

test_plans/non_uniform_groups.asciidoc
steffenlarsen: I have added a list with the specific group setups here.

test_plans/non_uniform_groups.asciidoc
steffenlarsen: Should be addressed by the list of group-constructions below.

test_plans/non_uniform_groups.asciidoc
steffenlarsen: I am of two minds regarding this. On one hand, I agree that an optimizer might elect to make that optimization, but on the other hand I don't think the CTS should make tests based on potential optimizations. An implementation could simply turn off optimizations in the CTS and pass and there could be a plethora of other patterns that could also be used to cause mayhem here, so it seems a little contrived.

Pennycook: Oh, I see.  I think this is fine, then -- I'm just not used to reading the test plans.

steffenlarsen:
Draft PR with the tests: https://github.com/KhronosGroup/SYCL-CTS/pull/875

test_plans/non_uniform_groups.asciidoc
steffenlarsen: I see what you mean. Would it be enough to do it just for the API testing? I.e. all the algorithm tests can still just use one control flow? I would assume in general having the two paths in one test should be good enough to weed out the worst cases of this kind of optimizations.

Pennycook: Yes, I think that's sufficient.  That will give us some confidence that the groups are being constructed correctly in weird corner cases.  We're already not going to test all possible code they can put in those branches, so I'm okay if we skip the algorithms too!

steffenlarsen: Alright! I have tried specifying this in the test plan.

test_plans/non_uniform_groups.asciidoc
```diff
@@ -190,9 +191,11 @@ Check that `leader()` return type is `bool` and return value is equal to
 
 === The `tangle_group` class API
 
-The `get_tangle_group` is called with the `sub_group` of the invocation. This
-will only be called by the first `N` items of the sub-group, where `N` is
-strictly less than the size of the sub-group.
+The `get_tangle_group` is called with the `sub_group` of the invocation. 
+Let `M` be the result of `get_local_range()` on this `sub_group` and let `N` be
+some value strictly less than `M`. `get_tangle_group` is called it two split
```
Pennycook, line 196:
```suggestion
some value strictly less than `M`. `get_tangle_group` is called in two split
```

Pennycook (approved):
Typo aside, LGTM!

test_plans/non_uniform_groups.asciidoc
```diff
@@ -0,0 +1,400 @@
...
+value is equal to `get_local_range()` converted to `linear_id_type`.
+
+==== leader
+
+Check that `leader()` return type is `bool` and return value is equal to
+`get_local_id() == 0`.
+
+=== The `fixed_size_group` class API
+
+Let `N` be some power-of-two value greater than 1 that is expected to be a
```
keryell, line 119: Are you excluding the case of 1. That seems interesting to test corner cases, otherwise.

keryell: Oh I thought you were running with several N and in that case testing also the boundaries would have made sense.

bader:
@keryell, could you please approve the PR if all your concerns are addressed?

keryell (approved):
That looks interesting!

PR #869, opened 2024-02-17
Update nd_item constructors and equality tests.

bader:
This change allows stateless nd_item implementation as discussed in https://github.com/KhronosGroup/SYCL-Docs/issues/532. The list of changes includes removal of checks which compare nd_team objects shared between work-items or on the host from tests/nd_item/nd_item_equality.cpp and removal of checks invoking "device-only" methods on the host from tests/nd_item/nd_item_constructors.cpp.

Now tests check equality only on device within single work-item.


keryell (approved):
Thanks!

PR #870, opened 2024-02-17
Various smaller bugfixes in tests

fknorr:
This PR fixes a bunch of discrepancies with the SYCL spec and minor bugs and ambiguities that surfaced when we attempted to build the CTS against [SimSYCL](https://github.com/celerity/SimSYCL).

Some of these issues might not be visible when building against a SYCL implementation (such as DPC++) that implements non-standard type conversions, member functions or other behavior.

In individual commits:
* `test_accessor`: `get_multi_ptr()` is only defined for device- and local accessors
* `test_accessor_legacy`: legacy accessors do not have member function `size()`
* `test_atomic_ref`: Typo in function name `memory_scope_is_suppoted`
* `test_atomic_ref_stress`: `id<1> == int` is ambiguous
* `test_device_info`: `info::device::image_max_array_size` is not part of the Spec
* `test_device_info`: `info::device::opencl_c_version` is not part of the Spec
* `test_device_selector`: `sycl::runtime_error` is not part of the Spec
* `test_exceptions`: `sycl::exception` is not default-constructible
* ~~`test_hierarchical`: `sycl::id` is not convertible to `sycl::range`~~ (adressed by rebase)
* `test_id`: `sycl::id<1> - 1` is ambiguous
* `test_kernel`: remove `template` from non-dependent expression
* `test_marray`: `marray<bool>` does not have `operator++/--`
* `test_namespace`: Avoid multiple definitions of kernel names
* `test_scalars`: Regex typo in CMakeLists.txt
* `test_vector_load_store`: `sycl::vec` cannot load/store from accessor or load from non-const `multi_ptr`
* `test_vector_operators`: `vec<bool>` does not provide `operator++/--`
* `util`: Call to `erase_if` is ambiguous with C++20


steffenlarsen:
Thank you for going through all these bugs and addressing them! It makes me wonder if some of these should be part of the spec, simply for usability.
@gmlueck @AerialMantis @keryell @illuhad @tomdeakin - Tagging for awareness.

tests/accessor/generic_accessor_api_common.h
```diff
@@ -58,8 +58,23 @@ void test_accessor_methods(const AccT &accessor,
 #endif
 }
 
+template<typename Accessor>
+extern const sycl::target accessor_target_v;
```
keryell, line 62: Why this `extern` definition?

tests/accessor_legacy/accessor_api_local_common.h
```diff
@@ -69,17 +69,6 @@ class check_local_accessor_api_methods {
...
-          /** check size() method
-          */
-          auto accessorCount = acc.size();
-          check_return_type<size_t>(log, accessorCount, "size()");
-          const auto expectedCount = ((dims == 0) ? 1 : count);
-          if (accessorCount != expectedCount) {
-            fail_for_accessor<T, dims, mode, target>(log, typeName,
-                "accessor does not return the correct count");
-          }
-        }
```
keryell, line 82: What is the problem here?

tests/atomic_ref_stress/atomic_ref_stress_common.h
```diff
@@ -172,15 +172,15 @@ class aquire_release {
           refA.store(0);
           refB.store(0);
           sycl::group_barrier(item.get_group());
-          if (item.get_local_id() == 0) {
+          if (item.get_local_id() == sycl::id(0)) {
```
keryell, line 175: I do not think so, the constructors are implicit https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#id-class

tests/atomic_ref_stress/atomic_ref_stress_common.h
```diff
@@ -172,15 +172,15 @@ class aquire_release {
...
             x = refA.load();
             refB.store(1);
           } else {
             y = refB.load();
             refA.store(1);
           }
           sycl::group_barrier(item.get_group());
-          if (item.get_local_id() == 0)
+          if (item.get_local_id() == sycl::id(0))
```
keryell, line 183: Idem.

tests/exceptions/exceptions_constructors.cpp
```diff
@@ -194,8 +194,14 @@ TEST_CASE("Constructors for sycl::exception with sycl::errc error codes",
       check_exception(copy, errcode, sycl::sycl_category());
     }
     SECTION("operator=(const exception& other)") {
+      // sycl::exception is not default-constructible, so explicitly construct
```
keryell, line 197: Good point!

tests/hierarchical/hierarchical_non_uniform_local_range.cpp
```diff
@@ -50,6 +50,17 @@ void check_expected(const std::vector<sycl::int3> &data, unsigned local_id,
   }
 }
 
+// there is no id -> range conversion in the spec
+static sycl::range<1> id_to_range(sycl::id<1> id) {
+  return sycl::range<1>(id[0]);
```
keryell, line 55: This is true but there is 
```c++
// only available if Dimensions == 1
operator size_t() const;
```
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#id-class

tests/hierarchical/hierarchical_non_uniform_local_range.cpp
```diff
@@ -50,6 +50,17 @@ void check_expected(const std::vector<sycl::int3> &data, unsigned local_id,
...
+}
+static sycl::range<2> id_to_range(sycl::id<2> id) {
```
keryell, line 57:
```suggestion
static sycl::range<2> to_range(sycl::id<2> id) {
```
since the `id` is clear as the parameter?

tests/id/id.cpp
```diff
@@ -239,7 +239,8 @@ TEMPLATE_TEST_CASE_SIG("id can be implicitly conversion-constructed from item",
   q.submit([r, &result_buf](sycl::handler& cgh) {
      sycl::accessor result{result_buf, cgh, sycl::write_only};
      cgh.parallel_for<kernel_id<D>>(r, [=](sycl::item<D> itm) {
-       if (itm.get_id() == id<D>{r} - 1) {
+       // `ul` suffix is necessary to resolve ambiguity for id<1>
+       if (itm.get_id() == id<D>{r} - 1ul) {
```
keryell, line 243: Why `l`?
Just `u` does not work?
The real type would be `uz` but it is C++23.

tests/marray/marray_operators.h
```diff
@@ -822,13 +822,22 @@ class check_marray_pre_unary_operators_for_type {
     INFO("prefix unary operators for type \"" << type_name << "\": ");
 
     const auto num_elements = marray_common::get_num_elements();
-
-    static const auto unary_operators =
-        named_type_pack<op_upos, op_uneg, op_pre_inc, op_pre_dec, op_lnot,
-                        op_bnot>::generate("unary +", "unary -", "pre ++",
-                                           "pre --", "!", "~");
+    static const auto unary_operators = make_unary_operators();
```
keryell, line 825: Isn't it a good opportunity for a lambda with immediate invocation?

keryell (commented):
What a ride!

tests/atomic_ref_stress/atomic_ref_stress_common.h
keryell: Same here https://github.com/triSYCL/triSYCL/blob/master/include/triSYCL/detail/small_array.hpp#L381

tomdeakin:
The SYCL WG discussed this, and the review process should continue to merge this. We will then want to follow up on the new features proposed to possibly include in the spec. So, please continue to review and update this PR based on those reviews. Thanks!

bader:
> It makes me wonder if some of these should be part of the spec, simply for usability.

@steffenlarsen, could you open issues in SYCL-Docs project for the items which you think should be part of the spec, please?

steffenlarsen:
> > It makes me wonder if some of these should be part of the spec, simply for usability.
> 
> @steffenlarsen, could you open issues in SYCL-Docs project for the items which you think should be part of the spec, please?

Opened https://github.com/KhronosGroup/SYCL-Docs/issues/540 and https://github.com/KhronosGroup/SYCL-Docs/issues/539.

tests/accessor/generic_accessor_api_common.h
fknorr: It's the base template that's being specialized below. For a variable template marking it extern allows me to omit the (invalid) definition and cause a compiler error when it's used in a constexpr context.

tests/accessor_legacy/accessor_api_local_common.h
fknorr: This tests the legacy `accessor<target::local>` specialization, which does not have a `size()` member.

tests/atomic_ref_stress/atomic_ref_stress_common.h
fknorr: So either the special-casing needs to be in the spec, or this explicit construction is correct.

tests/hierarchical/hierarchical_non_uniform_local_range.cpp
fknorr: Yep, but this is a generic function and the case through size_t only works for Dims=1.

tests/id/id.cpp
fknorr: Not sure, but `size_t` is at least `unsigned long` on all relevant architectures, no?

tests/marray/marray_operators.h
fknorr: It is!

tests/vector_operators/generate_vector_operators.py
```diff
@@ -207,92 +207,93 @@
...
-    resVec = --testVec1Copy.${swizzle};
-    if (!check_vector_values(resVec, resArr)) {
-      resAcc[0] = false;
-    }
-    if (!check_vector_values(testVec1Copy, resArr)) {
-      resAcc[0] = false;
-    }
+  // The standard does not require ++ and -- (either prefix or postfix) for the
+  // bool type, so we skip tests for these operators if DataT is bool. This cannot
+  // be done via `if constexpr` because the type is not a dependent type.
```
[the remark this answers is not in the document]
fknorr: second.

tests/hierarchical/hierarchical_non_uniform_local_range.cpp
fknorr: This is obsolete now because it has been fixed on SYCL-2020 branch in the meantime.

fknorr:
Rebased, which allowed me to drop one commit whose problem has been addressed on the target branch in the meantime.

keryell (approved):
Thanks!

tests/accessor/generic_accessor_api_common.h
keryell: Why not. :-)

keryell (approved):
Thanks!

bader:
@fknorr, please, fix the formatting check.

PR #873, opened 2024-02-17
Compatibility: Only use `cl_*` types if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS is set

fknorr:
This improves compatibility with implementations that do not expose OpenCL interop (notably [SimSYCL](https://github.com/celerity/SimSYCL), and AFAIK also AdaptiveCpp / hipSYCL).

This change is probably incomplete, but enough to get some tests to build with SimSYCL that otherwise wouldn't. 

`group_async_work_group_copy.cpp` had `!is_same` checks between cl types and stdint types - please correct me if it's wrong to remove these checks, it appears to me as though performing the same check twice is not an issue - but maybe there's symbol clashes.

### DIFF
--- a/tests/group/group_async_work_group_copy.cpp
+++ b/tests/group/group_async_work_group_copy.cpp
@@ -93,28 +93,59 @@ class TEST_NAME : public util::test_base {
 #endif
 
 #ifdef INT8_MAX
-    for_type_and_vectors<check_type, std::int8_t>(log, "std::int8_t");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
+    if (!std::is_same<sycl::cl_char, std::int8_t>::value)
 #endif
+      for_type_and_vectors<check_type, std::int8_t>(log, "std::int8_t");
+#endif
+
 #ifdef INT16_MAX
-    for_type_and_vectors<check_type, std::int16_t>(log, "std::int16_t");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
+    if (!std::is_same<sycl::cl_short, std::int16_t>::value)
 #endif
+      for_type_and_vectors<check_type, std::int16_t>(log, "std::int16_t");
+#endif
+
 #ifdef INT32_MAX
-    for_type_and_vectors<check_type, std::int32_t>(log, "std::int32_t");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
+    if (!std::is_same<sycl::cl_int, std::int32_t>::value)
 #endif
+      for_type_and_vectors<check_type, std::int32_t>(log, "std::int32_t");
+#endif
+
 #ifdef INT64_MAX
-    for_type_and_vectors<check_type, std::int64_t>(log, "std::int64_t");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
+    if (!std::is_same<sycl::cl_long, std::int64_t>::value)
 #endif
+      for_type_and_vectors<check_type, std::int64_t>(log, "std::int64_t");
+#endif
+
 #ifdef UINT8_MAX
-    for_type_and_vectors<check_type, std::uint8_t>(log, "std::uint8_t");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
+    if (!std::is_same<sycl::cl_uchar, std::uint8_t>::value)
 #endif
+      for_type_and_vectors<check_type, std::uint8_t>(log, "std::uint8_t");
+#endif
+
 #ifdef UINT16_MAX
-    for_type_and_vectors<check_type, std::uint16_t>(log, "std::uint16_t");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
+    if (!std::is_same<sycl::cl_ushort, std::uint16_t>::value)
 #endif
+      for_type_and_vectors<check_type, std::uint16_t>(log, "std::uint16_t");
+#endif
+
 #ifdef UINT32_MAX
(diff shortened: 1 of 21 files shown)


tests/nd_item/nd_item_async_work_group_copy_fp16.cpp
```diff
@@ -37,8 +37,10 @@ class TEST_NAME : public util::test_base {
       // Test using queue constructed already
       for_type_and_vectors<check_type, sycl::half>(queue, log,
           "sycl::half");
+#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS
       for_type_and_vectors<check_type, sycl::cl_half>(queue, log,
           "sycl::cl_half");
```
bader, line 42: The tests seem to be outdated.
Spec [says](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_data_types_2):

> For the purpose of interoperability and portability, SYCL defines a set of aliases to C++ types within the sycl::opencl namespace using the cl_ prefix. 

```suggestion
          "sycl::opencl::cl_half");
```
Right?

tests/group/group_async_work_group_copy.cpp
```diff
@@ -89,38 +90,31 @@ class TEST_NAME : public util::test_base {
         "sycl::cl_ulong");
     for_type_and_vectors<check_type, sycl::cl_float>(log,
         "sycl::cl_float");
+#endif
 
 #ifdef INT8_MAX
-    if (!std::is_same<sycl::cl_char, std::int8_t>::value)
-      for_type_and_vectors<check_type, std::int8_t>(log, "std::int8_t");
+    for_type_and_vectors<check_type, std::int8_t>(log, "std::int8_t");
```
bader, line 96: This introduces duplication into the test. But you can move the if condition under `#if SYCL_CTS_ENABLE_OPENCL_INTEROP_TESTS`.

tests/group/group_async_work_group_copy.cpp
```diff
@@ -89,38 +90,31 @@ class TEST_NAME : public util::test_base {
         "sycl::cl_ulong");
     for_type_and_vectors<check_type, sycl::cl_float>(log,
         "sycl::cl_float");
+#endif
 
 #ifdef INT8_MAX
-    if (!std::is_same<sycl::cl_char, std::int8_t>::value)
```
keryell, line 94: I wonder how these tests could have been false at the first place.

keryell:
@fknorr Some merge from upstream?

bader (approved):
@fknorr, thanks!
LGTM. Please, fix `check-clang-format`.

tests/nd_item/nd_item_async_work_group_copy_fp16.cpp
fknorr: I've updated most uses of `cl_*`, however there's the `cl_float4` etc vector types which also seem to be deprecated, and DPC++ doesn't alias them into `opencl::`. I've left those alone.

tests/group/group_async_work_group_copy.cpp
fknorr: Done, although this makes things a lot more verbose.

tests/group/group_async_work_group_copy.cpp
fknorr: I don't think they can, integer widths are part of the spec.

tests/nd_item/nd_item_async_work_group_copy_fp16.cpp
bader: > there's the `cl_float4` etc vector types which also seem to be deprecated, and DPC++ doesn't alias them into `opencl::`.

@steffenlarsen, could you please check the DPC++, please? I found https://github.com/intel/llvm/commit/e98d732406cc4ba9734cfd8ece96f6d19ddf7267 by @AlexeySachkov, which adds aliases for scalar types, but I don't see definitions for vector aliases in https://github.com/intel/llvm/blob/e89da3251187ec101f995a8dbf6e832043baebdc/sycl/include/sycl/aliases.hpp.

We need to create trackers to fix both CTS tests (i.e. `sycl::cl_float4` -> `sycl::opencl::cl_float4`) and DPC++ implementation (i.e. implement `sycl::opencl::cl_float4`).

bader: I found the spec defines aliases only for scalar OpenCL types https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_data_types_2.

> The OpenCL C language standard [Section 6.11](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#opencl12) defines its own built-in scalar data types, and these have additional requirements in terms of size and signedness on top of what is guaranteed by ISO C++. For the purpose of interoperability and portability, SYCL defines a set of aliases to C++ types within the sycl::opencl namespace using the cl_ prefix. These aliases are described in [Table 182](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.types.aliases).

I'm not sure what is the reason to skip vector types, but at the moment, we can open an issue for SYCL-CTS to replace OpenCL vector type aliases with sycl::vec type (i.e. `sycl::cl_float4` -> `sycl::vec<sycl::opencl::cl_float, 4>`). 

I recalled that @AlexeySachkov already filed an issue about missing vector aliases - https://github.com/KhronosGroup/SYCL-Docs/issues/335, but it's still unresolved.

steffenlarsen: I agree with you analysis, but it sounds like DPC++ is defining the right aliases until https://github.com/KhronosGroup/SYCL-Docs/issues/335 is resolved. Please correct me if I'm wrong.

keryell (approved):
Thanks!

PR #874, opened 2024-02-21
Rename hipSYCL to AdaptiveCpp, update CI and regenerate filter

### LINKED ISSUE #796: Bump hipSYCL version and regenerate filters
We haven't updated the hipSYCL version used in CI for over a year ([3d8b1cd ](https://github.com/KhronosGroup/SYCL-CTS/blob/c6711e6dc6c970a374c03a3e2a55b01383991656/.github/workflows/cts_ci.yml#L61C22-L61C29) is from [January 2022](https://github.com/OpenSYCL/OpenSYCL/commit/3d8b1cd)).

We then should [regenerate](https://github.com/KhronosGroup/SYCL-CTS/blob/282acab13ce78bbbeca5a4fac96eeda88f18514a/ci/generate_exclude_filter.py) the CI filters for hipSYCL as well. Although one problem we have here is that presumably most of the tests are compile-time disabled for hipSYCL, and we currently don't have a good way of figuring out which ones we can re-enable...

psalz:
NOTE: This includes @fknorr's recent PRs (#870, #871, #872, #873), so those should be merged first.

This renames (hopefully) all uses of hipSYCL to AdaptiveCpp and updates the version checked against in CI to the latest develop commit (the previous version was over 2 years old at this point).

I've also gone ahead and re-generated the CI exclude filters: AdaptiveCpp is now able to compile 7 additional categories (device, exception_handling, image, marray, nd_range, range and vector_alias), while failing 2 new categories (group_functions and optional_kernel_features, although I believe those didn't even exist yet the last the filter was generated).

I had to bump our CI docker container base image to Ubuntu 22.04, as the GCC included in 20.04 is too old for current AdaptiveCpp.

I've also taken the liberty of removing the long deprecated `Intel_SYCL` option for `SYCL_IMPLEMENTATION`.

cc @illuhad

Closes #796.


bader:
> NOTE: This includes @fknorr's recent PRs (#870, #871, #872, #873), so those should be merged first.

I'm not sure it's a good idea. These PRs have some unresolved comments and pre-commit failures, which will take time to resolve. Requesting them to be merge first blocks other useful work, which is not directly related to adding SimSYCL support.

I think we can merge fix for https://github.com/KhronosGroup/SYCL-CTS/issues/796 and remove the Intel_SYCL separately.

psalz:
> I'm not sure it's a good idea. These PRs have some unresolved comments and pre-commit failures, which will take time to resolve. Requesting them to be merge first blocks other useful work, which is not directly related to adding SimSYCL support.
> 
> I think we can merge fix for #796 and remove the Intel_SYCL separately.

Are we in a hurry to get this merged? IMO waiting another week or two to get those issues resolved doesn't really make a difference at this point. My main goal for today's call was to assess AdaptiveCpp's current CTS coverage, and not including @fknorr's fixes will negatively affect that.

keryell:
How do we move on this?

keryell:
@fknorr Some merge conflicts.
We could use also Ubuntu 24.04 to have some recent compilers out of the box?

bader:
> I've also taken the liberty of removing the long deprecated `Intel_SYCL` option for `SYCL_IMPLEMENTATION`.

This part has been removed from the PR.

docker/common/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```
bader, line 1: Can we switch to 24.04?

psalz: Yes sure, good idea!

psalz:
Well this has taken way too long, sorry about that. I've re-done all of the text replacements and updated the AdaptiveCpp image to a recent commit. It looks like there's fewer test categories compiling now than when I originally opened the PR, but I haven't checked any of the errors, so there might be some low hanging fruit.

bader (approved):

docker/adaptivecpp/Dockerfile
```diff
@@ -14,15 +14,22 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
+    cd /tmp/AdaptiveCpp && \
     git checkout $IMPL_VERSION && \
-    cmake /tmp/hipSYCL -G Ninja -B /tmp/build \
+    cmake /tmp/AdaptiveCpp -G Ninja -B /tmp/build \
+      -DWITH_SSCP_COMPILER=OFF \
+      -DWITH_STDPAR_COMPILER=OFF \
+      -DWITH_ACCELERATED_CPU=OFF \
+      -DWITH_CUDA_BACKEND=OFF \
+      -DWITH_ROCM_BACKEND=OFF \
+      -DWITH_OPENCL_BACKEND=OFF \
```
illuhad, line 28: For #1012, we should decide for whether we want an LLVM-based build. In this context, I wonder whether there is any issue or challenge enabling the compiler features? 

Since AdaptiveCpp even when only targeting CPU now typically relies on LLVM (via `--acpp-targets=generic` by either JIT-compiling for OpenMP or OpenCL), we could use such a build without having to set up CUDA or ROCm and without needing GPU hardware.

As most users also are going to be using `--acpp-targets=generic`, that might be useful to have in CTS CI, if possible.

bader: @illuhad, I managed to install LLVM and successfully built default version of AdaptiveCpp. Unfortunately, I can't merge #1012 because SYCL-CTS code base relies on hipSYCL. I think this PR resolve the issue blocking #1012.
Could you please approve this PR and review #1012?

psalz: I suggest to merge this PR first and then look into switching to SSCP in a follow up. It should only require changes to the Dockerfile (hopefully), and by merging this now we can get all the string replacements out of the way before they accrue too many conflicts with other PRs.

bader: Sounds good to me. @illuhad?

illuhad: Yep, okay for me!

illuhad (approved):

PR #875, opened 2024-02-21
Adds test for the non_uniform_groups oneAPI extension

steffenlarsen:
This commit adds tests for the [sycl_ext_oneapi_non_uniform_groups](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_non_uniform_groups.asciidoc) extension in accordance with the test plan currently in review here: https://github.com/KhronosGroup/SYCL-CTS/pull/866


tests/extension/oneapi_non_uniform_groups/ballot_group_api.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    cgh.parallel_for<struct BallotGroupAPI>(nd_range, [=](sycl::nd_item<1> it) {
+      auto& results = acc[it.get_global_id()];
+
+      sycl::sub_group sg = it.get_sub_group();
+      size_t sub_group_size = sg.get_local_range().size();
+      size_t split = sub_group_size / 3;
+      bool is_left = sg.get_local_linear_id() < split;
+
+      auto ballot = oneapi_ext::get_ballot_group(sg, is_left);
```
dm-vodopyanov, line 72: Should this be checked explicitly?
```
Constraints: Available only if `sycl::is_group_v<std::decay_t<Group>> &&
std::is_same_v<Group, sycl::sub_group>` is true.
```

tests/extension/oneapi_non_uniform_groups/non_uniform_group_common.h
```diff
@@ -0,0 +1,206 @@
...
+      case 3:
+        return "predicate is false for all work-items";
+    }
+    return "";
+  }
+};
+
+template <size_t PartitionSize>
+struct NonUniformGroupHelper<
+    oneapi_ext::fixed_size_group<PartitionSize, sycl::sub_group>> {
```
dm-vodopyanov, line 87: Should this be checked explicitly?
```
`PartitionSize` must be positive and a power of 2.
```

tests/extension/oneapi_non_uniform_groups/fixed_size_group_api.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    cgh.parallel_for<struct FixedSizeGroupAPI>(
+        nd_range, [=](sycl::nd_item<1> it) {
+          auto& results = acc[it.get_global_id()];
+
+          sycl::sub_group sg = it.get_sub_group();
+          size_t sub_group_size = sg.get_local_range().size();
+
+          auto fixed_size =
+              oneapi_ext::get_fixed_size_group<partition_size>(sg);
```
dm-vodopyanov, line 75: Should there be a negative test?
```
Preconditions: `PartitionSize` must be less than or equal to the result of `group.get_max_local_range()`. `group.get_local_linear_range()` must be evenly divisible by `PartitionSize`.
```

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
```diff
@@ -0,0 +1,182 @@
...
+                               sycl::range<1>{num_items}};
+
+    cgh.parallel_for<struct OpportunisticGroupAPI>(
+        nd_range, [=](sycl::nd_item<1> it) {
+          auto& results = acc[it.get_global_id()];
+
+          size_t sub_group_size = it.get_sub_group().get_local_range().size();
+
+          auto opportunistic =
+              oneapi_ext::this_kernel::get_opportunistic_group();
```
dm-vodopyanov, line 71: Should this be checked?
```
Each call to `get_opportunistic_group()` returns a different group.
```

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
```diff
@@ -0,0 +1,182 @@
...
+  // Aliases.
+  STATIC_CHECK(std::is_same_v<opportunistic_group_t::id_type, sycl::id<1>>);
+  STATIC_CHECK(
+      std::is_same_v<opportunistic_group_t::range_type, sycl::range<1>>);
+  STATIC_CHECK(std::is_same_v<opportunistic_group_t::linear_id_type, uint32_t>);
+
+  // Static constexpr members.
+  STATIC_CHECK(opportunistic_group_t::dimensions == 1);
+  STATIC_CHECK(opportunistic_group_t::fence_scope ==
+               sycl::sub_group::fence_scope);
```
dm-vodopyanov, line 125: Is `sycl::sub_group::fence_scope` the same as `sycl::memory_scope::sub_group`?

tests/extension/oneapi_non_uniform_groups/ballot_group_api.cpp
steffenlarsen: I don't believe the CTS normally does negative testing for availability like this. The test will fail if it isn't correctly available.

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
steffenlarsen: That is a good question. @Pennycook - What is the intention here? Do the groups have [common by-value semantics](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:byval-semantics) and if so, does this make the guarantee that `get_opportunistic_group() != get_opportunistic_group()`, which in turn means the groups will never have the same members values?

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
steffenlarsen: They are the same, but you are right that we should check this based on the definition of the class, so I've changed it.

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
Pennycook: That is a very good question.

The intention of the wording was really to say that the members _may_ be different, and that in cases where the members _are_ different, the group objects may be different.  Without thinking of any specific implementation: if the group members are different, the representation (e.g., as a mask) will probably be different; even if the group members are the same, I can imagine situations where the different group objects would still have dedicated scratchpad memory or barriers.

I think the simplest thing to do for now is to guarantee that equality returns false.  It's probably more useful to have a comparison that is always useless than to try and define the cases in which equality should/could return true.  I'd prefer not to provide a comparison operator at all, but I've heard suggestions recently that this would break a lot of things.

tests/extension/oneapi_non_uniform_groups/group_shift.h
```diff
@@ -0,0 +1,149 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2023 The Khronos Group Inc.
```
dm-vodopyanov, line 5:
```suggestion
//  Copyright (c) 2024 The Khronos Group Inc.
```

tests/extension/oneapi_non_uniform_groups/group_permute_fp64.cpp
```diff
@@ -0,0 +1,44 @@
+/*******************************************************************************
+//
+//  SYCL 2020 Conformance Test Suite
+//
+//  Copyright (c) 2023 The Khronos Group Inc.
```
dm-vodopyanov, line 5:
```suggestion
//  Copyright (c) 2024 The Khronos Group Inc.
```

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
gmlueck: > I think the simplest thing to do for now is to guarantee that equality returns false. It's probably more useful to have a comparison that is always useless than to try and define the cases in which equality should/could return true. I'd prefer not to provide a comparison operator at all, but I've heard suggestions recently that this would break a lot of things.

Are you saying that `opportunistic_group` needs to have a comparison operator?  If you are saying that things would break if there is no comparison operator, it seems surprising that those things wouldn't also break if the comparison operator always returned false.

Are you saying that the comparison operator would return false even when comparing an object with a copy?  For example:

```
opportunistic_group og1 = /*...*/;
opportunistic_group og2 = og1;
bool same = (og1 == og2);  // returns false?
```

Pennycook: > Are you saying that `opportunistic_group` needs to have a comparison operator? If you are saying that things would break if there is no comparison operator, it seems surprising that those things wouldn't also break if the comparison operator always returned false.

I don't have complete information here.  This came up in a discussion we were having about some of the other types (e.g., `nd_item`), and my takeaway was that that if we were to remove certain operators then the classes wouldn't play nicely with certain type traits.  I might have gotten the wrong end of the stick.

> Are you saying that the comparison operator would return false even when comparing an object with a copy? For example:
> 
> ```
> opportunistic_group og1 = /*...*/;
> opportunistic_group og2 = og1;
> bool same = (og1 == og2);  // returns false?
> ```

Yes, that's what I was suggesting.

I'm also open to defining a real comparison operator for groups, but I don't know what the criteria should be.  I can see an argument that it should be based on which work-items are in the group (which would allow comparisons across groups of different types) but can also think of cases where that could lead to surprising results (e.g., just because two groups have the same work-items in them doesn't necessarily mean they can be used interchangably).

keryell:
Interesting discussion about defining group equality and usability. Related to also whether we can use them in another work-group or kernel by storing them into an intermediate buffer.
To clarify for SYCL Next. :-)

tests/extension/oneapi_non_uniform_groups/opportunistic_group_api.cpp
steffenlarsen: If this is an open issue for groups in general, could we leave it out of these tests for now and address it for all groups when it has been discussed and consensus has been reached on how it should work?

tests/extension/oneapi_non_uniform_groups/group_barrier.h
```diff
@@ -0,0 +1,215 @@
...
+
+            GroupT non_uniform_group =
+                NonUniformGroupHelper<GroupT>::create(sub_group, test_case);
+
+            size_t llid = non_uniform_group.get_local_linear_id();
+            size_t max_id = non_uniform_group.get_local_linear_range() - 1;
+
+            ASSERT_RETURN_TYPE(
+                void, sycl::group_barrier(non_uniform_group),
+                "Return type of group_barrier(GroupT g) is wrong\n");
```
AlexeySachkov, line 137: In the other test you've used plain `static_assert`. Should we unify? 
Note: my preference would be in using plain `static_assert`, because it is easier to understand at a very first glance.

tests/group_functions/group_shift.h
```diff
@@ -2,7 +2,7 @@
 //
 //  SYCL 2020 Conformance Test Suite
 //
-//  Copyright (c) 2023 The Khronos Group Inc.
+//  Copyright (c) 2024 The Khronos Group Inc.
```
AlexeySachkov, line 5: Unrelated change?

tests/extension/oneapi_non_uniform_groups/group_barrier.h
steffenlarsen: I agree. All uses of `ASSERT_RETURN_TYPE` have been made `static_assert` instead.

tests/group_functions/group_shift.h
steffenlarsen: Because the test files have the same names, it just seems that I changed the year in the wrong `group_shift.h`. It has been addressed.

dm-vodopyanov (approved):
LGTM!

AlexeySachkov (approved):

PR #881, opened 2024-03-21
Generate clamp inputs according to precondition in the spec

lbushi25:
This PR aims to modify input generation for the clamp function tests which require its third argument to be greater than or equal to its second argument in every dimension as per the spec. Without this PR the input generation is completely random and not restricted according to the above conditions. A small bug has also been fixed where a local map variable was left uninitialized and it was causing random test failures in the math_builtin_api suite.


lbushi25:
@AlexeySachkov FYI

keryell:
Ouch!
While you are here, can you move some knowledge of your PR description as some comments in your code so future generation of readers will get the idea?

lbushi25:
> Ouch! While you are here, can you move some knowledge of your PR description as some comments in your code so future generation of readers will get the idea?

Sure!

lbushi25:
Update: Made a small bug fix in util/math_helper.h where a map variable was left uninitialized.

keryell (approved):
Thanks for the fix plus the teaching comments!

PR #883, opened 2024-03-22
Remove `std_classes` test

AlexeySachkov:
The test seems to be 1.2.1 legacy, which doesn't make sense in SYCL 2020 testing:
- the test does not check anything at runtime
- the only compile-time "check" which is performed is successful compilation of the test source itself and there are no static assertions
- test source code contains nothing SYCL-specific and only uses standard C++ functionality

The test used to check type aliases like `cl::sycl::string_class`, but it lost its purpose in e08cb4541df374d82537f9cdc0db752d1cf6fd56 when it was updated to match SYCL 2020 interface that directly uses standard C++ containers.


keryell (approved):
Good catch!

PR #884, opened 2024-03-25
Align marray ctor checking with comments

jiezzhang:
Remove redundant host `run_checks` in marray_operators.h. It will overwrite all device checking results.


keryell (approved):
Thanks.

PR #885, opened 2024-03-26
Vec<1> to scalar conversion

lbushi25:
This PR adds a test to verify that the conversion of a vector of 1 element to a scalar value of that element's type works.

### DIFF
--- a/tests/vector_api/generate_vector_api.py
+++ b/tests/vector_api/generate_vector_api.py
@@ -109,7 +109,6 @@ def gen_checks(type_str, size):
         test_string += lo_hi_odd_even_template.substitute(
         type=type_str,
         vals=', '.join(vals_list))
-    test_string += vector1_to_scalar_convert
     string = wrap_with_kernel(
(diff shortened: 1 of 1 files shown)


lbushi25:
@AlexeySachkov FYI

tests/vector_api/generate_vector_api.py
```diff
@@ -75,6 +75,11 @@
         }
 """)
 
+vector1_to_scalar_convert = Template("""
+    ${type} scalar = sycl::vec<${type}, 1>();
+    CHECK(${type}{} == scalar);
+    CHECK(sycl::vec<${type}, 1>().size() == 1);
```
keryell, line 81: Could you add also some real value instead of the default one.
For example if the assignment would be a no-op, this test would pass. :-)

lbushi25:
Hi @keryell , turns out that this behavior is already tested in tests/vector_operators/generate_vector_operators.py.
I'm therefore closing the PR. Sorry for jumping the gun!

PR #887, opened 2024-04-09
[NFC] Refactor kernel_compiler_spirv tests

0x12CC:
Updates the test to load the SPIR-V module files using CMake instead of `std::ifstream`. This change makes the test binary more portable since it now contains the required SPIR-V modules.


tests/extension/oneapi_kernel_compiler_spirv/kernel_compiler_spirv.cpp
```diff
@@ -22,22 +22,26 @@
 
 namespace kernel_compiler_spirv::tests {
 
+static const std::vector<uint8_t> kernels{
```
keryell, line 25: Add a comment explaining why it is not possible to have `std::vector<std::byte>` everywhere at the first place or use `std::vector<std::byte>`.

keryell:
@0x12CC Ping.

keryell (approved):
Thanks!

PR #888, opened 2024-04-22
List excluded generated tests

KornevNikita:
There are some test categories (e.g. math_builtin_api) that need to be generated. Such categories are not listed during build configuration as "skipped" if they are excluded.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -198,12 +198,16 @@ function(add_cts_test)
   # To make check that any .cpp files are passed
   # List created because, direct check on "${ARGN}" gives false-positive result
   set(tests_list "${ARGN}")
+  get_filename_component(test_dir ${CMAKE_CURRENT_SOURCE_DIR} NAME)
   if (tests_list)
-    get_filename_component(test_dir ${CMAKE_CURRENT_SOURCE_DIR} NAME)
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -198,12 +198,16 @@ function(add_cts_test)
   # To make check that any .cpp files are passed
   # List created because, direct check on "${ARGN}" gives false-positive result
   set(tests_list "${ARGN}")
+  get_filename_component(test_dir ${CMAKE_CURRENT_SOURCE_DIR} NAME)
   if (tests_list)
-    get_filename_component(test_dir ${CMAKE_CURRENT_SOURCE_DIR} NAME)
     set(test_exe_name ${test_dir})
```
keryell, line 203: Then this `test_dir` variable is now undefined?

KornevNikita: Hm, why, am I missing something? Now we just defining it in advance before `if` block, so we don't have to do it both in `if` and `else` blocks.

keryell: Ah yes. I do not know why I have not seen the change above too!
Sorry for the confusion. :frowning_face:

keryell (approved):
Thanks!

PR #890, opened 2024-05-01
Update oclmath to add AArch64 support.

hvdijk:
This replaces fpcontrol.h and rounding_mode.cpp/.h with the versions from OpenCL CTS. The old version included in SYCL-CTS did not support AArch64, and although it would be possible to just make the smallest possible change to get that to work, it seems like it will be easier for future changes for the OpenCL and SYCL versions of these files to be the same.

### DIFF
--- a/oclmath/rounding_mode.h
+++ b/oclmath/rounding_mode.h
@@ -1,58 +1,57 @@
-
-/******************************************************************
- //
- //  OpenCL Conformance Tests
- // 
- //  Copyright:	(c) 2008-2013 by Apple Inc. All Rights Reserved.
- //
- ******************************************************************/
-
+//
+// Copyright (c) 2017 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
 #ifndef __ROUNDING_MODE_H__
 #define __ROUNDING_MODE_H__
 
 #include "compat.h"
 
-typedef enum
-{
-    kDefaultRoundingMode = 0,
-    kRoundToNearestEven,
-    kRoundUp,
-    kRoundDown,
-    kRoundTowardZero,
-
-    kRoundingModeCount
-}RoundingMode;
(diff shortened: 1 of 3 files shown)


hvdijk:
CI is showing that the OpenCL CTS versions of these files are not clang-formatted. But if we clang-format them here, we do not avoid the problem that we needlessly have two different versions of these files. My thinking here is that two alternatives are preferable: we could exclude them from clang-formatting, leaving the responsibility with OpenCL CTS to keep it maintainable. Or we could apply the formatting to OpenCL CTS, if its style aligns sufficiently closely with SYCL CTS. Feedback would be welcome here.

keryell:
Can anyone remember me why we had these files in the CTS at the first place?

bader:
> Can anyone remember me why we had these files in the CTS at the first place?

@keryell, I think it was added to check the accuracy requirements for SYCL math functions in SYCL-1.2.1. @AerialMantis, did I get it right?

bader:
I think we need someone with ARM expertise to review these changes. Considering that we borrowed this code from OpenCL-CTS, I think it's reasonable to ask @kpet to review this change.

keryell:
@kpet nous avons besoin d'aide !

kpet:
@keryell I'm on holiday at the moment but I've sent myself a reminder to look into this PR when I get back.

keryell:
> @keryell I'm on holiday at the moment but I've sent myself a reminder to look into this PR when I get back.

Thanks! There is no emergency since unfortunately this PR is old, but it would be nice to have the feedback from a specialist.

kpet:
We've been using the OpenCL-CTS code on AArch64 targets for a good few years now and have not identified issues with it yet (that I know of!). I think this change can go in as is if the logic was ported from the OpenCL CTS. FWIW, I couldn't spot anything glaringly wrong after a very quick review.

Long term we should think about getting support for low-level FP control into C libraries (or maybe C++). Any volunteers? :P

keryell (approved):
Thanks!

oclmath/rounding_mode.h
```diff
@@ -1,58 +1,57 @@
...
-extern RoundingMode set_round( RoundingMode r, Type outType );
-extern RoundingMode get_round( void );
-extern void *FlushToZero( void );
-extern void UnFlushToZero( void *p);
-
-#ifdef __cplusplus
-}
+#if (defined(_WIN32) && defined(_MSC_VER))
+#include "errorHelpers.h"
+#include "testHarness.h"
```
bader, line 23: @hvdijk, these files are missing! The build of `oclmath` with MSVC-compatible compiler fails. Can you fix the build, please?

hvdijk: Oh dear, sorry about that. Taking a look now.

hvdijk: #984

PR #891, opened 2024-05-07
Replace reserve with resize for write-back buffer

steffenlarsen:
In the buffer storage tests, the case using a vector iterator reserves size in the buffer and then expects to be able to access through the specified size. However, this is not the intention of reserve, as it does not change the end of the vector. Instead, this should use resize.

### DIFF
--- a/tests/buffer/buffer_storage_common.h
+++ b/tests/buffer/buffer_storage_common.h
@@ -70,7 +70,7 @@ class buffer_storage_test {
 
     // Case 5 - Vector data
     std::vector<T> data_vector;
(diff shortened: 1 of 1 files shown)


keryell (approved):
Good security catch!

bader:
> However, this is not the intention of **resize**, as it does not change the end of the vector.

resize -> reserve.

steffenlarsen:
> > However, this is not the intention of **resize**, as it does not change the end of the vector.
> 
> resize -> reserve.

Force-pushed to update the message. 👍

PR #892, opened 2024-05-07
Add "absolute error tolerance" accuracy mode check

steffenlarsen:
Though the expected precision of math builtins are now defined by the backend specifications, the CTS currently checks the error accuracy based on the tolerances specified by the OpenCL C specification. For `mix` and `smoothstep` these tolerances are defined in "absolute error tolerance" rather than in ULP. This commit makes it possible for the tests to switch between ULP and "absolute error tolerance" checks based on the builtins. Likewise, it applies these to `mix` and `smoothstep` in accordance with the OpenCL C specification.

### DIFF
--- a/tests/math_builtin_api/math_builtin.h
+++ b/tests/math_builtin_api/math_builtin.h
@@ -79,16 +79,28 @@ inline sycl::half min_t<sycl::half>() {
   return static_cast<sycl::half>(powf(2.0f, -14.0f));
 }
 
+enum AccuracyMode { ULP, AbsoluteTolerance };
+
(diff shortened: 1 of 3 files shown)


tests/math_builtin_api/math_builtin.h
```diff
@@ -79,16 +79,27 @@ inline sycl::half min_t<sycl::half>() {
   return static_cast<sycl::half>(powf(2.0f, -14.0f));
 }
 
+enum AccuracyMode { ULP, AbsoluteTolerance };
```
keryell, line 82: `enum class`

keryell (approved):
Thanks!

PR #895, opened 2024-05-15
Update test_accessor according to the spec change

KornevNikita:
accessor::get_pointer() should return global_pointer<value_type>, but not global_pointer<DataT>
Spec: https://github.com/KhronosGroup/SYCL-Docs/pull/549


KornevNikita:
@keryell hi! Could you please take a look?

keryell (approved):
Thanks.

PR #901, opened 2024-06-13
Split up vector_swizzle tests into smaller chunks to improve compilation times

lbushi25:
The vector swizzle tests are failing to compile on machines with 16 GB of RAM due to the huge size of some of the test .cpp files that are generated by the scripts. This PR makes the necessary changes to enforce splitting of these test files into smaller and more manageable files that can be compiled without requiring a large memory footprint. As of the time of this writing, I've chosen the number of batches aka chunks to be 32 so that each large file will be subdivided into 32 smaller files. I chose this number because with 16 chunks I was seeing flaky results where a compilation would succeed or fail depending on the state of the machine at the time of compilation. With 32 however, compilation was succeeding constantly. Nevertheless, this can easily be adjusted according to our needs.

### DIFF
--- a/tests/common/common_python_vec.py
+++ b/tests/common/common_python_vec.py
@@ -23,6 +23,7 @@
 from collections import defaultdict
 from string import Template
 from itertools import product
+from math import ceil, floor
 
 class Data:
     signs = [True, False]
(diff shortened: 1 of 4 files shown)


tests/common/common_python_vec.py
```diff
@@ -23,6 +23,7 @@
 from collections import defaultdict
 from string import Template
 from itertools import product
+from math import ceil, floor
```
keryell, line 26: While you are here, you could sort the imports. :-)

keryell (approved):
Thanks!
Your PR description is an interesting design argumentation which should appear somewhere in some comments to avoid being lost.

tests/common/common_python_vec.py
lbushi25: Sure, I went ahead and ordered them  based on the module name. Hope that's the preferred way of doing it.

PR #898, opened 2024-05-31
Add test plan for `sycl_ext_oneapi_enqueue_functions` extension

0x12CC:

PR #902, opened 2024-06-13
Add tests for `sycl_ext_oneapi_enqueue_functions` extension

0x12CC:
Implement the tests from #898.


keryell (approved):

tests/extension/oneapi_enqueue_functions/enqueue_functions.cpp
```diff
@@ -0,0 +1,397 @@
...
+//  limitations under the License.
+//
+*******************************************************************************/
+
+#include "../../common/common.h"
+
+namespace enqueue_functions::tests {
+
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS
+namespace syclex = sycl::ext::oneapi::experimental;
```
steffenlarsen, line 26: Small nit, but I would prefer to rename this to something like `oneapi_ext` or similar, to make it clear it's a oneapi ext.

steffenlarsen (commented):
All the tests that run with `syclex::submit`, should we also test those with `submit_with_event`?

steffenlarsen:
@gmlueck - Any thoughts on the case mentioned in https://github.com/KhronosGroup/SYCL-CTS/pull/902#issuecomment-2165786005? Can we use a `kernel` from a `kernel_bundle` or are the arguments of those less well specified than those from interop kernels?

gmlueck:
> @gmlueck - Any thoughts on the case mentioned in [#902 (comment)](https://github.com/KhronosGroup/SYCL-CTS/pull/902#issuecomment-2165786005)? Can we use a `kernel` from a `kernel_bundle` or are the arguments of those less well specified than those from interop kernels?

I agree with @0x12CC, there is no portable way to test `handler::set_arg` using APIs from the core SYCL specification.  Probably the best way to test these APIs is to use [sycl_ext_oneapi_free_function_kernels](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_oneapi_free_function_kernels.asciidoc), but we are still implementing that extension in DPC++.  I'd suggest waiting until we write CTS tests for sycl_ext_oneapi_free_function_kernels, and then exercising these overloads of the enqueue functions as part of the free-function kernel tests.

steffenlarsen (approved):
Looks good!

PR #903, opened 2024-06-20
Fix check for composite devices not being root devices

steffenlarsen:
The logic in the test for checking that composite devices are not root devices as it checks that "none of the root devices *are not* the same as any composite device" instead of "none of the root devices *are* the same as any composite device." This commit fixes this check.


AlexeySachkov (approved):

keryell (approved):
Thanks.

PR #909, opened 2024-07-10
Add raw pointer cases to math builtin tests

steffenlarsen:
Following https://github.com/KhronosGroup/SYCL-Docs/pull/555 this commit adds raw pointer tests to the cases generated by the SYCL math builtin generator script.


keryell (approved):
Thanks!

PR #910, opened 2024-07-10
Expand testing of vec load and store

steffenlarsen:
This commit expands the testing of vec load and store with:
* Test cases for multi_ptr with local and private address space.
* Test cases for raw pointers with global, local and private address space. This functionality was added in https://github.com/KhronosGroup/SYCL-Docs/pull/555.


keryell (approved):
Thanks!

PR #912, opened 2024-07-12
Avoid non-uniform group contesting access to memory in group_of tests

steffenlarsen:
The group_of tests have all active items trying to write to the same index in the memory. For non-uniform groups this can be problematic as not all the results are the same, which means optimizations like vectorization may stumble. This commit splits the results into separate values that are in turn validated. This also allows more precise identification of error origin.


keryell (approved):
Thanks!

PR #913, opened 2024-07-12
Generate non-uniform group tests for group type

steffenlarsen:
This commit changes most non-uniform group tests to generate separate cases for each variant of the group types we are testing with. This was done because the current structure will skip an entire test case if just one of the group types are not supported by the device, even if the others are. By splitting them, only the test case with the unsupported group type will be skipped.


steffenlarsen:
https://github.com/KhronosGroup/SYCL-CTS/pull/914 replaces this.

PR #919, opened 2024-07-26
Template and split marray tests to reduce memory consumption

steffenlarsen:
Some of the test categories in the suite require a lot of memory when compiling with DPC++. One of these is the marray category.

This commit splits the marray test into multiple categories and templates most of the test source files on the tested types.


keryell (approved):
What a PR!
Thanks.

PR #923, opened 2024-08-08
Build test_conformance instead of test_all in CI

steffenlarsen:
In our CI testing, the system is running out of disk space while building the `test_all` executable. That said, the resulting executable is reundant, as it is a combined executable of all the other test category executables produced. This commit changes the target built by CI to `test_conformance` which only builds the test category executables.


bader (approved):

tomdeakin:
WG approval to merge.

PR #924, opened 2024-08-08
Remove test_all target

steffenlarsen:
After the conformance submission discourages the use of test_all and the CI has moved away from using test_all, all the benefit of it becomes convenience when wanting to run all tests. However, as it is an overall target and with the size of it increasing with the test suite, to general users it will just be an inconvenience, as it results in longer compilation times and higher resource requirements. Given this, this commit removes the test_all target.


psalz (approved):

keryell (approved):
Thanks!

PR #930, opened 2024-09-02
[NFC] Transfer host variable to sycl kernel to avoid using unsupported memory capabilities.

haonanya:
Test checks if memory capabilities are supported in the host code, and then early return in the final run time for unsupported capabilities. When use const host variable MemoryScope in the sycl kernel directly, aot compile will fail in the compile time if MemoryScope is not supported. So this patch transfers host variable to sycl kernel to avoid using unsupported memory capabilities.

### DIFF
--- a/tests/atomic_fence/atomic_fence.cpp
+++ b/tests/atomic_fence/atomic_fence.cpp
@@ -194,6 +194,7 @@ class run_atomic_fence {
                                    " and test_type = " + test_type_name)
                 .create()) {
       auto queue = once_per_unit::get_queue();
+      // Early return for unsupported memory order abd memory scope.
       if (!check_memory_order_scope_capabilities(queue, MemoryOrder,
                                                  MemoryScope, memory_order_name,
                                                  memory_scope_name)) {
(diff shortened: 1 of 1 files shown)


haonanya:
@keryell, @ProGTX, can you please take a look? Thanks very much.

tests/atomic_fence/atomic_fence.cpp
```diff
@@ -194,6 +194,7 @@ class run_atomic_fence {
                                    " and test_type = " + test_type_name)
                 .create()) {
       auto queue = once_per_unit::get_queue();
+      // Early return for unsupported memory order abd memory scope.
```
keryell, line 197:
```suggestion
      // Early return for unsupported memory order or memory scope.
```

tests/atomic_fence/atomic_fence.cpp
```diff
@@ -226,11 +227,32 @@ class run_atomic_fence {
         sycl::buffer<bool> res_buf(&res, sycl::range<1>(1));
         sycl::buffer<int> sync_buffer(&sync, sycl::range<1>(1));
         sycl::buffer<int> data_buffer(&data, sycl::range<1>(1));
+        // Using the const host variable MemoryScope in the kernel directly
+        // may cause compile fail for AOT build. We transfer MemoryScope to
```
keryell, line 231: Is this a SYCL limitation or a compiler bug?
In https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:language.restrictions.kernels:
> Variables with static storage duration that are odr-used inside a [device function](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#device-function), must be either const or constexpr, and must also be either zero-initialized or constant-initialized.
So, why not using directly `MemoryOrder` instead of intermediate `order_write` and so on?

haonanya: Hi, @keryell , not a SYCL limitation or a compiler bug. The early return in the host code occurs in the run time. For aot compile, use MemoryOrder directly may cause compile failure in compile time if MemoryOrder is not supported by device. So the test may fail in build stage and fail to generate a binary.

keryell (approved):
Thanks!

tests/atomic_fence/atomic_fence.cpp
gmlueck: Is the problem that the test tries to use an atomic fence order or scope that is not supported by the device?  If that is the case, the test should use `info::device::atomic_fence_order_capabilities` and `info::device::atomic_fence_scope_capabilities` to determine the supported order and scope values, and it should avoid submitting the kernel to the device if the device doesn't support that fence order or scope.

haonanya: @gmlueck, yes, there is early return https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/atomic_fence/atomic_fence.cpp#L197 in the original tests, the check works in the run time, but for aot compile it may compile fail in device build stage, so the test may fail to build final binary.

keryell (changes requested):
There are still some formatting issues to fix.
Thanks.

keryell:
@haonanya you can see the change to do about the formatting on https://github.com/KhronosGroup/SYCL-CTS/actions/runs/10674975860?pr=930
Could you apply it and remove the Draft as a blessed PR?

keryell:
I see.
But actually "compiling" for all the devices might just be the compiler workflow not crashing on the FPGA and just generating an empty kernel for this device for example, since the runtime will check at launch time either it can run or not because of an unsupported feature.
Does it mean you can close the PR then?

PR #935, opened 2024-09-09
Add missing tests for `multi_ptr` aliases for `access::address_space::generic`

victor-eds:
Add `(decorated_|raw_)generic_ptr` aliases definitions.

These aliases were previously referenced in the SYCL specification, but not defined and not tested in the CTS.

### DIFF
--- a/tests/multi_ptr/multi_ptr_explicit_conversions.h
+++ b/tests/multi_ptr/multi_ptr_explicit_conversions.h
@@ -184,17 +184,6 @@ void check_pointer_aliases(const std::string &type_name) {
               sycl::multi_ptr<T, sycl::access::address_space::private_space,
                               sycl::access::decorated::no>>);
     }
-// Not yet supported
-#if !SYCL_CTS_COMPILING_WITH_HIPSYCL && !SYCL_CTS_COMPILING_WITH_DPCPP
-    {
-      INFO("raw_generic_ptr");
-      STATIC_CHECK(
-          std::is_same_v<
-              sycl::raw_generic_ptr<T>,
-              sycl::multi_ptr<T, sycl::access::address_space::generic_space,
-                              sycl::access::decorated::no>>);
-    }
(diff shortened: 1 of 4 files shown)


victor-eds:
Does not compile as this adds a new alias to the spec.

psalz:
> Does not compile as this adds a new alias to the spec.

Please move the new checks into a [compile-time disabled test case](https://github.com/KhronosGroup/SYCL-CTS/tree/SYCL-2020/docs#compile-time-disabled-test-cases)!

victor-eds:
> Please move the new checks into a [compile-time disabled test case](https://github.com/KhronosGroup/SYCL-CTS/tree/SYCL-2020/docs#compile-time-disabled-test-cases)!

I went with a simple approach as that'd mean having conditionally compiled template functions to avoid using an undefined type and would overcomplicate tests too much to get same degree of coverage.

victor-eds:
Note this same approach is taken in other places like https://github.com/KhronosGroup/SYCL-CTS/blob/02c22520992d69cd2b5b94b441b506369edde5b7/tests/sub_group/sub_group_semantics.cpp#L119

victor-eds:
> > I went with a simple approach as that'd mean having conditionally compiled template functions to avoid using an undefined type and would overcomplicate tests too much to get same degree of coverage.
> 
> It might be simpler but it also hides the fact that an implementation does in fact not fully pass the test case. If you use the `DISABLED_FOR_TEST_CASE` method, the test will fail at runtime. AdaptiveCpp (formerly hipSYCL) does not strive to achieve conformance with 2020, so the code is still littered with `SYCL_CTS_COMPILING_WITH_HIPSYCL`. Intel on the other hand made an effort to get rid of all of these before doing their first conformance submission. Adding these back in now means that we have to remember to remove them before the next conformance submission is made.
> 
> Edit: I can see that it might be annoying to get it working with the whole typing boilerplate that is going on for those tests. Alternatively you could add an `#else` branch with an explicit call to `FAIL`!

I defined two additional tests. I still need conditional compilation as they would fail to compile otherwise.

keryell (approved):
Thanks!

psalz (approved):
LGTM!

PR #936, opened 2024-09-10
Remove mixed use of tabs and spaces

frasercrmck:
Tabs can been replaced with spaces as spaces were predominant. Formatting was done with clang-format on functions containing tabs - rather than the whole file - to try and reduce the diff.

The mixed tabs and spaces can cause issues in readability, for instance misleading indentation (some compilers pick up on this).

### DIFF
--- a/oclmath/reference_math.cpp
+++ b/oclmath/reference_math.cpp
@@ -7,13 +7,19 @@
 ******************************************************************/
 
 #if defined( _MSC_VER )
-# pragma warning( push )
-# pragma warning( disable : 4056 ) /* overflow in floating-point constant arithmetic                        */
-# pragma warning( disable : 4101 ) /* unreferenced local variable                                           */
-# pragma warning( disable : 4146 ) /* unary minus operator applied to unsigned type, result still unsigned  */
-# pragma warning( disable : 4244 ) /* conversion from 'double' to 'float', possible loss of data            */
-# pragma warning( disable : 4723 ) /* potential divide by 0                                                 */
-# pragma warning( disable : 4756 ) /* overflow in constant arithmetic                                       */
+#pragma warning(push)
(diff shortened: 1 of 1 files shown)


psalz:
AFAIK this file was copied from the [corresponding file in the OpenCL CTS](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/math_brute_force/reference_math.cpp), so I wonder if instead of reformatting it ourselves, we should just pull in the newer upstream version.

frasercrmck:
> AFAIK this file was copied from the [corresponding file in the OpenCL CTS](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/math_brute_force/reference_math.cpp), so I wonder if instead of reformatting it ourselves, we should just pull in the newer upstream version.

Yeah maybe that's a better way of doing it. Though we'd either have to examine the differences between the two files and/or trust that our testing is robust enough to catch the inevitable minor differences in reference maths that'll come with taht.

keryell (commented):
Yes, it sounds good to first fix this upstream and then check for the differences.

frasercrmck:
That said, since upstream's version has already been formatted, I wonder if it would make sense for us to format our copy first before pulling, so that any differences between the two files would become more apparent, especially for reviewers?

psalz:
Yes, although the OpenCL CTS uses a different clang-format configuration. I'm actually not sure where this file was taken from originally, because the [version in the first public commit of the OpenCL CTS](https://github.com/KhronosGroup/OpenCL-CTS/blob/2821bf132378a9222ffb5830d08a6de6be67c699/test_conformance/math_brute_force/reference_math.c) is already different (mostly whitespace changes as far as I can tell). Maybe we should start by updating to that version, and then see if we need any of the newer changes made since then.

PR #948, opened 2024-09-24
Split accessor tests to reduce memory usage

steffenlarsen:
The accessor tests are heavy w.r.t. memory resources during compilation. To reduce this overhead, the tests are split into separate categories and template the existing categories on the types.

Note that this is not intended to be a functional change to the tests.


keryell (approved):
This is like a PR!
That looks plausible.

PR #960, opened 2024-10-07
Implement tests for `SYCL_KHR_DEFAULT_CONTEXT` extension

0x12CC:
Implement the tests from https://github.com/KhronosGroup/SYCL-CTS/issues/944.

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       # Run on our default base branch to prime ccache for faster CI runs in PRs.
-      - SYCL-2020
+      - main
 
 jobs:
   # Pushing container images requires DockerHub credentials, provided as GitHub secrets.
--- a/test_plans/kernel.asciidoc
+++ b/test_plans/kernel.asciidoc
@@ -7,7 +7,7 @@ This is a test plan cover sycl::kernel functionality that is not covered by curr
 
 == Testing scope
 
-Tests intend to cover functions from 4.11.13. The kernel class that are not covered in https://github.com/KhronosGroup/SYCL-CTS/tree/SYCL-2020/tests/kernel
+Tests intend to cover functions from 4.11.13. The kernel class that are not covered in https://github.com/KhronosGroup/SYCL-CTS/tree/main/tests/kernel
 
 === Device coverage
 
--- a/test_plans/kernel_bundle.asciidoc
+++ b/test_plans/kernel_bundle.asciidoc
@@ -7,7 +7,7 @@ This is a test plan cover kernel_bundle functionality that is not covered by cur
 
 == Testing scope
 
-Tests intend to cover functions from 4.11. Kernel bundles that are not covered in https://github.com/KhronosGroup/SYCL-CTS/tree/SYCL-2020/tests/kernel_bundle
+Tests intend to cover functions from 4.11. Kernel bundles that are not covered in https://github.com/KhronosGroup/SYCL-CTS/tree/main/tests/kernel_bundle
 
 === Device coverage
 
@@ -116,7 +116,7 @@ Check that result of `get_kernel_ids`:
 
 === Tests for working with specialization constants
 
-Partially tested in https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/specialization_constants/specialization_constants_via_kernel_bundle.h
+Partially tested in https://github.com/KhronosGroup/SYCL-CTS/blob/main/tests/spec_constants/spec_constants_via_kernel_bundle.h
 
 There are two spec constant defined: `SpecName` and `OtherSpecName`.
 kernel_handler::get_specialization_constant<OtherSpecName>() shouldn't be used in any kernel in the application.
--- a/test_plans/queue_constructors.asciidoc
+++ b/test_plans/queue_constructors.asciidoc
@@ -7,7 +7,7 @@ This is a test plan for the for new constructors of class `sycl::queue` describe
 
 == Testing scope
 
-Tests intend to cover not covered constructors from 4.6.5.1. Queue interface that are not covered in https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/tests/queue/queue_constructors.cpp
+Tests intend to cover not covered constructors from 4.6.5.1. Queue interface that are not covered in https://github.com/KhronosGroup/SYCL-CTS/blob/main/tests/queue/queue_constructors.cpp
 
 === Device coverage
 
--- a/tests/invoke/parallel_for_simplifications.cpp
+++ b/tests/invoke/parallel_for_simplifications.cpp
@@ -21,8 +21,8 @@
 //  parallel_for({N}, some_kernel)
 //  parallel_for({N1, N2}, some_kernel)
 //  parallel_for({N1, N2, N3}, some_kernel)
-//  Test plan:
-https://github.com/KhronosGroup/SYCL-CTS/blob/SYCL-2020/test_plans/parallel_for_simplifications.asciidoc
+//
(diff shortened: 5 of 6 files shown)


tests/extension/khr_default_context/default_context.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace default_context::tests {
+
+#ifdef SYCL_KHR_DEFAULT_CONTEXT
+
+static void testDefaultContext() {
+  sycl::platform platform{};
+  sycl::context defaultContext = platform.khr_get_default_context();
+  CHECK(defaultContext.get_devices() == platform.get_devices());
+}
```
psalz, line 31: If we decide that trying to compile a KHR extension test for a SYCL implementation that does not implement said extension should result in an error, I'd prefer the tests to use a more idiomatic style, e.g.:

```suggestion
TEST_CASE("The default context contains all of the platform's devices") {
  sycl::platform platform{};
  sycl::context defaultContext = platform.khr_get_default_context();
  CHECK(defaultContext.get_devices() == platform.get_devices());
}
```

tests/extension/khr_default_context/default_context.cpp
```diff
@@ -0,0 +1,79 @@
...
+        sycl::queue{syclContext, syclDevice, propList}.get_context());
+  CHECK(syclContext ==
+        sycl::queue{syclContext, syclDevice, asyncHandler, propList}
+            .get_context());
+}
+
+#endif
+
+TEST_CASE("Test case for \"Default Context\" extension",
+          "[khr_default_context") {
```
psalz, line 70:
```suggestion
          "[khr_default_context]") {
```

psalz (commented):
Since this is the first KHR extension test, we should think about how we want to approach this in general. Currently you have both a CMake option to enable/disable compilation of tests for the extension, as well as compile-time guards that check whether the extension is available. I'm wondering whether we need both.

If we keep the former (which I think is a good idea, at least for now -- we'll see whether it scales), I'd argue we can get rid of the `#ifdef SYCL_KHR_DEFAULT_CONTEXT`. If someone wants to test the extension, they'd likely want to know that it isn't available during compilation, not at runtime.

gmlueck:
> If we keep the former (which I think is a good idea, at least for now -- we'll see whether it scales), I'd argue we can get rid of the `#ifdef SYCL_KHR_DEFAULT_CONTEXT`. If someone wants to test the extension, they'd likely want to know that it isn't available during compilation, not at runtime.

I think this makes sense.  If the CTS is run with `SYCL_CTS_ENABLE_EXT_KHR_DEFAULT_CONTEXT_TESTS`, then it seems like the test suite should fail if the implementation does not define `SYCL_KHR_DEFAULT_CONTEXT`.  Otherwise, an implementation could pass the CTS for this extension even without implementing it.

psalz:
> If I get rid of this macro, someone that wants to test an implementation that implements most features and not this one would need to enumerate all of the relevant CMake options. They wouldn't be able to use the SYCL_CTS_ENABLE_KHR_TESTS option since this test would fail to compile.

Is that really a common use-case though? Realistically, I would assume that it's mostly going to be Intel that runs the CTS for the time being. If it does get out of hand, we could always provide [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) for the supported implementations.

> Catch2 distinguishes passing test cases from skipped ones.

True, I'm just not sure whether having that skip message is worth sacrificing legibility of the code by either grouping the tests into guarded functions instead of using individual `TEST_CASE`s (like you did in this PR), or by littering `#ifdef`s all over the place.

psalz:
Of course we could also get fancy and do a configuration-time check ([`try_compile`](https://cmake.org/cmake/help/v3.31/command/try_compile.html)) to see whether an extension is available, and if not, compile a proxy translation unit that only prints the skipped message.

gmlueck:
> Catch2 distinguishes passing test cases from skipped ones. Here's the output for this test when the macro is not defined:
> 
> ```
> SYCL-CTS/tests/extension/khr_default_context/default_context.cpp:72: SKIPPED:
> explicitly with message:
>   SYCL_KHR_DEFAULT_CONTEXT is not defined
> 
> ================================================================================
> test cases: 1 | 1 skipped
> assertions: - none -
> ```

This requires a human to examine the test output, though.  I think it would be better for the test to fail (return a non-zero exit status) if the macro isn't defined.  This will make the tests easier to run in a CI system, which just tests the exit status.

I think it would also be good for the Khronos workgroup when a vendor submits a conformance report.  The WG won't want to look through the test output to see if there are SKIPPED messages.  It will be easier for the WG to just rely on the fact that the CTS passed when determining a vendor's conformance.

> If you both still agree that the SYCL_KHR_DEFAULT_CONTEXT macro guard is not needed then I'll remove it from the test.

I think we should go a step further.  We should remove the macro guard, and also add a test that explicitly tests that the macro is defined.  Part of an extension's API is that the macro is supposed to be set, so the CTS should check this.

> Also, if we do choose to remove it, should we remove the macros from the other extension tests (in a different PR)?

The other extensions are all "oneapi" extensions, right?  Putting my Khronos hat on ... these extensions exist in the Khronos CTS as a convenience to DPC++, so DPC++ can decide what works best for them.

Putting my DPC++ hat on ... I think this probably does make sense.  We've had situations in the past where we forgot to define the macro for an extension.  Restructuring the tests like this would catch these errors.  Of course, we'd need to have some cmake option to control when the oneapi tests are run.  Do we have something like `SYCL_CTS_ENABLE_ONEAPI_TESTS` already?

> If I get rid of this macro, someone that wants to test an implementation that implements most features and not this one would need to enumerate all of the relevant CMake options. They wouldn't be able to use the SYCL_CTS_ENABLE_KHR_TESTS option since this test would fail to compile.

One more comment about this.  When vendors submit official conformance results, I think we do *not* want them to use `SYCL_CTS_ENABLE_KHR_TESTS` because it's not obvious which extensions are being tested.  Wouldn't we want official runs to pass a list of `SYCL_CTS_ENABLE_KHR_DEFAULT_CONTEXT_TESTS`, ... options, so we know exactly which extensions the vendor is claiming conformance?

tests/extension/khr_default_context/default_context.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "../../common/common.h"
+
+namespace default_context::tests {
+
+TEST_CASE(
+    "the default context extension defines the SYCL_KHR_DEFAULT_CONTEXT macro",
+    "[khr_default_context]") {
+#ifndef SYCL_KHR_DEFAULT_CONTEXT
+  static_assert(false, "SYCL_KHR_DEFAULT_CONTEXT is not defined");
```
psalz, line 29: I think this is a good solution!

tests/extension/khr_default_context/default_context.cpp
```diff
@@ -0,0 +1,75 @@
...
+          "[khr_default_context]") {
+  const sycl::property_list& propList = {};
+  cts_async_handler asyncHandler;
+  const auto& deviceSelector = sycl::default_selector_v;
+  sycl::device syclDevice;
+  sycl::context syclContext;
+  sycl::context defaultContext = sycl::platform{}.khr_get_default_context();
+
+  // Check that a default-constructed context is not the default context.
+  CHECK(syclContext != defaultContext);
```
psalz, line 50: Minor: This could be pulled out into it's own test case IMO.

psalz (approved):

gmlueck (approved):
I think this looks great!

If you are on a roll with tests for default context behavior, we also need tests for #943.  Note that those tests are for the core spec, so they are not tied to the extension.

tomdeakin:
Thanks! Waiting for further feedback from AdaptiveCpp.

illuhad (approved):

tomdeakin:
WG approved to merge.

PR #963, opened 2024-10-16
Gather linker flags to prevent /link break linking stage

jiezzhang:
`-DCMAKE_EXE_LINKER_FLAGS="/link /stack:2097152"` will break linking stage on Windows because all options following "/link" are passed to linker. And "-o" passed to linker can't be recognized by linker and trigger an error


jiezzhang:
CI fail because the job reaches timelimit.

bader (approved):

PR #964, opened 2024-10-16
Fix accessor using unsuitable property

KornevNikita:
Accessor shouldn't use buffer::property.
It looks like there is no real need for `properties` to contain any property, as the test just checks all possible constructors.

### DIFF
--- a/tests/accessor_legacy/accessor_constructors_buffer_utility.h
+++ b/tests/accessor_legacy/accessor_constructors_buffer_utility.h
@@ -170,8 +170,7 @@ public:
       using verifier = check_accessor_constructor_buffer<accTag, property_list>;
 
       auto context = util::get_cts_object::context();
-      property_list properties {
-          sycl::property::buffer::context_bound(context)};
+      property_list properties;
 
       {
         const auto constructorName = usesHander ?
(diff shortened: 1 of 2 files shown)


KornevNikita:
For some reason the runner is reaching the time limit.

AlexeySachkov:
> For some reason the runner is reaching the time limit.

It seems like CI attempts to build *all* the CTS tests, which takes a while. Not sure why, because `ccache` should be in place and in general we should probably only re-build affected suites, not all of them

KornevNikita:
@bader hi, how should we proceed here?

bader:
> @bader hi, how should we proceed here?

Pass CI + get approval from non-Intel implementer => merge PR.

Are you asking something else?

KornevNikita:
> > @bader hi, how should we proceed here?
> 
> Pass CI + get approval from non-Intel implementer => merge PR.
> 
> Are you asking something else?

Yes:) The patch can't pass CI. Looks like it builds the whole project and reaches the timelimit.

bader:
> Yes:) The patch can't pass CI. Looks like it builds the whole project and reaches the timelimit.

I see at least two options to investigate:
- figure out why ccache doesn't work for your PR. NOTE: it works here - https://github.com/KhronosGroup/SYCL-CTS/pull/966.
- speed up the build of the whole project to fit into GitHub Actions limits.

keryell (approved):
Thanks.

KornevNikita:
@bader can we treat the current pre-commit status as passed? Looks like the build was successful, the workflow just failed the "Post Set up ccache" step (I guess due to time limit?). Can we merge this?

bader:
I hope https://github.com/KhronosGroup/SYCL-CTS/pull/982 will fix problems with `ccache`.

PR #966, opened 2024-10-22
Split conformant product fields in submission template

steffenlarsen:
This commit splits one of the fields in the "Conformant Product"  and "Conformance submission details" sections of the submission_details_template file. This brings the fields closer in line with the conformant products page.

### DIFF
--- a/conformance/submission_details_template.txt
+++ b/conformance/submission_details_template.txt
@@ -58,10 +58,12 @@ Target SYCL device:
 # and device combinations.
 #
 SYCL implementation version(s):
+Device(s):
+Platform:
+Backend:
 Host architecture(s):
 Host operating system(s) and version(s):
 Host toolchains(s) and version(s):
-SYCL backend/platform/device combinations(s):
 
 # Environment requirements.
 #
--- a/conformance/submission_details_template.txt
+++ b/conformance/submission_details_template.txt
@@ -59,8 +59,8 @@ Target SYCL device:
 #
 SYCL implementation version(s):
 Device(s):
-Platform:
-Backend:
+Platform(s):
+Backend(s):
 Host architecture(s):
 Host operating system(s) and version(s):
 Host toolchains(s) and version(s):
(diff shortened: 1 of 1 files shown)


conformance/submission_details_template.txt
```diff
@@ -58,10 +58,12 @@ Target SYCL device:
 # and device combinations.
 #
 SYCL implementation version(s):
+Device(s):
+Platform:
```
AerialMantis, line 62:
```suggestion
Platform(s):
```

conformance/submission_details_template.txt
```diff
@@ -58,10 +58,12 @@ Target SYCL device:
...
+Backend:
```
AerialMantis, line 63:
```suggestion
Backend(s):
```

conformance/submission_details_template.txt
AerialMantis: Minor tweak, I'd add a `(s)` as the process allows for a submission which supports other platforms//backends.

AerialMantis (approved):
Looks good!

conformance/submission_details_template.txt
```diff
@@ -58,10 +58,12 @@ Target SYCL device:
 # and device combinations.
 #
 SYCL implementation version(s):
+Device(s):
+Platform(s):
```
gmlueck, line 62: Can we remove "Platform" here, and also above under "Conformance submission details"?  This is just redundant information because it's the same as "Backend".

gmlueck: WG proposes that we could combine these two lines as:

> Backend / platform(s):

gmlueck (approved):

KornevNikita (commented):
Just wondering, should we also update the block above?
```
Target SYCL backend:
Target SYCL platform:
Target SYCL device:
```
in `# Conformance submission details.`

gmlueck:
It does seem like it would be clearer if the "Conformance submission details" section had the same format as "Conformant products".  We could change "Conformance submission details" to look like:

```
SYCL implementation version:
Device:
Backend / platform:
Host architecture:
Host operating system and version:
Host toolchain and version:
```

(I.e. the same as "Conformant products", but the entries are singular because you submit a conformance report based on a single run of the CTS.)

keryell (approved):
Thanks.

PR #967, opened 2024-10-22
Remove non-informative "Supported optional features" section

steffenlarsen:
The "Supported optional features" in the conformance submission template does not add any useful information that is not already covered by the rest of the document, so this patch proposes its removal.

### DIFF
--- a/conformance/submission_details_template.txt
+++ b/conformance/submission_details_template.txt
@@ -75,13 +75,6 @@ Environment requirements:
 # List of KHR extension supported by the SYCL implementation.
 # KHR extensions:
 
-# Supported optional features.
-#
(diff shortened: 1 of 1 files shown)


gmlueck (approved):
I think this makes sense.

I suspect this got copied from the OpenCL conformance submission form.  It did make sense for OpenCL because OpenCL has "optional features", which a vendor may or may not implement.  We don't have an exact analog in SYCL.

AerialMantis (approved):

TApplencourt (approved):

nliber (approved):

illuhad (approved):

keryell (approved):
Thanks.

PR #983, opened 2024-11-12
[range_api] Fix bitwise shift rhs out-of-range

wenju-he:
In function test_range_kernels, range[2] is 64 and range_two[2] is 128. It is undefined behavior to use either of them as rhs of bitwise shift operator because the rhs is not smaller than the number of bits in range value type. This PR divides rhs by 4 so that it is smaller than 64.

### DIFF
--- a/tests/range/range_api.cpp
+++ b/tests/range/range_api.cpp
@@ -50,6 +50,9 @@ void test_range_kernels(
   const sycl::range<dims> range_two_const(range_two);
   const sycl::range<dims> range_const(range);
 
+  sycl::range<dims> range_quarter(range / 4);
+  const sycl::range<dims> range_quarter_const(range_quarter);
+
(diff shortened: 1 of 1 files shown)


keryell:
> In function test_range_kernels, range[2] is 64 and range_two[2] is 128. It is undefined behavior to use either of them as rhs of bitwise shift operator because the rhs is not smaller than the number of bits in range value type. This PR divides rhs by 4 so that it is smaller than 64.

Great catch! Could you copy your analysis as a comment close to where the ranges are defines?
Otherwise we will forget about it and its rationale.

wenju-he:
> Great catch! Could you copy your analysis as a comment close to where the ranges are defines? Otherwise we will forget about it and its rationale.

done.

keryell (approved):
Thanks!

PR #984, opened 2024-11-13
Fix MSVC build.

hvdijk:
In commit 33be1b17, I updated rounding_mode.* to their current OpenCL-CTS versions without noticing that the old versions had an intentional deviation from OpenCL-CTS: a call to vlog_error had been commented out, and the headers that were needed only for that were also left out. This commit reapplies those changes.

### DIFF
--- a/oclmath/rounding_mode.cpp
+++ b/oclmath/rounding_mode.cpp
@@ -130,8 +130,6 @@ RoundingMode set_round(RoundingMode r, Type outType) {
 
   int err = _controlfp_s(&oldRound, 0, 0);  // get rounding mode into oldRound
   if (err) {
-    vlog_error("\t\tERROR: -- cannot get rounding mode in %s:%d\n", __FILE__,
-               __LINE__);
     return kDefaultRoundingMode;  // what else never happens
   }
 
--- a/oclmath/rounding_mode.h
+++ b/oclmath/rounding_mode.h
@@ -18,11 +18,6 @@
 
 #include "compat.h"
 
-#if (defined(_WIN32) && defined(_MSC_VER))
-#include "errorHelpers.h"
-#include "testHarness.h"
-#endif
-
 typedef enum {
   kDefaultRoundingMode = 0,
   kRoundToNearestEven,


hvdijk:
@bader Testing this on Visual Studio 2022, I could reproduce the error regarding the missing headers and can see `rounding_mode.cpp` compiles with this change, but I am unable to link test binaries because of other unrelated errors which may be something in my setup. Are you able to test whether MSVC builds actually work properly on setups where they previously worked?

hvdijk:
The CI shows no error other than "Error: The operation was canceled" after six hours, which I presume is a timeout? Should I do something to get the CI to re-run?

bader:
> The CI shows no error other than "Error: The operation was canceled" after six hours, which I presume is a timeout? Should I do something to get the CI to re-run?

I've updated your branch, which fixed ccache configuration. It should pass now. No actions should be needed from your side.

bader:
> @bader Testing this on Visual Studio 2022, I could reproduce the error regarding the missing headers and can see `rounding_mode.cpp` compiles with this change, but I am unable to link test binaries because of other unrelated errors which may be something in my setup. Are you able to test whether MSVC builds actually work properly on setups where they previously worked?

@hvdijk, thanks! @jiezzhang confirmed that this patch fixes the problem.

oclmath/rounding_mode.cpp
```diff
@@ -130,8 +130,8 @@ RoundingMode set_round(RoundingMode r, Type outType) {
 
   int err = _controlfp_s(&oldRound, 0, 0);  // get rounding mode into oldRound
   if (err) {
-    vlog_error("\t\tERROR: -- cannot get rounding mode in %s:%d\n", __FILE__,
-               __LINE__);
+    // vlog_error("\t\tERROR: -- cannot get rounding mode in %s:%d\n", __FILE__,
+    //            __LINE__);
```
bader, line 134: To be consistent, we can just remove this code as we do for header includes. What do you think? I suppose `vlog_error` is defined by one of the includes.

hvdijk: I went with commenting it out because that was the change that had previously been made to the older version of this file:

https://github.com/KhronosGroup/SYCL-CTS/blob/8623df08565e70e711e48d3a13cc87c1bcfcd518/oclmath/rounding_mode.cpp#L125

But I have no preference here -- if you prefer I remove it, I'm happy to.

hvdijk: Done.

keryell (approved):
Thanks!

PR #989, opened 2024-11-18
Add SYCL version check

dklochkov-emb:
Check of SYCL version was added according to agreement, see details:
https://github.com/KhronosGroup/SYCL-Docs/pull/634


tests/header/header_test.cpp
```diff
@@ -45,6 +45,7 @@ class TEST_NAME : public util::test_base {
 #define TEST_FAIL
     log.note("SYCL_LANGUAGE_VERSION not present");
 #else
+    static_assert(SYCL_LANGUAGE_VERSION, "202012L");
```
gmlueck, line 48: Shouldn't this be something like:

```
static_assert(SYCL_LANGUAGE_VERSION == 202012L)
```

Followed by a check that the type is really `long`.

tests/header/header_test.cpp
```diff
@@ -45,6 +45,7 @@ class TEST_NAME : public util::test_base {
...
     log.note("SYCL_LANGUAGE_VERSION = %d", static_cast<int>(SYCL_LANGUAGE_VERSION));
```
gmlueck, line 49: Not sure if this log message is still useful since we are now checking for a specific value.

tests/header/header_test.cpp
gmlueck: > Followed by a check that the type is really `long`.

Something like:

```
static_assert(std::is_same_v<decltype(SYCL_LANGUAGE_VERSION), long>);
```

tests/header/header_test.cpp
dklochkov-emb: As for me, we can remove this log message due to static_assert.

gmlueck (approved):

dklochkov-emb:
Ping

keryell (approved):
Thanks.
By the way, it is better to write SYCL in uppercase everywhere, even in the commit messages.

bader:
@dklochkov-emb, please, address [this feedback](https://github.com/KhronosGroup/SYCL-CTS/pull/989?new_mergebox=true#pullrequestreview-2456295342) by fixing commit message. Alternatively, I can do it myself by squashing all commits with my own commit message.

Please, fix CI failures.

tomdeakin:
Please bump the DPC++ version in the CI to ensure this test doesn't fail.
Resolve AdaptiveCpp either by updating implementation, or disabling this test.

psalz:
I realize this is not your fault, but the whole test setup used here is terribly outdated. The idiomatic way of writing this test would be something like

```c++
TEST_CASE("The implementation defines the correct SYCL_LANGUAGE_VERSION macro") {
#ifndef SYCL_LANGUAGE_VERSION
    FAIL("SYCL_LANGUAGE_VERSION is not defined");
#else
    STATIC_REQUIRE(std::is_same_v<decltype(SYCL_LANGUAGE_VERSION), long>);
    STATIC_REQUIRE(SYCL_LANGUAGE_VERSION == 202012L);
#endif
}
```

illuhad:
I've just aligned `SYCL_LANGUAGE_VERSION` in upstream AdaptiveCpp: https://github.com/AdaptiveCpp/AdaptiveCpp/pull/1630
So if we update the CI image, the test should also work in AdaptiveCpp.

AlexeySachkov:
> Edit: Actually, why does this check have to be a static assertion? Can't we just fail at runtime?

The earlier you get an error, the better. When you work on a SYCL implementation and static assert fails, you immediately know where and what failed: most likely some of interfaces don't have correct signatures (like type of a macro-defined constant here).

However, when you run CTS, you probably also want to have a little check saying "yes, interfaces are correct" simply for having all checks reported at the same time and be a part of a final CTS validation report.

`check2` allows to fully offload `STATIC_REQUIRE` to runtime phase via `CATCH_CONFIG_RUNTIME_STATIC_REQUIRE`. Perhaps that could be a default mode in which we run CI and do certification runs. Development could stay with compile-time reporting mode for shorter feedback loop.

psalz:
We should wait for #1013 to be merged and rebase/backmerge to see if this still compiles!

PR #992, opened 2024-12-03
Update checks for `sycl::info::device::preferred_interop_user_sync` and  `sycl::info::device::profile` in `device_info` test

DYNIO-INTEL:
Based on specification https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_device_information_descriptors 
 `sycl::info::device::preferred_interop_user_sync` and ``sycl::info::device::profile` info descriptors are supposed to throw an exception when the backend is not OpenCL.

### DIFF
--- a/tests/common/common.h
+++ b/tests/common/common.h
@@ -36,6 +36,7 @@
 #include "get_cts_object.h"
 #include "macros.h"
 #include "string_makers.h"
+#include "sycl/backend_types.hpp"
 #include "value_operations.h"
 
 #include <algorithm>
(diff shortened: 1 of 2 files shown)


gmlueck (approved):

tests/common/common.h
```diff
@@ -158,6 +158,22 @@ void check_get_info_param(const ObjectT& object) {
   auto returnValue = object.template get_info<InfoDesc>();
   check_return_type<ReturnT>(returnValue, "object::get_info()");
 }
+/**
```
bader, line 161: Let's keep existing style i.e. separate function templates by empty line.
```suggestion
}

/**
```

bader (commented):
It looks like `sycl::make_error_code` isn't supported by `hipSYCL`, so we probably should disable checks or the test case for that implementation.

DYNIO-INTEL:
> It looks like `sycl::make_error_code` isn't supported by `hipSYCL`, so we probably should disable checks or the test case for that implementation.

I updated code.  Could you look if it's okay?

bader:
@KhronosGroup/sycl-cts-reviewers, we need one approve from a person not affiliated with Intel. Could you take a look, please?

tomdeakin (approved):

PR #1002, opened 2024-12-05
Drop support for deprecated Intel_SYCL implementation

psalz:
This removes the long deprecated `Intel_SYCL` option for `SYCL_IMPLEMENTATION`, and aborts configuration if the latter is not set. I pulled this change out of #874 (which I'm finally revisiting).


bader (approved):

keryell (approved):
Thanks!

PR #1009, opened 2024-12-16
Remove CTS test for is_property_key trait

lbushi25:
This PR removes the test for the `is_property_key` trait because it has been removed from the spec of the corresponding extension in this [PR](https://github.com/intel/llvm/pull/16143).


lbushi25:
> If Intel are happy to merge this extension CTS ,please merge.

@bader Looks like this is good to merge!

PR #1011, opened 2024-12-17
`buffer::set_final_data` accepts `weak_ptr<T>`, not `<T[]>`

fknorr:
See [[4.7.2.1]](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_buffer_interface): 

> The `finalData` points to where the outcome of all the buffer processing is going to be copied to at destruction time, if the buffer was involved with a write accessor.
>
>Destination can be either an output iterator or a `std::weak_ptr<T>`. 
>
>Note that a raw pointer is a special case of output iterator and thus defines the host memory to which the result is to be copied.

Not sure if this was different in an earlier spec revision.

### DIFF
--- a/tests/buffer/buffer_storage_common.h
+++ b/tests/buffer/buffer_storage_common.h
@@ -63,21 +63,20 @@ class buffer_storage_test {
 
     // Case 3 - Weak pointer
     std::shared_ptr<T[]> data_shared_ptr(new T[size]);
-    std::weak_ptr<T[]> data_final3 = data_shared_ptr;
+    // construct an aliasing shared_ptr to get rid of the []
+    std::shared_ptr<T> data_shared_ptr_alias(data_shared_ptr,
+                                             data_shared_ptr.get());
+    std::weak_ptr<T> data_final3 = data_shared_ptr_alias;
 
-    // Case 4 - Shared pointer
-    std::shared_ptr<T[]> data_final4(new T[size]);
-
-    // Case 5 - Vector data
+    // Case 4 - Vector iterator
     std::vector<T> data_vector;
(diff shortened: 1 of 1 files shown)


TApplencourt (approved):

TApplencourt:
Agree. 

For example, for `buffer` in we have a special overload for `T[]`. ` buffer(const std::shared_ptr<T[]>&` . We don't have such for `final_data` as you noticed.

tomdeakin:
One more reviewer please (total 2 for this please)

tests/buffer/buffer_storage_common.h
```diff
@@ -63,21 +63,19 @@ class buffer_storage_test {
 
     // Case 3 - Weak pointer
     std::shared_ptr<T[]> data_shared_ptr(new T[size]);
-    std::weak_ptr<T[]> data_final3 = data_shared_ptr;
+    std::shared_ptr<T> data_shared_ptr_alias(data_shared_ptr,
```
keryell, line 66: Why doing this, since you initialize it also with the pointer from the shared pointer handling the ownership? :thinking: 
OK, after more thoughts you do this to have the same shared pointer but returning a pointer of the array decay.
You could add a comment. :smile:

keryell (commented):
I wonder whether this should be a specification fix to accept smart pointers on `T[]` also.

tests/buffer/buffer_storage_common.h
fknorr: Added a comment!

tomdeakin:
Thanks for finding the bug @fknorr !

PR #1012, opened 2024-12-19
[CI] Uplift DPC++ compiler version

bader:


bader:
@illuhad, I'm trying to uplift AdaptiveCpp version in Khronos SYCL-CTS CI, but installation fails due to missing dependency on LLVM. What would you recommend to fix this issue? 
I see that AdaptiveCpp build system supports [multiple profiles](https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/installing.md#cmake-options-to-configure-the-adaptivecpp-build). If I get it right some of these profiles remove the requirement to install LLVM libraries. Should we configure the build to use of these profiles or install LLVM?

illuhad:
@bader The AdaptiveCpp main compiler is the generic JIT compiler (`--acpp-targets=generic`). Since this is our main compiler, it would be good at least long term to have it in CTS CI. That will indeed require LLVM, but installing the necessary packages from `apt.llvm.org` should be sufficient. This is what we do in our CI: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/ba3fb96fbeb4a00929f4f93b57f0831e44ed903a/.github/workflows/linux.yml#L52

Now I don't know what is possible in the CTS CI image, or if there are any challenges around this. Perhaps @psalz knows more?

Otherwise it is indeed possible to install a library-only configuration using `-DACPP_COMPILER_FEATURE_PROFILE=none`, however, this is *not* the configuration that we recommend to end users, or the configuration that we would work on for eventual spec conformance.

docker/dpcpp/Dockerfile
```diff
@@ -4,6 +4,9 @@ ARG IMPL_VERSION
 
 FROM ghcr.io/intel/llvm/sycl_ubuntu2204_nightly:no-drivers-$IMPL_VERSION
 
+# Make sure that apt is executed with root previlegies
+USER root
```
psalz, line 8: Why is this suddenly required?

docker/hipsycl/Dockerfile
```diff
@@ -6,6 +6,19 @@ FROM khronosgroup/sycl-cts-ci:common
...
+    lsb-release \
+    software-properties-common \
+    gnupg
+
+RUN wget https://apt.llvm.org/llvm.sh && \
+    chmod +x llvm.sh && \
+    ./llvm.sh 18 && \
+    apt update && \
+    apt install -y --no-install-recommends \
+    libclang-18-dev clang-tools-18 libomp-18-dev
```
psalz, line 20: After the switch to 24.04 as the base image in #874 we can just install LLVM 18 though `apt`.

psalz:
#872 already updates the ACpp image to a revision from last week; why do we need another PR for this?

docker/dpcpp/Dockerfile
bader: DPC++ recently changed the default user from `root` to unprivileged one. 
https://github.com/intel/llvm/pull/16411

psalz (approved):

PR #1013, opened 2024-12-27
Fix AdaptiveCpp CI filter

psalz:
The bad news: There was an unfortunate interaction between the CI filter generation script and Ninja on the system I was running it on, leading to #874 essentially excluding all test categories for AdaptiveCpp. I only recently discovered this while looking at the CI runs for #872, for which AdaptiveCpp compiled the CTS in 2 seconds...

The good news: There's actually quite a lot categories that in fact do compile now (over 50%)!

cc @illuhad


illuhad (approved):

tomdeakin:
WG approves to merge.

PR #1018, opened 2025-01-15
[CI] Fix DPCPP container creation

bader:
Intel has restricted access to Docker containers with pre-installed DPC++ compiler.
Let's build the containers here using pre-built DPC++ binaries still available on GitHub.


bader:
Note for reviewers: I merged this PR before it was approved to fix post-commit job. Feel free to leave comments here. I will address them in a separate pull request.

PR #1021, opened 2025-01-20
Add a test plan for sycl_khr_free_function_commands

KornevNikita:
Spec: https://github.com/KhronosGroup/SYCL-Docs/pull/644
In general this is a copy-paste of test_plans/oneapi/enqueue_functions.asciidoc

### DIFF
--- a/test_plans/khr/free_function_commands.asciidoc
+++ b/test_plans/khr/free_function_commands.asciidoc
@@ -16,7 +16,7 @@ is selected on the CTS command line.
 
 === Feature test macro
 
-All of the tests should use `#ifdef SYCL_EXT_KHR_FREE_FUNCTION_COMMANDS` so they can be skipped
+All of the tests should use `#ifdef SYCL_KHR_FREE_FUNCTION_COMMANDS` so they can be skipped
 if feature is not supported.
 
 == Tests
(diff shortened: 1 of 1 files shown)


keryell (approved):
That looks good.

test_plans/khr/free_function_commands.asciidoc
```diff
@@ -0,0 +1,413 @@
...
+
+void mem_advise(sycl::queue q, void* ptr, size_t numBytes, int advice);
+
+}
+```
+
+=== Command and event barriers
+
+These tests should use `#ifdef SYCL_EXT_ONEAPI_ENQUEUE_BARRIER` so they can be skipped
+if feature is not supported. For each barrier function, enqueue some commands before and after enqueuing the barrier. Assert that the commands enqueued after the barrier do not execute until those enqueued before the barrier have completed. The barrier functions are the following:
```
Pennycook, line 389: The KHR doesn't have a dependency on the oneAPI extension.  The `command_barrier` and `event_barrier` functions are expected to be supported by any device that supports the KHR.

```suggestion
For each barrier function, enqueue some commands before and after enqueuing the barrier. Assert that the commands enqueued after the barrier do not execute until those enqueued before the barrier have completed. The barrier functions are the following:
```

test_plans/khr/free_function_commands.asciidoc
```diff
@@ -0,0 +1,413 @@
...
+== Testing scope
+
+=== Device coverage
+
+All of the tests described below are performed only on the default device that
+is selected on the CTS command line.
+
+=== Feature test macro
+
+All of the tests should use `#ifdef SYCL_EXT_KHR_FREE_FUNCTION_COMMANDS` so they can be skipped
```
Pennycook, line 19: We drop the `_EXT_` from KHR extensions, and just use `_KHR_`.

```suggestion
All of the tests should use `#ifdef SYCL_KHR_FREE_FUNCTION_COMMANDS` so they can be skipped
```

Pennycook (changes requested):
This mostly looks good.  I just spotted two minor things.

test_plans/khr/free_function_commands.asciidoc
KornevNikita: Thanks, copy-paste heritage. Updated.

Pennycook (approved):
I'm not sure if my approval will count or if @gmlueck also needs to approve, so pinging him just in case...

gmlueck:
> I'm not sure if my approval will count or if @gmlueck also needs to approve, so pinging him just in case...

I think @Pennycook 's approval counts.

TApplencourt (approved):

tomdeakin:
WG approved to merge

PR #1046, opened 2025-02-17
Add test plan for `sycl_ext_oneapi_num_compute_units`

dm-vodopyanov:


Pennycook:
> @Pennycook: You are the author of this extension, so it might be better for you to review the test plan. Can you think of any better tests beyond what is proposed here?

I don't think we can do better than this, if we're limited to selecting between "pass" and "fail" for the conformance tests.

If we had some middle option like "Warning: Your implementation is technically correct, but your behavior is potentially surprising." then we could probably do better for some specific device/backend combinations.  For example, if we could detect that the conformance tests were being run on a CUDA backend we could warn if the number of compute units does not match the number of SMs, or warn if it does not match the number of Xe Cores for Level Zero.

Something like this might be a good way to try and drive consistency across implementations, even when the specification doesn't mandate a specific behavior.  Such feedback might be helpful for implementers, too -- returning "1" from this query for every device/backend is technically correct and would be sufficient for conformance, but it would probably lead to a lot of bug reports.

Pennycook (approved):

tomdeakin:
As an intel extension, please merge when intel are happy.

PR #1055, opened 2025-02-24
Remove `vec::vector_t` tests

aelovikov-intel:
Following the specification change in https://github.com/KhronosGroup/SYCL-Docs/pull/676.


keryell (approved):
Thanks!

PR #1061, opened 2025-02-28
Don't call `accessor::get_pointer()` from outside command when not allowed

aelovikov-intel:
Both `target::constant_buffer` and `target::local` have the restriction that `get_pointer()` may only be called from within a command.


keryell (approved):
Good catch!

PR #1062, opened 2025-03-04
Add flag for unfinished KHR extensions

dyniols:
The purpose of this PR is to add flag which enables testing for unfinished KHR extensions.

### DIFF
--- a/cmake/FindDPCPP.cmake
+++ b/cmake/FindDPCPP.cmake
@@ -73,7 +73,7 @@ if(DPCPP_ENABLE_PREVIEW_CHANGES)
 endif()
 
 cmake_dependent_option(DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS
-  "Enable unfinished KHR extensions" ON
+  "Enable unfinished KHR extensions for DPC++ compiler" ON
   "SYCL_CTS_ENABLE_KHR_TESTS" OFF)
 if(DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS)
     add_definitions(-D__DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS)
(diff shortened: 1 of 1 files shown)


bader:
@dyniols, could you please clarify what "unfinished KHR extension" is? Is the spec not ratified? Is the DPC++ implementation of the extension is not 100% ready?

dyniols:
> @dyniols, could you please clarify what "unfinished KHR extension" is? Is the spec not ratified? Is the DPC++ implementation of the extension is not 100% ready?

In my understanding, this means that the DPC++ implementation of the extension isn't fully complete. To begin testing it with the CTS, we need to enable it by defining a macro `__DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS` .

bader:
> > @dyniols, could you please clarify what "unfinished KHR extension" is? Is the spec not ratified? Is the DPC++ implementation of the extension is not 100% ready?
> 
> In my understanding, this means that the DPC++ implementation of the extension isn't fully complete. To begin testing it with the CTS, we need to enable it by defining a macro `__DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS`

I think we already have infrastructure to control which tests are enabled/disabled per implementation. We have test filters to disable the whole test and compile time macros to disable parts of the tests. Can you use them to get what you want?

steffenlarsen:
The intention of the `__DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS` in DPC++ is to allow us to preemptively merge KHR extension implementations before they have been ratified. Doing so means we can test the proposals during ratification, but avoids users inadvertently thinking the unratified extension interfaces are final. As such, the intention is not to filter testing but rather to control what functional version of DPC++ we test with, similar to how we control the use of preview functionality.

bader (approved):
I found https://github.com/intel/llvm/blob/sycl/sycl/doc/developer/KHRExtensions.md document, which clarified some of my questions regarding unfinished KHR extensions implementation.

PR #1067, opened 2025-03-10
[CI][NFC] Update ccache workaround comments

bader:
DPC++ compiler fixed the problem blocking ccache functionality reported in https://github.com/intel/llvm/issues/5260.

Unfortunately, we can't remove the workaround due to the overhead added by the default "preprocessor mode". Rebuilding the cache in "preprocessor mode" on GitHub runners takes more than 6h, which is a limit for our plan.

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -123,7 +123,7 @@ jobs:
             version: 27ca7a05e65d24c784ba831225d0a53341719590
     env:
       container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
-      parallel-build-jobs: 2
+      parallel-build-jobs: 3
     container:
       image: khronosgroup/sycl-cts-ci:${{ matrix.sycl-impl }}-${{ matrix.version }}
     steps:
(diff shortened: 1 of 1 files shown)


keryell:
Why does this break with the other compilers? :thinking:

bader:
> Why does this break with the other compilers? 🤔

Internet says that exist code 137 means that application consumed too much memory. I guess the memory AdaptiveCPP compiler needs to compile/link 3 different tests in parallel exceeds the limits of GitHub runner. 😞

keryell (approved):
Looks good but need to have a different setting for each compiler?

bader:
> Looks good but need to have a different setting for each compiler?

I'm going to test new DPC++ compiler, which I expect to pass the pre-commit checks with 2 threads. Then we can use unified settings across all implementations.

.github/workflows/cts_ci.yml
```diff
@@ -123,7 +123,7 @@ jobs:
             version: 27ca7a05e65d24c784ba831225d0a53341719590
     env:
       container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
-      parallel-build-jobs: 2
+      parallel-build-jobs: 3
```
bader, line 126:
```suggestion
      parallel-build-jobs: 2
```

keryell:
Thanks!

PR #1068, opened 2025-03-10
[CI] Switch clang-format to cpp-linter action

bader:
Today we use our own implementation of clang-format check which posts
the diff proposed by clang-format to GitHub actions summary. Switching
to cpp-linter action enables use of review suggestions, which are more
user friendly.


bader:
Example of the updated check comments: https://github.com/bader/SYCL-CTS/pull/3.

@psalz, FYI.

keryell (approved):
Interesting!

PR #1066, opened 2025-03-10
[CI] Update Ubuntu image to latest version

bader:
GitHub deprecated Ubuntu 20.04 images on February 1. The support will
end completely on April 1.

Update image to `ubuntu-latest` to simplify future maintainanace of
docker images. All jobs we update are independent of a specific Ubuntu
version and are expected to work out of the box on any future Ubuntu
version.

PR #1069, opened 2025-03-12
[CI] Remove workaround for ccache/DPCPP

KornevNikita:
DPC++ compiler fixed the problem blocking ccache functionality reported
in https://github.com/intel/llvm/issues/5260.
Also there are 4 cores available now, see:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -123,7 +123,7 @@ jobs:
             version: 27ca7a05e65d24c784ba831225d0a53341719590
     env:
       container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
-      parallel-build-jobs: 4
+      parallel-build-jobs: 3
     container:
       image: khronosgroup/sycl-cts-ci:${{ matrix.sycl-impl }}-${{ matrix.version }}
     steps:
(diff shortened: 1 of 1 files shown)


bader:
> But GH also says ubuntu20.04 is unstable now, can we try 22.04?

I'm going to update Ubuntu images to the latest version. First patch is here: https://github.com/KhronosGroup/SYCL-CTS/pull/1066.

KornevNikita:
@bader I don't know if it's just luck or 3 cores are really faster, but the pre-commit passed.

bader:
@KornevNikita, you are very lucky. Switching to 3 threads breaks the build with AdaptiveCPP. Ran it twice in https://github.com/KhronosGroup/SYCL-CTS/pull/1067.

keryell (approved):
Cool to have more CPUs!

KornevNikita:
> @KornevNikita, you are very lucky. Switching to 3 threads breaks the build with AdaptiveCPP. Ran it twice in #1067.

Here I also used `ubuntu-latest`, but I don't claim it helps. Anyway, closing this one as it duplicates https://github.com/KhronosGroup/SYCL-CTS/pull/1067 & https://github.com/KhronosGroup/SYCL-CTS/pull/1066

PR #1075, opened 2025-04-07
OpenCL Interop: Use matching context when constructing queue

RossBrunton:
`make_queue` accepts both a native handle and a context. While the spec
doesn't explictly say so, it's my understanding that the "internal"
native handle of the provided context must match the context of the
provided handle.

This fixes the `opencl_interop_constructors` test, which was incorrectly
trying to construct a sycl::queue from a native handle created from a
different context.

### DIFF
--- a/tests/opencl_interop/opencl_interop_constructors.cpp
+++ b/tests/opencl_interop/opencl_interop_constructors.cpp
@@ -127,8 +127,10 @@ class TEST_NAME :
       /** check make_queue (cl_command_queue, const context&)
        */
       {
+        sycl::context context =
+            sycl::make_context<sycl::backend::opencl>(m_cl_context);
         sycl::queue queue = sycl::make_queue<sycl::backend::opencl>(
-            m_cl_command_queue, ctsContext);
+            m_cl_command_queue, context);
 
         cl_command_queue interopQueue =
             sycl::get_native<sycl::backend::opencl>(queue);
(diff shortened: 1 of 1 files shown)


bader (approved):

keryell (commented):
Would it be the same on line 95, 116, 199, 226, 345 like
```c++
            sycl::make_kernel<sycl::backend::opencl>(clKernel, ctsContext);
```
or is it compatible?

RossBrunton:
> Would it be the same on line 95, 116, 199, 226, 345 like
> 
> ```c++
>             sycl::make_kernel<sycl::backend::opencl>(clKernel, ctsContext);
> ```
> 
> or is it compatible?

What file are you looking at with those line numbers? `opencl_interop/opencl_interop_constructors.cpp` isn't calling `make_kernel` on those lines.

keryell:
> What file are you looking at with those line numbers?

Curious. I cannot find again what I was looking at. :thinking:

RossBrunton:
@keryell So are there any issues with this change? If we discover issues in another file then we can always fix them as another PR.

keryell (approved):
Thanks for the bug fix!

RossBrunton:
@bader I don't have merge permissions, could you merge this please?

keryell:
@RossBrunton I could merge it, but, if there is no emergency, I prefer @bader to do it as the code owner.

PR #1077, opened 2025-04-17
[NFC] Split `common.h` header [1/N]

AlexeySachkov:
This PR outlines functions like `linearize` and `unlinearize` to their own separate header, because they are only used by a few tests.

The change is expected to have positive (albeit probably unnoticeable) impact on overall CTS compile-times.

The main motivation/background is to refactor CTS in preparation for using modularized `sycl.hpp` (see KhronosGroup/SYCL-Docs#780). As of today all tests use `common.h` and therefore CTS won't benefit much from fine grained headers because everything anyways is included through a single entry point in form of `common.h`.

### DIFF
--- a/tests/common/common.h
+++ b/tests/common/common.h
@@ -833,45 +833,6 @@ inline bool kernel_supports_wg_size(sycl_cts::util::logger& log,
     }                                                   \
   } while (0);
 
-/// Linearizes a multi-dimensional index according to the specification.
-template <unsigned int dimension>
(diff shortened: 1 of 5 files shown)


AlexeySachkov:
> @AlexeySachkov So, do you plan to move this into `tests/common/range_index_space_id.h` instead?

Apology, I missed your comment. Yeah, if you agree, then let me move into an existing header instead

keryell (approved):
Looks good!
Thank you.

PR #1080, opened 2025-04-17
[NFC] Drop dead code from `common.h`

AlexeySachkov:


keryell (approved):
It looks like some pre-C++17 code.
Thanks!

PR #1081, opened 2025-04-17
Use built-in Catch2 functionality for enums printing

AlexeySachkov:
Documentation: [link](https://github.com/catchorg/Catch2/blob/5b3b22860344193bfd51e4c3c7429e94d8a5996f/docs/tostring.md#enums)

Note that this is not exactly an `NFC` change, because I've enabled more testing on AdaptiveCpp - it looks like some old assumptions of availability of some enumerations do not hold anymore. So, to simplify the codebase I dropped a few `#if`s


keryell (approved):
Interesting Catch2 feature.
Thanks!

PR #1082, opened 2025-04-23
fix: disable `Catch2` signal handlers

marekkozl:
* Catch2 framework is not respecting page fault handler chain affecting other layers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,12 @@ macro(add_submodule_directory RELPATH)
     add_subdirectory("${RELPATH}")
 endmacro()
 
+# Disable page fault handler feature, due to its implementation not respecting page fault handler chain
+if(WIN32)
+    set(CATCH_CONFIG_NO_WINDOWS_SEH ON CACHE BOOL "" FORCE)
+else()
+    set(CATCH_CONFIG_NO_POSIX_SIGNALS ON CACHE BOOL "" FORCE)
+endif()
 add_submodule_directory(vendor/Catch2)
 
 # set host compiler flags


steffenlarsen (approved):
LGTM!

CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ macro(add_submodule_directory RELPATH)
     add_subdirectory("${RELPATH}")
 endmacro()
 
+if(WIN32)
```
keryell, line 24: Could you add a comment here on what the problem is for you or an implementation?
Is there an issue with https://github.com/catchorg/Catch2/issues/2791 ?

keryell: I realize we use Catch2 from a submodule which is 2+ year old.
Would it make sense to first update Catch2 to latest release before changing the config?

marekkozl: The detailed comment is embedded inside commit description that will stay forever in the git blame. Please confirm if you want an explicit comment.

The issue that you have linked is describing Catch V3 compatibility, SYCL_CTS is using V2 which this commit is hot-fixing.

After this gets merged next step would be to create an issue in Catch explicitly about root of the issue -> https://github.com/catchorg/Catch2/blob/4c8671cfbbf0019d3827305d4d1c82a74eeb029a/src/catch2/internal/catch_run_context.cpp#L571-L574
which exists in the latest version

gmlueck: I'll add my $0.02.  We should not require future maintainers to use `git blame` to understand how the test suite works.  Yes, we should add a comment here.

keryell: Yes, please, add useful information to the code which should be readable by itself.

bader:
@keryell, please, review the comment added by https://github.com/KhronosGroup/SYCL-CTS/pull/1082/commits/6e25493203b1737c08aa570f1b95b5f11645281a.

keryell (approved):
Thanks!

PR #1083, opened 2025-04-23
[NFC] Use `device::has` query directly

AlexeySachkov:
This patch is done with the intent to eventually drop `util/extension.h` helper file. It probably had a reason to be there in SYCL 1.2.1 to avoid hardcoding OpenCL extension names everywhere. However, it is very easy to check for support for an optional feature in SYCL 2020 so there is less value in having this verbose helper.

Note that this commit does not remove all uses of `util/extension.h` - some of those uses are more complex than others and therefore their removal is left to another PR.

### DIFF
--- a/tests/address_space/address_space_fp64.cpp
+++ b/tests/address_space/address_space_fp64.cpp
@@ -9,7 +9,6 @@
 #define TEST_NAME address_space_fp64
 
 #include "../common/common.h"
-#include "./../../util/extensions.h"
 #include "address_space_common.h"
 
 namespace TEST_NAMESPACE {
@@ -24,9 +23,12 @@ class TEST_NAME : public sycl_cts::util::test_base {
   void run(util::logger &log) override {
     auto queue = util::get_cts_object::queue();
 
-    using availability =
-        util::extensions::availability<util::extensions::tag::fp64>;
-    if (!availability::check(queue, log)) return;
+    if (!queue.get_device().has(sycl::aspect::fp16)) {
+      WARN(
+          "Device doesn't support double precision floating point data type - "
+          "skipping the test");
+      return;
+    }
 
     test_types<double>(log);
   }
--- a/tests/multi_ptr/multi_ptr_accessor_constructor_fp16.cpp
+++ b/tests/multi_ptr/multi_ptr_accessor_constructor_fp16.cpp
@@ -6,7 +6,6 @@
 //
 *******************************************************************************/
 
-#include "../../util/extensions.h"
 #include "../common/common.h"
 #include "../common/type_coverage.h"
 #include "multi_ptr_accessor_constructor.h"
@@ -33,10 +32,10 @@ class TEST_NAME : public sycl_cts::util::test_base {
 
     auto queue = util::get_cts_object::queue();
 
-    using avaliability =
-        util::extensions::availability<util::extensions::tag::fp16>;
-    if (!avaliability::check(queue, log)) {
-      SKIP("Device does not support half precision floating point operations");
+    if (!queue.get_device().has(sycl::aspect::fp16)) {
+      SKIP(
+          "Device does not support half precision floating point operations - "
+          "skipping the test");
     }
 
     check_multi_ptr_accessor_constructor_for_type<sycl::half>{}(log,
--- a/tests/queue/queue_shortcuts_explicit_fp16.cpp
+++ b/tests/queue/queue_shortcuts_explicit_fp16.cpp
@@ -18,7 +18,6 @@
 //
 *******************************************************************************/
 
-#include "../../util/extensions.h"
 #include "../common/common.h"
 #include "queue_shortcuts_explicit.h"
 
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_explict;
 
 TEST_CASE("queue shortcuts explicit copy fp16", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp16>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp16)) {
+    SKIP(
         "Device does not support half precision floating point operations"
         "Skipping the test case.");
-    return;
   }
 
   check_queue_shortcuts_explicit_for_type<sycl::half>{}(queue, "sycl::half");
--- a/tests/queue/queue_shortcuts_explicit_fp64.cpp
+++ b/tests/queue/queue_shortcuts_explicit_fp64.cpp
@@ -18,7 +18,6 @@
 //
 *******************************************************************************/
 
-#include "../../util/extensions.h"
 #include "../common/common.h"
 #include "queue_shortcuts_explicit.h"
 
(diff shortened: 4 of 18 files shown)


tests/queue/queue_shortcuts_usm_fp64.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_usm;
 
 TEST_CASE("queue shortcuts unified shared memory fp64", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp64>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp64)) {
+    SKIP(
```
TApplencourt, line 32: Warm -> STOP

Just confirming that  the chang want intended

AlexeySachkov: > Just confirming that the change was intended

Yes, it was intended. Note `return;` statement which was removed from the `if`. I expect that the test will now be reported as actually skipped on an unsupported HW instead of reporting itself as passing one.

TApplencourt (approved):

keryell (approved):
Thanks for the cleaning.

tests/address_space/address_space_fp64.cpp
```diff
@@ -24,9 +23,12 @@ class TEST_NAME : public sycl_cts::util::test_base {
   void run(util::logger &log) override {
     auto queue = util::get_cts_object::queue();
 
-    using availability =
-        util::extensions::availability<util::extensions::tag::fp64>;
-    if (!availability::check(queue, log)) return;
+    if (!queue.get_device().has(sycl::aspect::fp16)) {
```
PeterTh, line 26: This seems wrong? Should be `fp64`, no?

tests/multi_ptr/multi_ptr_accessor_constructor_fp16.cpp
```diff
@@ -33,10 +32,10 @@ class TEST_NAME : public sycl_cts::util::test_base {
...
     auto queue = util::get_cts_object::queue();
 
-    using avaliability =
-        util::extensions::availability<util::extensions::tag::fp16>;
-    if (!avaliability::check(queue, log)) {
-      SKIP("Device does not support half precision floating point operations");
+    if (!queue.get_device().has(sycl::aspect::fp16)) {
+      SKIP(
+          "Device does not support half precision floating point operations - "
+          "skipping the test");
```
bader, line 38:
```suggestion
          "Device does not support half precision floating point operations");
```
I don't think we need explicitly add "- skipping the test" for SKIP macro.

We should avoid using WARN + return idiom in other tests and use SKIP instead.

tests/queue/queue_shortcuts_explicit_fp16.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_explict;
...
 TEST_CASE("queue shortcuts explicit copy fp16", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp16>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp16)) {
+    SKIP(
         "Device does not support half precision floating point operations"
         "Skipping the test case.");
```
bader, line 34:
```suggestion
        "Device does not support half precision floating point operations.");
```

tests/queue/queue_shortcuts_explicit_fp64.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_explict;
...
 TEST_CASE("queue shortcuts explicit copy fp64", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp64>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp64)) {
+    SKIP(
         "Device does not support double precision floating point operations"
         "Skipping the test case.");
```
bader, line 34:
```suggestion
        "Device does not support double precision floating point operations.");
```

tests/queue/queue_shortcuts_usm_fp64.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_usm;
...
         "Device does not support double precision floating point operations"
         "Skipping the test case.");
```
bader, line 34:
```suggestion
        "Device does not support double precision floating point operations.");
```

tests/queue/queue_shortcuts_usm_fp16.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_usm;
...
 TEST_CASE("queue shortcuts unified shared memory fp16", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp16>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp16)) {
+    SKIP(
         "Device does not support half precision floating point operations"
         "Skipping the test case.");
```
bader, line 34:
```suggestion
        "Device does not support half precision floating point operations.");
```

tests/queue/queue_shortcuts_kernel_fp64.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_kernel;
...
 TEST_CASE("queue shortcuts kernel function fp64", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp64>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp64)) {
+    SKIP(
         "Device does not support double precision floating point operations"
         "Skipping the test case.");
```
bader, line 34:
```suggestion
        "Device does not support double precision floating point operations.");
```

tests/queue/queue_shortcuts_kernel_fp16.cpp
```diff
@@ -29,13 +28,10 @@ using namespace queue_shortcuts_kernel;
...
 TEST_CASE("queue shortcuts kernel function fp16", "[queue]") {
   auto queue = util::get_cts_object::queue();
-  using availability =
-      util::extensions::availability<util::extensions::tag::fp16>;
-  if (!availability::check(queue)) {
-    WARN(
+  if (!queue.get_device().has(sycl::aspect::fp16)) {
+    SKIP(
         "Device does not support half precision floating point operations"
         "Skipping the test case.");
```
bader, line 34:
```suggestion
        "Device does not support half precision floating point operations.");
```

tests/address_space/address_space_fp64.cpp
keryell: Oops! Good point!

tomdeakin:
WG approved to merge

tests/multi_ptr/multi_ptr_accessor_constructor_fp16.cpp
AlexeySachkov: To clarify my motivation here - some tests are still using legacy test framework and I wasn't sure how `SKIP` (that comes from `Catch2`) interacts with it. Therefore, to be conservative I've avoided using `SKIP` in those tests.

AlexeySachkov:
> WG approved to merge

Thanks for applying comments :)

PR #1096, opened 2025-05-14
Remove all discard_queue_events tests for deprecated oneapi extension

JackAKirk:
This oneapi extension is now deprecated (https://github.com/intel/llvm/pull/18059) and not intended for sycl-cts tests. Some of these tests now fail since the implementation is a no-op, so that tests that check for expected warnings fail. https://github.com/intel/llvm/pull/18059 has kept core unit testing of basic SYCL functionality when using the no-op discard_events property (https://github.com/intel/llvm/tree/sycl/sycl/test-e2e/DeprecatedFeatures/DiscardEvents), which checks that the property doesn't alter expected behaviour. Since DPC++ is the only compiler using this extension, this should be sufficient and this duplicate CTS testing should be removed.


JackAKirk:
> @JackAKirk, please, sign CLA.

Done, thanks.

JackAKirk:
@bader Can this be merged? These failures are bothering CI people unnecessarily.

keryell (approved):
Thanks!

PR #1097, opened 2025-05-15
Fix flaky issue in test `oneapi_non_uniform_groups-group_barrier`

HPS-1:
Previously, test `oneapi_non_uniform_groups-group_barrier` has been failing flakily on PVC machines. It turns out to be a test issue caused by race conditions between sub groups inside the test case. Therefore, revising the test case to get rid of this issue.

### DIFF
--- a/tests/extension/oneapi_non_uniform_groups/group_barrier.h
+++ b/tests/extension/oneapi_non_uniform_groups/group_barrier.h
@@ -135,6 +135,7 @@ struct non_uniform_group_barrier_test {
 
               size_t llid = non_uniform_group.get_local_linear_id();
               size_t max_id = non_uniform_group.get_local_linear_range() - 1;
+              size_t offset = sub_group.get_group_id() * (sub_group.get_max_local_range().size());
 
               static_assert(
                   std::is_same_v<void, decltype(sycl::group_barrier(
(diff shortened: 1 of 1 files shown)


tests/extension/oneapi_non_uniform_groups/group_barrier.h
```diff
@@ -148,19 +148,21 @@ struct non_uniform_group_barrier_test {
                   "memory_scope fence_scope) is wrong\n");
 
               // test of default barrier
-              local_acc[llid] = llid;
-              sycl::group_barrier(non_uniform_group);
+              if (sub_group.get_group_linear_id() == 0){
```
AlexeySachkov, line 151: Tagging @steffenlarsen for feedback here. Would it be a better fix (in terms of coverage) to make the logic around `local_acc` more sophisticated to account for multiple sub-groups doing those calculations? I.e. we offset all indices that we use by `sub_group.get_group_id() * sub_group.get_max_local_range().size()` to make sure that every sub-group operates on a different section of the local accessor (adjusting the local accessor size accordingly to accommodate more data)

steffenlarsen: Agreed! Much better. ⭐

tests/extension/oneapi_non_uniform_groups/group_barrier.h
```diff
@@ -135,6 +135,7 @@ struct non_uniform_group_barrier_test {
 
               size_t llid = non_uniform_group.get_local_linear_id();
               size_t max_id = non_uniform_group.get_local_linear_range() - 1;
+              size_t offset = sub_group.get_group_id() * (sub_group.get_max_local_range().size());
```
steffenlarsen, line 138: Nit;
```suggestion
              size_t offset = sub_group.get_group_linear_id() * sub_group.get_max_local_range();
```

steffenlarsen (approved):

PR #1104, opened 2025-05-26
[NFC] Drop `device_eval.h`

AlexeySachkov:
Content of the header is only used within a single test, so there is no need for it to be located in common directory.

It seems useful at first glance, but tests for `range` are written using different helpers which seem more generic and functional than `DEVICE_EVAL`, so I think its fine to have the latter to be inlined into the only test that uses it.


keryell (commented):
On the other hands it does not cost anything (but the time to evaluate the `#include`) to keep it as is if there is another need in the future.

PR #1105, opened 2025-05-26
[NFC] Drop unnecessary `sycl.hpp` includes

AlexeySachkov:
Those files only define macro and those macro don't even use anything SYCL-specific.


tests/common/macros.h
```diff
@@ -22,8 +22,6 @@
 #ifndef __SYCLCTS_TESTS_COMMON_MACROS_H
 #define __SYCLCTS_TESTS_COMMON_MACROS_H
 
-#include <sycl/sycl.hpp>
```
keryell, line 25: If we go pedantic, to be replaced by
```c++
#include <string>
```
which is used in this file.

PR #1106, opened 2025-05-27
[NFC] Split `common.h` header [2/N]

AlexeySachkov:
This PR moves type checking helpers that are only used in a single test into that test.


keryell (approved):
Thanks!

PR #1114, opened 2025-06-09
Replace reference math functon with sycl version in common_vec.h

wenju-he:
Reference math function uses std math function, which may not work for a backend target if the function fails to be lowered to llvm intrinsic that the target supports.
Using sycl version in device code should work for all targets.

### DIFF
--- a/tests/common/common_vec.h
+++ b/tests/common/common_vec.h
@@ -26,7 +26,6 @@
 #include <sycl/sycl.hpp>
 
 #include "../../util/accuracy.h"
-#include "../../util/math_reference.h"
 #include "../../util/proxy.h"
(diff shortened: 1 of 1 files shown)


bader (approved):
Good catch.
Usage of `std::*` functions in the device code is not guaranteed by the spec.

ProGTX (approved):

bader:
We need approval from non-[Intel/Codeplay] for the merge. Maybe @tomdeakin, @TApplencourt or @keryell can take a look.

TApplencourt:
Indeed, seem to be a trivial fix (to a non trivial problem :) )

wenju-he:
@bader please help to merge, thanks.

PR #1115, opened 2025-06-09
Allow the selection of non-default build directories

frasercrmck:
The default is still 'build' but this change allows users to specify a different directory, with the --build-dir option. The build system is now invoked via cmake. As such, the -c/--build-system-call flag has been removed. To pass additional arguments through cmake to the underlying build system (e.g., Ninja) a new option has been added: --build-system-args.

Additionally, the script no longer changes directory internally; paths are relative to the current working directory.

### DIFF
--- a/run_conformance_tests.py
+++ b/run_conformance_tests.py
@@ -38,11 +38,16 @@ def handle_args(argv):
         type=str,
         required=True)
     parser.add_argument(
-        '-c',
-        '--build-system-call',
-        help='The call to the used build system.',
+        '--build-system-args',
+        help='Additional args to pass to the build system through CMake',
         type=str,
         required=True)
+    parser.add_argument(
+        '--build-dir',
+        help='The name of the build directory to use/create',
+        type=str,
+        default='build',
+        required=False)
     parser.add_argument(
         '--build-only',
         help='Whether to perform only a build without any testing.',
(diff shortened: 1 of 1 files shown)


run_conformance_tests.py
```diff
@@ -310,9 +316,9 @@ def main(argv=sys.argv[1:]):
     ctest_call = generate_ctest_call(additional_ctest_args)
 
     # Make a build directory if required and enter it
-    if not os.path.isdir('build'):
-        os.mkdir('build')
-    os.chdir('build')
+    if not os.path.isdir(build_dir):
```
frasercrmck, line 319: On a related note, I'm not a huge fan of using `chdir`. I was wondering if we could instead provide cmake and ctest the path to the build directory? They both have flags to do that.

frasercrmck: Ah so unfortunately it's a little more complicated than I had hoped. We take the `--build-system-call` and execute that unconditionally. I don't think we particularly want to have users pass `-c 'ninja -C <build-dir>' *and* `--build-dir <build-dir>`, do we?

We could still have the `chdir` but avoid the `mkdir`, of course. But at that stage, is it worth it?

keryell:
I think if you use the high-level CLI API of `cmake` it removes a lot of issues.
```
cmake --build $BUILD_DIR --target my-target --parallel --verbose
```
There is no reference to the back-end (`make`, `ninja`...) command details.

frasercrmck:
> I think if you use the high-level CLI API of `cmake` it removes a lot of issues.
> 
> ```
> cmake --build $BUILD_DIR --target my-target --parallel --verbose
> ```
> 
> There is no reference to the back-end (`make`, `ninja`...) command details.

Sorry for the delay. Yep good idea, I've done that now.

The script invokes `cmake --build $BUILD_DIR --parallel` and the user can pass additional arguments which are forwarded on to the build tool, using `--build-system-args`. That way they can pass `--build-system-args '-j1 -v' if they wanted to build single-threaded and verbose, for example. I think having `--parallel` is a sensible default. There's not a way of passing additional args to `cmake --build ...` itself, but I don't expect that to be an immediate concern.

The `BuildSystemCall` XML entry is repurposed as taking the full `cmake --build ... -- <additional-args>` invocation. I trust that's okay?

PR #1129, opened 2025-07-14
Fix `bundle_factory` to avoid compiling unnecessary kernels

againull:
Currently `bundle_factory` in `spec_constants_via_kernel_bundle` tests creates `kernel_bundle` without specifying kernel_id for some scenarios when it doesn't matter which bundle to use. That leads to the situation that all device images from executable are used which might be a lot if all tests are compiled into a single executable.

I.e., before change
`./bin/test_spec_constants  "*via_kernel_bundle_core"`
has been causing compilation of all kernels from the executable (including the ones completely  unrelated to kernel_bundle/spec_constants) which is unnecessary.

So, this change fixes bundle_factory to use specific kernel id.

### DIFF
--- a/tests/spec_constants/spec_constants_via_kernel_bundle.h
+++ b/tests/spec_constants/spec_constants_via_kernel_bundle.h
@@ -14,7 +14,9 @@
 #include "../common/type_coverage.h"
 #include "spec_constants_common.h"
 
+#include <algorithm>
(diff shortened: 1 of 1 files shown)


TApplencourt (approved):

keryell (approved):
Thank you.

PR #1131, opened 2025-07-31
Update queue creation to avoid using default device selection in `oneapi_enqueue_functions` tests

dyniols:
This PR updates queue creation to avoid using default device selection because if they are created by default device selector. It causes CTS run on unexpected devices sometimes.

For example:
Test is run on machine with CPU and GPU, and test is specified for to run on specific device and by using default selector test might not run on desired device.


TApplencourt:
LGTM. Not sure if t's the most idiomatic way of doing it (I'm sure ronan will said that we should never use `=` on new c++, only `{}`), but should solve the problem :)

tomdeakin:
This is a PR regarding a vendor extension, and so I believe this just requires a review from Intel (the originator).

keryell (approved):
Thanks.

PR #1142, opened 2025-09-16
Pass real filename to catch in legacy tests

Maetveis:
Catch2 has the ability to allow execution of tests from a specific file using the `-# [#<filename>]` command line arguments. This already works well for the tests directly using Catch2 macros, but not for the legacy tests which are registered through the `test_proxy` class.

Instead of passing `__FILE__` which always resolves to the proxy header `util/proxy.h`, take the filename we already stash into the `test_base::info` struct and pass that to Catch2. This allows to run legacy tests per-file using the `-# [#<filename>]` syntax as well.

To ensure the lifetime of the `test_base::info` struct is long enough, store it as a member variable of the `test_proxy` class, the instances of which are global and thus live until program termination.

### DIFF
--- a/util/proxy.h
+++ b/util/proxy.h
@@ -40,16 +40,16 @@ template <typename T>
 class test_proxy {
  public:
   test_proxy() {
-    test_base::info info;
-    T{}.get_info_legacy(info);
+    T{}.get_info_legacy(m_info);
 
     Catch::AutoReg(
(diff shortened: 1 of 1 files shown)


bader:
@Maetveis, please, sign CLA.

keryell (approved):
Thanks!

PR #1150, opened 2025-10-23
Fix free_function_commands test

KornevNikita:
Wrong order of arguments. Also updating CHECKs to properly verify results.
spec: https://github.com/KhronosGroup/SYCL-Docs/pull/922


tomdeakin:
WG discussed this, and we should check `dst[i] == i` instead to have caught this bug.

tomdeakin:
WG approved to merge.

PR #1155, opened 2025-11-10
Avoid bleeding async exceptions between cases in event tests

steffenlarsen:
This commit adds a event::wait_and_throw to the case checking that event::wait() does not consume asynchronous exceptions. This is to avoid the unconsumed exception from bleeding into other test cases in implementations that consume such exceptions when consuming other unconsumed exceptions.


tomdeakin (approved):
LGTM

PR #1159, opened 2025-12-11
Speed up oneapi_memcpy2d tests by avoiding redundant setup

Maetveis:
Due to the way the Catch2 framework interacts with `for_all_combinations` in the SYCL-CTS, we were inadvertently re-running setup code (allocation, fill, deallocation)
for each <T, src_ptr_type, dest_ptr_type> combination in the oneapi_memcpy2d tests.

Add a warning to the documentation of `for_all_combinations` to highlight this perfomance pitfall, and wrap the test code in an outer SECTION to avoid re-executing expensive setup operations on every test case re-execution.

On my local machine with an Intel GPU, the runtime of the test changes from 14s to 2.7s.
When running on an internal simulator it takes the test from more than 20 minutes (I killed it after that) to 9 minutes.

Disclaimer: I used a generative AI model to adjust the phrasing of the comments, but I reviewed every line of output.

### DIFF
--- a/tests/common/type_coverage.h
+++ b/tests/common/type_coverage.h
@@ -285,10 +285,10 @@ using is_not_a_type_pack =
  * @param head The first non-pack argument to pass into the functor
  * @param args The rest of the arguments to pass into the functor
  *
- * @warning When using this function with an `Action` that contains Catch2 SECTIONs,
- * be aware that any expensive operations (such as memory allocations, data
- * initialization, or device setup) performed before or between SECTIONs will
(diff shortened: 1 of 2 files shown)


bader (approved):
I am affiliated with Intel as you, so my vote doesn't count.
That's the reason I don't approve Intel contributions to SYCL-CTS, except tests for Intel extensions and non-CTS related areas like CI.
At the same time, I review all Intel PRs and comment only when I have something to say.
This PR looks good to me. 👍

keryell (approved):
Thanks!

PR #1167, opened 2026-01-06
Remove `[[noreturn]]` attribute from functions that return

0x12CC:
This commit removes the `[[noreturn]]` attribute from the `SYCL_EXTERNAL` function declarations to avoid undefined behavior since the function definitions return.


keryell (approved):
Thanks!

PR #1170, opened 2026-01-06
Fix check for exceptions with a `what_arg`

0x12CC:
The SYCL specification requires the string returned by `what()` contain the `what_arg` as a substring. This commit updates the test to check the same condition.


keryell (approved):
Thanks!

PR #1172, opened 2026-01-15
Placeholder should be used for link executable

frengels:
When setting the dpcpp executable directly it bypasses any additional flags which cmake would set. For example, this currently prevents cmake from setting `--target=` when cross-compiling the SYCL-CTS. Since `CMAKE_CXX_COMPILER` is already set to `DPCPP_CXX_EXECUTABLE` this will have the same desired effect.

### DIFF
--- a/cmake/FindDPCPP.cmake
+++ b/cmake/FindDPCPP.cmake
@@ -89,7 +89,9 @@ set_target_properties(DPCPP::Runtime PROPERTIES
 
 set(CMAKE_CXX_COMPILER ${DPCPP_CXX_EXECUTABLE})
 # Use DPC++ compiler instead of default linker for building SYCL application
-set(CMAKE_CXX_LINK_EXECUTABLE "${DPCPP_CXX_EXECUTABLE} <FLAGS> <OBJECTS> -o <TARGET> \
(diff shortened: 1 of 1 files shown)


bader:
> When setting the dpcpp executable directly it bypasses any additional flags which cmake would set.

I think this was done intentionally. CMake configuration was inherited from the Codeplay ComputeCPP implementation, which has an option to use system C++ compiler as a host compiler for the SYCL code. DPC++ has this option as well, but it's not a default one. 

> For example, this currently prevents cmake from setting `--target=` when cross-compiling the SYCL-CTS

@frengels, did you considered using `DPCPP_FLAGS` for setting DPC++ compiler options?

> Since `CMAKE_CXX_COMPILER` is already set to `DPCPP_CXX_EXECUTABLE` this will have the same desired effect.

Considering that ComputeCPP is EOLed, we should consider dropping DPC++ specific variables (i.e. `DPCPP_CXX_EXECUTABLE` and `DPCPP_FLAGS`) and use SYCL compiler as standard C++ host compiler as well. This will require more clean-up. 

@frengels, what cmake options do use for cross-compiling SYCL-CTS?

PR #1174, opened 2026-01-26
Fix bug in event.cpp where queue_impl outlives async handler and causes a segmentation fault

lbushi25:
This PR fixes a flaky bug in `event.cpp` where an async exception handler associated to a queue had already gone out of scope by the time the `queue_impl` destructor was called. Since the latter(destructor) calls the former(exception handler), this is clearly a problem as we get undefined behavior. This PR moves the exception handler, previously declared in function scope, to global scope so that it always outlives the queue resources.

### DIFF
--- a/tests/event/event.cpp
+++ b/tests/event/event.cpp
@@ -225,6 +225,7 @@ static test_exception_handler teh1;
 static test_exception_handler teh2;
 
 TEST_CASE("event::wait does not report asynchronous errors", "[event]") {
+  teh1.clear();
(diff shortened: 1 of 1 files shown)


TApplencourt:
In theory better to `clear();` at the beginning of the test, not at the end. 

Indeed if a test fail, maybe `clear()` will not be triggered and leave the handle in a bad state. But 🤷🏽 , not that critical. 
Thanks for the fix!

PR #1182, opened 2026-02-26
Add cases for *_vector_width_long_long

KornevNikita:
https://github.com/KhronosGroup/SYCL-CTS/issues/1178

### DIFF
--- a/tests/device/device_info.cpp
+++ b/tests/device/device_info.cpp
@@ -110,6 +110,8 @@ TEST_CASE("device info", "[device]") {
                          uint32_t>(dev);
     check_get_info_param<sycl::info::device::preferred_vector_width_long,
                          uint32_t>(dev);
+    check_get_info_param<sycl::info::device::preferred_vector_width_long_long,
+                         uint32_t>(dev);
     check_get_info_param<sycl::info::device::preferred_vector_width_float,
                          uint32_t>(dev);
     check_get_info_param<sycl::info::device::preferred_vector_width_double,
@@ -123,6 +125,8 @@ TEST_CASE("device info", "[device]") {
     check_get_info_param<sycl::info::device::native_vector_width_int, uint32_t>(
         dev);
     check_get_info_param<sycl::info::device::native_vector_width_long,
+                         uint32_t>(dev);
+    check_get_info_param<sycl::info::device::native_vector_width_long_long,
                          uint32_t>(dev);
     check_get_info_param<sycl::info::device::native_vector_width_float,
                          uint32_t>(dev);
--- a/tests/device/device_info.cpp
+++ b/tests/device/device_info.cpp
@@ -110,8 +110,10 @@ TEST_CASE("device info", "[device]") {
                          uint32_t>(dev);
     check_get_info_param<sycl::info::device::preferred_vector_width_long,
                          uint32_t>(dev);
+#if !SYCL_CTS_COMPILING_WITH_SIMSYCL && !SYCL_CTS_COMPILING_WITH_DPCPP
     check_get_info_param<sycl::info::device::preferred_vector_width_long_long,
                          uint32_t>(dev);
+#endif
     check_get_info_param<sycl::info::device::preferred_vector_width_float,
                          uint32_t>(dev);
     check_get_info_param<sycl::info::device::preferred_vector_width_double,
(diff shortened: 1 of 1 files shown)


bader (changes requested):
@KornevNikita, please, fix CI checks.

KornevNikita:
No compiler supports this change yet, therefore adding under macros.

bader (approved):
👍

TApplencourt (approved):

TApplencourt:
No check for adativeCPP? But I guess the CI pass, so not required :)

KornevNikita:
> No check for adativeCPP? But I guess the CI pass, so not required :)

@TApplencourt ACPP doesn't support the whole block:
https://github.com/KhronosGroup/SYCL-CTS/blob/cea4ae6147ea179d95e9281173dd4e43f90fbac0/tests/device/device_info.cpp#L84-L85

TApplencourt:
Thanks!
- WG Approval
- Can Merge SYCL-DOC PR at the same time

tests/device/device_info.cpp
```diff
@@ -110,6 +110,10 @@ TEST_CASE("device info", "[device]") {
                          uint32_t>(dev);
     check_get_info_param<sycl::info::device::preferred_vector_width_long,
                          uint32_t>(dev);
+#if !SYCL_CTS_COMPILING_WITH_SIMSYCL && !SYCL_CTS_COMPILING_WITH_DPCPP
```
gmlueck, line 113: I don't think we should disable tests via `SYCL_CTS_COMPILING_WITH_DPCPP`.  DPC++ claims conformance to the CTS, so we can't do this honestly if we disable some of the tests.  I do not think any other tests are currently disabled for DPC++ with this macro.

bader: > DPC++ claims conformance to the CTS

My understanding DPC++ claims conformance to specific versions of the spec/CTS. It doesn't mean that any DPC++ version passes any CTS version.

The spec, implementation and CTS all are moving targets. Here we test new functionality, which DPC++ implementation doesn't support yet.

KornevNikita: If we don't want to use this macro, then DPCPP needs to support this, there is no another way, right?

Anyways, I can quickly implement some dummy solution so the test passes, WDYT?

gmlueck: I was assuming that test would just fail for DPC++ until we implement this feature, and that failure would ensure that we implement the feature before we submit our next conformance.

If we instead disable the test with the macro, what process do we have to ensure that the test is re-enabled by the time we submit our next conformance?  I'm concerned that if we add the macro now to disable the test, we will just forget about it, and then we'll end up claiming conformance in the future for a CTS with some tests disabled.

bader: Conformance logs which implementer sends the Khronos include SYCL-CTS commit hash and the diff applied locally on top of it (if any). Khronos working group is expected to raise concerns about disabled tests and not acknowledge tested implementation as conformant if any changes are considered to be unreasonable.

On DPC++ side, I expect @KornevNikita to handle it as usual (i.e. open a tracker, implement feature in DPC++, uplift DPC++ version in Khronos SYCL-CTS CI and remove guards).

gmlueck: I think it is unreasonable to expect the WG to read through the entire test suite each time there is a conformance submission to see if tests have been disabled.  Instead, the CTS should be structured to make this obvious from the test run submission.  Can we create a cmake config option like `-D DISABLE_FAILING_TESTS_DPCPP`, and structure the CTS so that tests are only disabled when that option is specified?  That way, it will be obvious from the submission whether tests have been disabled.  It will also be obvious to vendors even before they attempt to submit a conformance run.  My concern is not that a vendor will intentionally try to seek conformance for a failing implementation.  Instead, I'm concerned that we will simply forget to re-enable tests after they have been disabled.

bader: > I think it is unreasonable to expect the WG to read through the entire test suite each time there is a conformance submission to see if tests have been disabled.

I did this for the previous submission. It's quite easy because we use one standard macro to disable test parts per implementation. It takes less than a minute to check.

> Instead, the CTS should be structured to make this obvious from the test run submission.

CTS provides an option to [disable a test case](https://github.com/KhronosGroup/SYCL-CTS/tree/main/docs#compile-time-disabled-test-cases). The submission log will state that test case is disabled.
Using this option for this change will mean that all device information query tests will be disabled.

TApplencourt:
WG: Agree to merge.

gmlueck (approved):

PR #1187, opened 2026-03-24
test_oneapi_device_global_unique...: use the device filter

Maetveis:
device_global_functional_unique_for_every_device has its own logic to select (2) devices, but it does not apply the regex filtering used by other tests (`--device` command line option). This PR adds the filtering to the test, by calling `cts_selector` and checking for a negative return value before considering to use a device.


TApplencourt (approved):

illuhad (approved):

keryell (approved):

TApplencourt:
Enough approval. @bader you can merge : )Thanks!

PR #1199, opened 2026-05-18
Remove simsycl from CI configuration

TApplencourt:
Removed simsycl implementation from CI workflow, as it's not maintained.

We can discuss we want to remove simsycl from other place in the CTS, but this is the minimal change to avoid burning CI cycle.


bader (approved):

keryell (approved):
That is sad.

PR #1201, opened 2026-05-27
Add ProtoSYCL to the CTS

0x12CC:
This PR adds support for running CTS tests with ProtoSYCL.


docker/protosycl/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    rm llvm.sh
+
+# Set clang-22 as the default clang version.
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-22 100 && \
+    update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-22 100
+ENV CC=clang
+ENV CXX=clang++
+
+# Build ProtoSYCL.
+RUN git clone https://github.com/0x12CC/ProtoSYCL.git && \
```
keryell, line 32: Is the `FetchContent` in  [‎cmake/FindProtoSYCL.cmake](https://github.com/KhronosGroup/SYCL-CTS/pull/1201/changes#diff-9ea4510011c48d9f560c6521967bc57de9a2560e70bd816275c386c7029c4835) not enough?

0x12CC: The `FetchContent` will build the compiler and runtime library. The reason I've set it up this way is so that we can select the implementation version in `.github/workflows/cts_ci.yml` just like for other implementations. It also ensures the SYCL compiler is built before the first target in the CTS. I'm hoping there will be a better way once CMake supports SYCL.

TApplencourt:
- CCache: Cannot cache.
- We will ask for a new runner

-- Try to fix CCache this week (Thomas will try)

Next Week:
- No matter what we will merge.

TApplencourt:
Amazing. That worked. Thank claude ;'0

0x12CC:
Thank you, @TApplencourt!

CMakeLists.txt
```diff
@@ -3,6 +3,8 @@ project(sycl_cts LANGUAGES CXX)
 
 if(SYCL_IMPLEMENTATION STREQUAL SimSYCL)
     set(CMAKE_CXX_STANDARD 20)
+elseif(SYCL_IMPLEMENTATION STREQUAL ProtoSYCL)
```
0x12CC, line 6: @bader, you asked me to move this into `FindProtoSYCL.cmake` in a previous PR. I’m adding this comment here since I closed that PR.

I haven’t found a way to move this without also changing the branches for the other SYCL implementations. If you have any suggestion for how I can change this, please let me know.

TApplencourt: Will do this in a follow-up PR.

TApplencourt:
Thomas: Fix email address.

.github/workflows/cts_ci.yml
```diff
@@ -188,3 +203,47 @@ jobs:
...
+          chmod -R +x bin/
+          find bin/ -type f -executable | while read -r binary; do
+            echo "=== Running test category: $binary ==="
+            "$binary"
+          done
+
+  cts-passes-for-all-implementations:
+    needs: run-cts
+    if: always()
+    runs-on: ubuntu-22.04
```
keryell: Why so old?

docker/protosycl/configure.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -o errexit -o pipefail -o noclobber -o nounset
+cmake . -G Ninja -B build \
+    -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+    -DCMAKE_CXX_COMPILER=/ProtoSYCL/build/sycl++ \
+    -DCMAKE_CXX_STANDARD=23 \
```
keryell, line 6: Is this compatible with Ubuntu *22*?

CMakeLists.txt
bader: > I haven’t found a way to move this without also changing the branches for the other SYCL implementations. If you have any suggestion for how I can change this, please let me know.

I'm not sure I understand the problem. Could you please clarify why other SYCL implementations are impacted? I would expect `FindProtoSYCL.cmake` to be used only for ProtoSYCL implementation. Other implementations should not use `FindProtoSYCL.cmake` and therefore change C++ standard.

TApplencourt:
~Oh no... I think I screwed up the force push >< Sorry!! ( at least it's my correct email now...)~
All good! (I hate git)

Sorry greg, didn't wanted to add you as reviewer. Miss-clicked somehow

.github/workflows/cts_ci.yml
bader: It's a copy-paste of `cts-compiles-for-all-implementations` job. I think it should be fine to update both jobs to use `ubuntu-latest` runners.

docker/protosycl/configure.sh
bader: We compile and run tests on latest Ubuntu. 
The only job we run on Ubuntu 22 reports the execution status: pass or fail.

.github/workflows/cts_ci.yml
```diff
@@ -138,7 +142,7 @@ jobs:
             -DSYCL_CTS_EXCLUDE_TEST_CATEGORIES="${{ env.container-workspace }}/ci/${{ matrix.sycl-impl }}.filter" \
             -DSYCL_CTS_MEASURE_BUILD_TIMES=ON
       - name: Set up ccache
-        uses: actions/cache@v3
+        uses: actions/cache@v5
```
keryell:
```suggestion
        uses: actions/cache@v6
```

.github/workflows/cts_ci.yml
```diff
@@ -172,6 +182,11 @@ jobs:
         with:
           name: build-times-${{ matrix.sycl-impl }}
           path: ${{ env.container-workspace }}/build/build_times.log
+      - name: Upload test binaries artifact
+        uses: actions/upload-artifact@v4
```
keryell:
```suggestion
        uses: actions/upload-artifact@v7
```

.github/workflows/cts_ci.yml
```diff
@@ -61,6 +61,8 @@ jobs:
             version: fcd26a61b600cd4b653724e59e72d4f7d1da259c
           - sycl-impl: simsycl
             version: 960c155109b066b26c1ecb6b2348ff9f6b69704d
+          - sycl-impl: protosycl
+            version: f48ad43d9c1f85d6ac5cc3abb00ee764d7f1edbd
     steps:
       - name: Checkout
         uses: actions/checkout@v3
```
keryell:
```suggestion
        uses: actions/checkout@v7
```
While we are here 😄

.github/workflows/cts_ci.yml
```diff
@@ -188,3 +203,47 @@ jobs:
...
+        include:
+          - sycl-impl: protosycl
+            version: f48ad43d9c1f85d6ac5cc3abb00ee764d7f1edbd
+    env:
+      container-workspace: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}
+    container:
+      image: khronosgroup/sycl-cts-ci:${{ matrix.sycl-impl }}-${{ matrix.version }}
+    steps:
+      - name: Download test binaries artifact
+        uses: actions/download-artifact@v4
```
keryell, line 223:
```suggestion
        uses: actions/download-artifact@v7
```
Probably a good opportunity to have some Agentic AI to update all the official actions to their latest versions.

keryell (approved):

docker/protosycl/configure.sh
0x12CC: I've switched this to the latest Ubuntu.

.github/workflows/cts_ci.yml
TApplencourt: Todo in a follow up PR. Thanks!

PR #1202, opened 2026-05-27
Disambiguate `nextafter`

0x12CC:
Argument-dependent lookup makes the call ambiguous since the `sycl::half` argument is defined in the `sycl` namespace. This PR changes the reference definitions to call the other reference definitions.


0x12CC:
@bader, could you please review this change? It's blocking #1201.

bader (approved):

keryell (approved):

TApplencourt (approved):

PR #1207, opened 2026-07-01
Disambiguate math reference functions

0x12CC:
Follow-up to #1202. Argument-dependent lookup makes these calls ambiguous since the `sycl::half` type is defined in the `sycl` namespace. This change ensures the reference definitions call other reference definitions.


TApplencourt (approved):

TApplencourt:
Thanks!

PR #1208, opened 2026-07-02
Skip kernel bundle tests when `online_compiler` not available

0x12CC:
According to section 4.11.7 of the SYCL 2020 specification, certain calls to `get_kernel_bundle` throw `errc::invalid` when a device does not have `aspect::online_compiler` or `aspect::online_linker`. This change updates certain kernel bundle tests to skip if the device does not have the required aspects.


TApplencourt (approved):

TApplencourt:
Thank!

PR #1209, opened 2026-07-02
Prevent CI tests from being skipped

0x12CC:
GitHub Actions default to checking `success()` in `if` conditionals unless the expression includes another status check function. This change ensures that the `run-cts` job always runs, even when upstream jobs for building container images are skipped.

### DIFF
--- a/.github/workflows/cts_ci.yml
+++ b/.github/workflows/cts_ci.yml
@@ -206,7 +206,7 @@ jobs:
 
   run-cts:
     needs: cts-compiles-for-all-implementations
-    if: needs.cts-compiles-for-all-implementations.result == 'success'
+    if: always() && !cancelled() && needs.cts-compiles-for-all-implementations.result == 'success'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/cts_ci.yml
```diff
@@ -206,7 +206,7 @@ jobs:
 
   run-cts:
     needs: cts-compiles-for-all-implementations
-    if: needs.cts-compiles-for-all-implementations.result == 'success'
+    if: always() && !cancelled() && needs.cts-compiles-for-all-implementations.result == 'success'
```
TApplencourt, line 210: `// Usefull for fork`

0x12CC, line 210:
```suggestion
    # This expression ensures the job runs even when other jobs are skipped.
    if: always() && !cancelled() && needs.cts-compiles-for-all-implementations.result == 'success'
```

0x12CC:
Looks like it fixed the issue! @TApplencourt, I've added a descriptive comment in 46fdf641e8fd034e231caa23c2a4915ed3db6d44.

PR #1210, opened 2026-07-02
Move `CMAKE_CXX_STANDARD` to implementation CMake files

0x12CC:
This changes upates the CMake files to enable each SYCL implementation to choose a required C++ standard.

### DIFF
--- a/cmake/AddSYCLExecutable.cmake
+++ b/cmake/AddSYCLExecutable.cmake
@@ -20,7 +20,19 @@ find_file(SYCL_IMPLEMENTATION_ADAPTER
   Adapt${SYCL_IMPLEMENTATION}.cmake
   PATHS ${CMAKE_MODULE_PATH}
 )
+set(SYCL_IMPLEMENTATION_CXX_STANDARD "" CACHE INTERNAL "")
 include("${SYCL_IMPLEMENTATION_ADAPTER}")
+
+if (NOT SYCL_IMPLEMENTATION_CXX_STANDARD MATCHES "^[0-9]+$")
+    message(FATAL_ERROR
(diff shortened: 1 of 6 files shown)


cmake/AddSYCLExecutable.cmake
```diff
@@ -20,8 +20,20 @@ find_file(SYCL_IMPLEMENTATION_ADAPTER
   Adapt${SYCL_IMPLEMENTATION}.cmake
   PATHS ${CMAKE_MODULE_PATH}
 )
+set(SYCL_IMPLEMENTATION_CXX_STANDARD "" CACHE INTERNAL "")
 include("${SYCL_IMPLEMENTATION_ADAPTER}")
 
+if (NOT SYCL_IMPLEMENTATION_CXX_STANDARD MATCHES "^[0-9]+$")
```
bader: This is more complex than I expected it to be. Do we really need to add `SYCL_IMPLEMENTATION_CXX_STANDARD` variable? Can't Adapt*.cmake set `CMAKE_CXX_STANDARD` directly and let CMake to validate the value?

0x12CC: I don't think we need to add `SYCL_IMPLEMENTATION_CXX_STANDARD`. We can set `CMAKE_CXX_STANDARD` directly. I added the error to make sure implementations don't forget to set the variable. New SYCL implementations are added infrequently so I think either approach should work fine.

0x12CC: I've made the change to use `CMAKE_CXX_STANDARD` directly in de86a25fb492a0d812473f1c32b5b845445230bc.

bader (approved):

PR #1214, opened 2026-07-07
Fix invalid reqd_work_group_size in kernel_features_reqd_work_group_size.cpp

wenju-he:
Recently https://github.com/llvm/llvm-project/commit/6794e31e894d adds LLVM IR verification that product of reqd_work_group_size elements should not exceed uint64 max. The new rule make sense because SYCL 2020 spec (Secion 4.9.1.1, Table 79) defines range::size() as product of its dimensions and return type is size_t.

This PR fixes out-of-range reqd_work_group_size values in the test.

### DIFF
--- a/tests/optional_kernel_features/CMakeLists.txt
+++ b/tests/optional_kernel_features/CMakeLists.txt
@@ -1,5 +1,3 @@
 file(GLOB test_cases_list *.cpp)
 
-add_compile_options(-Wno-implicitly-unsigned-literal)
-
 add_cts_test(${test_cases_list})
--- a/tests/optional_kernel_features/CMakeLists.txt
+++ b/tests/optional_kernel_features/CMakeLists.txt
@@ -1,3 +1,5 @@
 file(GLOB test_cases_list *.cpp)
 
+add_compile_options(-Wno-implicitly-unsigned-literal)
+
 add_cts_test(${test_cases_list})
--- a/tests/optional_kernel_features/kernel_features_reqd_work_group_size.cpp
+++ b/tests/optional_kernel_features/kernel_features_reqd_work_group_size.cpp
@@ -31,32 +31,34 @@ namespace kernel_features_reqd_work_group_size {
 using namespace sycl_cts;
 using namespace kernel_features_common;
 
-template <size_t N, int Dimensions>
+template <size_t N0, size_t N1, size_t N2, int Dimensions>
 class Functor {
  public:
-  [[sycl::reqd_work_group_size(N)]] void operator()(sycl::nd_item<1>) const {}
-  [[sycl::reqd_work_group_size(N)]] void operator()(sycl::group<1>) const {}
+  [[sycl::reqd_work_group_size(N0)]] void operator()(sycl::nd_item<1>) const {}
+  [[sycl::reqd_work_group_size(N0)]] void operator()(sycl::group<1>) const {}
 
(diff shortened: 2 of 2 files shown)


tests/optional_kernel_features/kernel_features_reqd_work_group_size.cpp
```diff
@@ -124,8 +129,10 @@ void test_size() {
 DISABLED_FOR_TEMPLATE_TEST_CASE_SIG(AdaptiveCpp)
 ("Exceptions thrown by [[reqd_work_group_size(N)]] with unsupported size",
  "[kernel_features]", ((int Dimensions), Dimensions), 1, 2, 3)({
-  test_size<4, Dimensions>();
-  test_size<4294967295, Dimensions>();
+  test_size<4, 4, 4, Dimensions>();
+  // Product of reqd_work_group_size elements should fit in range::size() return
+  // type size_t.
```
TApplencourt: Do we execute this code? I think we do in `run_separate_lambda_nd_range`. 

If so can we  `4294967295 + 1` or something?

 I 'm affraid that `std::numeric_limits<std::size_t>::max()` take too long to execute for some backend.  For single thread CPU it will take arround 55 year...

TApplencourt: If I remember correctly think it's the strategy we adopted to test the khr max work group query (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:khr-max-work-group-queries)

tests/optional_kernel_features/kernel_features_reqd_work_group_size.cpp
```diff
@@ -26,96 +26,101 @@
...
 #include <cmath>
+#include <limits>
 
 namespace kernel_features_reqd_work_group_size {
 using namespace sycl_cts;
 using namespace kernel_features_common;
 
-template <size_t N, int Dimensions>
+template <size_t N0, size_t N1, size_t N2, int Dimensions>
 class Functor {
```
TApplencourt, line 36: Should we put an `assert (N0 * N1 * N2 <= uint64_max)` so futur people will not make the same mistake as prevoous people ? :)

tests/optional_kernel_features/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
 file(GLOB test_cases_list *.cpp)
 
+add_compile_options(-Wno-implicitly-unsigned-literal)
```
TApplencourt: Is this really required? Or should fix the our literal using `UL` when required?

tests/optional_kernel_features/kernel_features_reqd_work_group_size.cpp
wenju-he: done, limited to `4294967295 + 1` when size_t size is 64

tests/optional_kernel_features/kernel_features_reqd_work_group_size.cpp
wenju-he: done

tests/optional_kernel_features/CMakeLists.txt
wenju-he: remove, now use 4294967296ull

TApplencourt:
Thanks!

PR #1219, opened 2026-07-16
[SYCL][OpenCL] Use qualified OpenCL types and include OpenCL header

uditagarwal97:
These tests fail if implementation stops pulling in OpenCL headers transitively. 
This PR makes tests use OpenCL types in `sycl::opencl::` namespace, in accordance to SYCL Spec (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_data_types_2).

`tests/host_task/host_task_interop_api.cpp` uses OpenCL APIs but don't include OpenCL header and rely on SYCL headers to bring them in. This PR also fixes that.


tests/opencl_interop/opencl_interop_scalars.cpp
```diff
@@ -57,7 +57,7 @@ class TEST_NAME : public util::test_base {
       bool device_supports_fp64 = device.has(sycl::aspect::fp64);
 
       // Integral Interop Data Types
-      if (!check_type_min_size<cl_bool>(1)) {
+      if (!check_type_min_size<sycl::opencl::cl_bool>(1)) {
```
gmlueck, line 50: Note that these types in the `sycl::opencl` namespace are only guaranteed to exist for implementations that support the OpenCL backend (Appendix C).  Does this code need to be protected by `#ifdef SYCL_BACKEND_OPENCL`?

bader: I suggest moving the whole file to `opencl_interop` test.

uditagarwal97: Done in 625b80fb5d4558dfde5e021e17d9a5c13b61fd86

gmlueck (approved):

TApplencourt (approved):

uditagarwal97:
@TApplencourt please merge the PR. Thanks!

bader (approved):

TApplencourt:
Thanks a lot! Sorry for the delay

PR #1220, opened 2026-07-17
[SYCL] Remove usage of non-standard vector data types in SYCL 2020 accessor conformance tests

lbushi25:
Delete two header files that are not relevant for testing core functionality anymore. See https://github.com/KhronosGroup/SYCL-CTS/pull/1220#issuecomment-5121792637 further down more more details.


bader:
@lbushi25, the goal of the conformance test suite is to verify that implementation follows the SYCL 2020 specification. Please, re-write the PR title and description to demonstrate where the test fails to achieve this goal.
My understanding is SYCL 1.2.1 specification defines these types and the accessor test was copied from the CTS for SYCL 1.2.1. We can remove uses of `::sycl::cl_<vector type>` because SYCL 2020 specification doesn't require implementation to define them.

lbushi25:
> @lbushi25, the goal of the conformance test suite is to verify that implementation follows the SYCL 2020 specification. Please, re-write the PR title and description to demonstrate where the test fails to achieve this goal. My understanding is SYCL 1.2.1 specification defines these types and the accessor test was copied from the CTS for SYCL 1.2.1. We can remove uses of `::sycl::cl_<vector type>` because SYCL 2020 specification doesn't require implementation to define them.

I've rewritten it in a more direct way. Does the description justify the changes now?

bader:
> > @lbushi25, the goal of the conformance test suite is to verify that implementation follows the SYCL 2020 specification. Please, re-write the PR title and description to demonstrate where the test fails to achieve this goal. My understanding is SYCL 1.2.1 specification defines these types and the accessor test was copied from the CTS for SYCL 1.2.1. We can remove uses of `::sycl::cl_<vector type>` because SYCL 2020 specification doesn't require implementation to define them.
> 
> I've rewritten it in a more direct way. Does the description justify the changes now?

Yes. Thanks!

TApplencourt (approved):

TApplencourt:
Thanks

tests/accessor_legacy/accessor_types_image_core.h
```diff
@@ -42,7 +42,7 @@ class check_all_types_image_core {
     if (!availability::check(queue, log)) return;
 
     const auto types =
-        named_type_pack<sycl::cl_int4, sycl::cl_uint4, sycl::cl_float4,
+        named_type_pack<sycl::int4, sycl::uint4, sycl::float4,
                         user_alias>::generate("sycl::opencl::cl_int",
                                               "sycl::opencl::cl_uint",
                                               "sycl::opencl::cl_float",
```
TApplencourt: Should we change the name of the string too for coherency ?

lbushi25: I think these are fine, they are aliases which, unlike their vector counterparts, are actually defined by the SYCL standard.

bader: If I get it right, the names must correspond to the changed types. If so, type names must be updated as well.

lbushi25: I've replaced the types with a more verbose sycl::vec<...> which is equivalent and makes the relationships between the type and its underlying element type evident.

bader (changes requested):
@lbushi25, I just realized that these header files validate SYCL 1.2.1 image functionality.
@0x12CC removed SYCL 1.2.1 image tests in https://github.com/KhronosGroup/SYCL-CTS/pull/1165. My understanding is that `tests/accessor_legacy/accessor_types_image_core.h` and `tests/accessor_legacy/accessor_types_image_fp16.h` should have been removed in #1165. I see no uses of `check_all_types_image_core`/`check_all_types_image_fp16` class methods. @0x12CC, please, confirm.

I suggest we just remove these two headers.

lbushi25:
> @lbushi25, I just realized that these header files validate SYCL 1.2.1 image functionality. @0x12CC removed SYCL 1.2.1 image tests in #1165. My understanding is that `tests/accessor_legacy/accessor_types_image_core.h` and `tests/accessor_legacy/accessor_types_image_fp16.h` should have been removed in #1165. I see no uses of `check_all_types_image_core`/`check_all_types_image_fp16` class methods. @0x12CC, please, confirm.
> 
> I suggest we just remove these two headers.

Ok, I'm on standby for confirmation and if it is indeed the case, I will remove them.

0x12CC:
I don't think these class templates are related to SYCL 1.2.1. The core functionality is to check `queue.get_device().get_info<sycl::info::device::image_support>()` and conditionally skip tests based on the query result. The query itself is deprecated so it would make sense to replace it with `device::has(aspect::image)`.

My preference would be to delete these headers. Although one could use them for SYCL 2020 image tests, we don't.

lbushi25:
@bader Can you approve or request further changes please?

bader (approved):

TApplencourt:
We can merge. Thanks a lot!

PR #1228, opened 2026-08-19
Atomic test from inclusive scope

### LINKED ISSUE #1226: Atomic Inclusive Memory Scope
- Need to add tests for: 
  - https://github.com/KhronosGroup/SYCL-Docs/pull/1032

TApplencourt:
Fix https://github.com/KhronosGroup/SYCL-CTS/issues/1226


tests/atomic_ref_stress/atomic_ref_stress_common.h
```diff
@@ -41,27 +83,37 @@ class atomicity_device_scope {
...
+          });
         });
-      });
+      }
+      bool res;
+      if constexpr (std::is_floating_point_v<T>)
+        res = atomic_ref::tests::common::compare_floats<T>(val, size * 2);
+      else
+        res = (val == size * 2);
+      CHECK(res);
```
bader: I suggest simplifying to:
```suggestion
      if constexpr (std::is_floating_point_v<T>)
        CHECK(atomic_ref::tests::common::compare_floats<T>(val, size * 2));
      else
        CHECK(val == size * 2);
```

bader (approved):

tests/atomic_ref_stress/atomic_ref_stress_common.h
TApplencourt: Indeed! `tmp` variable are ugly (even more when they are bool)

PR #1231, opened 2026-08-27
Add tests for nd_range default constructor

KornevNikita:
KhronosGroup/SYCL-Docs#1043 adds a default constructor to nd_range, so that an nd_range can be declared before its global and local ranges are known, matching range and id which are already default constructible.

Extend the existing by-value semantics checks with:
 - a static_assert that nd_range is default constructible,
 - verification that every component of the global range, local range and offset of a default constructed nd_range is 0, and that it compares equal to an nd_range built from zero ranges,
 - verification that a fully specified nd_range can be assigned to a default constructed one.

get_group_range() is deliberately not queried on a default constructed nd_range, as that would divide the global range by a local range of zeros.

### DIFF
--- a/tests/nd_range/nd_range_constructors.cpp
+++ b/tests/nd_range/nd_range_constructors.cpp
@@ -18,6 +18,8 @@ enum class op_codes : size_t {
   ctor_move = 2,
   assign_copy = 3,
   assign_move = 4,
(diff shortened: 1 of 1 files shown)


keryell (approved):
Thanks!

PR #1237, opened 2026-09-18
noexcept move constructible & assignable tests

Robertkq:
This PR is needed for [SYCL-Docs-1044](https://github.com/KhronosGroup/SYCL-Docs/pull/1044) issue, which introduces `noexcept` specifier to the move constructor and move assignment operator for both `by value` and `by reference` common semantics (more info in the issue itself).

* Adds `noexcept` specifier checks to the shared type traits test for `by reference`, alike `by value` which does not have a shared suite between them, as such it adds checks manually. `id.cpp` seems to be quite different compared to the other `by value` classes, I've kept the relative coding style

* Adds a missing test plan for the `by value` tests
> AI-Generated -> modified to contain accurate information

Assisted by Claude.

Note:
There is already an [open PR](https://github.com/intel/llvm/pull/23184) in DPCPP to extend their own testing to this, which uses quite a clever trick to have a strict check for `T& operator=(T&&)` and then test `noexcept` too, as:
```cpp
(std::is_nothrow_move_constructible_v<T>);
(std::is_nothrow_move_assignable_v<T>);
  ```
  does not assert existance of `T(T&&)` or `T& operator=(T&&)`, the copy overload **might** be able to also satisfy it. Although they have been able to assert existance of `T& operator=(T&&)`, the same cannot be done for `T(T&&)` AFAIK.
  -- wanted to point this out in case such a check would be desired for CTS as well.

### CODE REVIEW

**Comment by Robertkq** [comment]
This is my first PR in `SYCL-CTS`, I hope that all requested changes are up to standards and correct, please let me know any feedback and I'll try to cover them as soon as possible, thanks :)

**Comment by Robertkq** [comment]
@TApplencourt kind ping, asking for review

File: tests/nd_item/nd_item_equality.cpp [hunk 1]
```diff
@@ -35,6 +35,9 @@ class TEST_NAME : public util::test_base {
     using item_t = sycl::nd_item<numDims>;
     using kernel_t = nd_item_equality_kernel<numDims>;
 
+    // Check that move construction/assignment are noexcept
+    common_by_value_semantics::check_move_noexcept<item_t>();
```
TApplencourt: Oh... it's also name `item_it` here, even if it's `nd_item`.  Whatever, I agree with you better to keep the existing convention.

TApplencourt:
Amazing thanks a lot!

I'm not an expert, will let C++ guru @nliber  review if he have some free time.

But my recap:

- The the by-reference semantic only require the modification of the `trait` and then will be instantiate for all the type listed in 4.5.2. Common reference semantics
- For the `by-value` is a case-by-case no strict rules. But I checked that all of the object list `4.5.3. Common by-value semantics` have been modified.

LGTM ! Thanks again!

TApplencourt (approved):

TApplencourt:
All the CTS passing made me a little suspicious... Does all the implementation skip this tests / already where no-throw?

nliber (approved):
LGTM

Robertkq:
> All the CTS passing made me a little suspicious... Does all the implementation skip this tests / already where no-throw?

Just so I understand correctly, part of the CI is running CTS against supported implementations, such as DPCPP and others?

I'm not really surprised the tests pass, as noted in the PR description, the standard type traits do not check for the move constructor / assignment, rather if the move expression can be satisfied (this is either via move or copy alternative), however noexcept is still checked.

We could enforce checking of `T& operator=(T&&) noexcept` but we cannot do the same for the constructor

Do you believe the PR can be merged as is or you want me to add the enforced checking for the move assignment operator?

keryell (approved):

TApplencourt:
Oh I see. 
We got 2 approval, so I think we are good then : )

PR #1238, opened 2026-09-22
Fix/group broadcast wg limit 

sfuloria10:
### **Problem**

work_group_range<D>(queue) in tests/common/get_group_range.h queries device::max_work_group_size to determine the launch size. That is a device-wide value and does not account for the per-kernel work-group size limit a driver may
report after compiling a specific kernel. Some platforms do not support the device maximum work-group size for every kernel — a kernel whose register or private-memory footprint lowers occupancy gets a smaller limit — so launching at
the device-wide maximum throws for those kernels. This was hit by `group_broadcast` at D=3 on one such platform

### **Fix**

Add a KernelName-templated overload of work_group_range_for_kernel that:

1. Pre-compiles the kernel via get_kernel_bundle
2. Queries the actual per-kernel limit using kernel_device_specific::work_group_size
3. Uses that limit for the hypercubic decomposition instead of the device-wide max

The decomposition logic is extracted into detail::make_hypercubic_range shared by both overloads. The existing device-wide overload is unchanged — no other callers are affected. broadcast_group is updated to pass broadcast_group_kernel<D, T> to the new overload.


tests/common/get_group_range.h
```diff
@@ -36,18 +36,15 @@ inline sycl::range<3> get_default_range() {
...
+ *        and the device's max_work_item_sizes.
  * @tparam Dimensions Dimension to use for group instance
  */
 template <int Dimensions>
-sycl::range<Dimensions> work_group_range(
-    sycl::queue queue,
-    size_t work_items_limit = std::numeric_limits<size_t>::max()) {
-  // query device for work-group sizes
+sycl::range<Dimensions> make_hypercubic_range(sycl::queue& queue,
+                                              size_t max_work_group_size) {
```
bader: Can we keep this function unchanged and just pass the kernel limit via `work_items_limit` parameter?
I think we must keep respecting device limits.

bader (changes requested):

tests/common/get_group_range.h
sfuloria10: Good point — the overload was substituting the per-kernel limit for device::max_work_group_size rather than applying both, so the device limit wasn't respected.

Reworked in the latest commit: work_group_range is unchanged, and a query-only helper max_work_group_size_for_kernel<KernelName>() returns the per-kernel limit, which the call site passes as work_items_limit. That way work_group_range still does min(device_max, limit) and both limits hold. This also matches how group_barrier already passes a memory-derived cap through the same parameter, and it means the shared decomposition helper is no longer needed.

Let me know if this looks right to you.

bader:
Please, update PR title:
"Fix/group broadcast wg limit" -> "Fix work-group limit for group_broadcast test."

tests/common/get_group_range.h
bader: New version looks better. Thanks!

tests/common/get_group_range.h
```diff
@@ -87,6 +87,25 @@ sycl::range<Dimensions> work_group_range(
...
+ * @tparam KernelName SYCL Kernel name (must be defined in the same TU)
+ */
+template <typename KernelName>
+size_t max_work_group_size_for_kernel(sycl::queue queue) {
+  auto kernel_id = sycl::get_kernel_id<KernelName>();
+  auto bundle = sycl::get_kernel_bundle<sycl::bundle_state::executable>(
+      queue.get_context(), {queue.get_device()}, {kernel_id});
+  return bundle.get_kernel(kernel_id)
+      .get_info<sycl::info::kernel_device_specific::work_group_size>(
+          queue.get_device());
```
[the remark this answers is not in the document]
sfuloria10: Reworked as you suggested — work_group_range is untouched and the per-kernel limit goes through its existing work_items_limit parameter, so min(device_max, kernel_limit) applies. Same pattern group_barrier already uses for its memory-derived cap.

Two CI failures on the first attempt, both fixed: get_info needed .template (dependent type), and AdaptiveCpp has no sycl::get_kernel_bundle at all, so that path is now guarded with a fallback to device::max_work_group_size.

On the fallback — my analysis, which I'd like you to check since it drives the choice: at the commits pinned in CI, both AdaptiveCpp and ProtoSYCL implement kernel_device_specific::work_group_size as dev.get_info<info::device::max_work_group_size>(). If that's right, the query only differs from the device maximum on DPC++, so the fallback gives up nothing. Copilot suggested using sycl::program there instead, but that reaches the same value via a dummy program/kernel, and sycl::program was removed in SYCL 2020.

Does that reasoning hold, or have I missed something?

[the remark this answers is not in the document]
bader: Make sense to me. Thanks!

keryell (approved):
Thanks!

